PR #31, opened 2025-10-17
add another complex chi router to testdata

sattellite:
This is my real project structure and `apispec` cannot generate any data from it. I just suggest adding it to the project.


ehabterra:
@sattellite It should be fixed now in v0.3.1. Thank you for helping me address this issue, and I'm looking forward to your feedback!

PR #85, opened 2026-06-24
Fix external-type resolution, route param scoping, mount duplication, and generic determinism

ehabterra:
Four independent, validated fixes to the spec generator. Each was verified by diffing the pre- vs post-change binary with identical flags so the effect is exactly the intended one (the engine has some pre-existing run-to-run nondeterminism, so single-run snapshot diffs were cross-checked across multiple runs).

## 1. External types resolved via a layered registry (`c02e0de`)
External named types with a custom JSON/Text marshaler (uuid.UUID, decimal.Decimal, sql.Null*, …) were collapsed to a bare `string` at metadata-build time, dropping their semantic format and making the documented `typeMapping` escape hatch unreachable. Resolution now lives in the right layer: metadata records facts only; the spec layer applies config → a well-known registry → a structural marshaler rule. `uuid.UUID` now renders as `{string, format: uuid}`; `[]uuid`/`*uuid`/repeated fields inline with no dangling `$ref`s.

## 2. Scope net/http `Get` param patterns to their receiver (`3a85741`)
The header-param pattern matched any call named `Get`, so `http.Get("https://…")` produced a bogus `{name: "https://…", in: header}` parameter. Scoped to `net/http.Header`; added a `net/url.Values` query pattern so `r.URL.Query().Get("q")` is correctly a query param.

## 3. Drop spurious partial-mount duplicate routes (`75708e2`)
Nested mounts (main mounts `/api`→handler, handler mounts `/user`→userRoutes) emitted the same handler at every subset of its mount chain (`/api/user/{id}` plus `/api/{id}`, `/user/{id}`, `/{id}`). complex_chi_router went 13→40 paths and the duplication caused nondeterministic response resolution. A post-traversal pass drops routes whose mount segments are a strict subsequence of a more-complete one; genuine multi-mounts at distinct prefixes survive. complex_chi_router is back to a deterministic 13 paths.

## 4. Deterministic generic type-argument rendering (`1e80a63`)
A generic handler's operationId concatenated type args by iterating a map, flipping `HandleRequest[A, B]` ↔ `[B, A]` between runs. Now sorted.

## Testing
`go test ./...` passes; `go vet` and `golangci-lint` clean. New unit tests cover the external-type chain, wrapped-form composition, and the mount-dedup subsequence logic.

## Release Notes

* **New Features**
  * Enhanced external/custom-named type handling with marshaler-aware schema inference (Text/JSON), including improved string/format mapping (e.g., UUID) and safer inlining for primitive-shaped results.

* **Bug Fixes**
  * Improved determinism in generated OpenAPI: stable type-argument formatting and conflict-free response selection.
  * Reduced duplicate route variants by dropping subsumed mount-prefix routes.
  * Tightened header `Get` matching to avoid misclassification.

* **Tests**
  * Added unit tests for external type resolution (registry, facts, config precedence) and for mount/deduplication behavior.

### DIFF
--- a/internal/metadata/metadata.go
+++ b/internal/metadata/metadata.go
@@ -225,6 +225,9 @@ func GenerateMetadataWithLogger(pkgs map[string]map[string]*ast.File, fileToInfo
 
 		// Set the current module path
 		CurrentModulePath: currentModulePath,
+
+		// External-type facts discovered during the type walk.
+		ExternalTypes: make(map[string]ExternalTypeFact),
 	}
 
 	for pkgName, files := range pkgs {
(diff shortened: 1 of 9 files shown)


ehabterra:
Thanks @coderabbitai — all five points addressed in f62355a:

1. **`marshalerKind` priority + signature** (`metadata.go`): now checks `MarshalJSON` before `MarshalText` (encoding/json prefers `MarshalJSON`), and validates the full `func() ([]byte, error)` signature rather than a zero-arg method of the right name.
2. **Cycle detection** (`recordExternalTypeFacts`): added a visited `*types.TypeName` set so recursive externals like `type T []T` can't overflow the stack.
3. **Registry correctness** (`external_types.go`): dropped ULID (Crockford Base32 — `format: uuid` would reject it) and `database/sql.Null*` (they marshal as structs, not scalars). Both now resolve correctly through the general rules (ULID → string via `MarshalText`; `sql.Null*` → struct component).
4. **Wrapped-type matching**: short-name matching is now gated to bare named types (`typeNamesMatch`), so a scalar `uuid.UUID` mapping no longer matches `[]uuid.UUID`; explicit wrapper mappings still match exactly. Added tests.
5. **Primitive `typeMapping` components** (`mapper.go`): primitive-shaped mapped schemas are no longer marked as used types, so a second occurrence can't emit a `$ref` to a component that's never generated.

Verified: full test suite + lint pass, and the changes produce no output differences on existing testdata (pure robustness).

ehabterra:
Follow-up addressed in the latest commit: `lookupConfigSchema`/`configHasExternalType` now match exact names in a first pass across all entries before falling back to short-name matching, so an earlier short match can no longer shadow a later exact one. Added a regression test.

ehabterra:
@coderabbitai review

PR #87, opened 2026-06-24
Resolve all status codes for a response helper called from multiple branches

ehabterra:
Follow-up to #85.

## Problem
A handler whose responses go through a single indirection helper called from several branches with different status literals only surfaced the **first** status. For example `HandleRequest`'s wrapper:

```go
respondWithError(w, "Invalid request", http.StatusBadRequest)      // 400 (decode failure)
respondWithError(w, err.Error(), http.StatusInternalServerError)   // 500 (handler error)
respondWithSuccess(w, response)                                    // 200 (success)
```

`testdata/generic` produced only `200` + `400` — the `500` was dropped. `extractResponseFromNode` deduped by the callee edge alone, so the second `respondWithError` call found the **shared** `WriteHeader` edge already visited and skipped its `500`.

## Fix
- **Qualify the visited-edge key by call site** (parent node key) so each distinct status is resolved per branch rather than collapsed to the first.
- That also surfaces the helper's response via internal encoder-chain paths that carry no status context, yielding a redundant `default`. `buildResponses` now drops a `default` ("status could not be determined") that adds **no response-body information** — either its body already appears at a resolved status, or it is the bare generic `object` fallback — while keeping a `default` whose body is concrete and distinct (a real response whose status genuinely couldn't be resolved).

`testdata/generic` now yields `200`/`400`/`500`. Validated across all testdata: only intended changes (generic gains `500`; enum_validation/fiber drop a *redundant* `default` whose body is already at a resolved status — `User` at 201, `fiber.Map` at 400/404). No regressions; pre-existing response nondeterminism unchanged.

Added unit tests for the default-drop logic.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved response extraction to avoid incorrectly merging results when the same callee is reached through different call sites.
  * Removed redundant OpenAPI `default` responses from generated output when they are uninformative compared to concrete HTTP statuses.
  * Preserved meaningful `default` responses when they differ from other statuses or when they are the only response present.
* **Tests**
  * Added coverage to verify uninformative `default` dropping behavior and ensure distinct defaults aren’t conflated.

### DIFF
--- a/internal/spec/extractor.go
+++ b/internal/spec/extractor.go
@@ -487,7 +487,7 @@ func (e *Extractor) extractRouteChildren(routeNode TrackerNodeInterface, route *
 		// Extract responses (multiple if status fan-out applies — see
 		// ExtractResponse / issue #39). Each emitted ResponseInfo lands
 		// under its own status-keyed slot in route.Response.
-		for _, resp := range e.extractResponseFromNode(child, route, visitedEdges) {
+		for _, resp := range e.extractResponseFromNode(child, route, visitedEdges, routeNode.GetKey()) {
 			if resp != nil && (resp.BodyType != "" || resp.StatusCode != 0) {
 				route.Response[fmt.Sprintf("%d", resp.StatusCode)] = resp
 			}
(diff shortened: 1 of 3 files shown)


ehabterra:
Addressed in the latest commit: `uninformativeDefault` now compares the actual rendered `Schema` (via marshaled identity) rather than the `BodyType` label, so two responses sharing an empty `BodyType` but different bodies are no longer conflated. Added a regression test for that case.

PR #119, opened 2026-07-13
chore: delete dead code kept alive only by its own tests

ehabterra:
## What

Phase 1 of the coverage push (after #118's measurement fix): a dead-code audit with `golang.org/x/tools/cmd/deadcode` (RTA over all four binaries), cross-checked against per-function coverage and manual caller searches. **-3,697 lines net**, zero output drift.

## Deleted (unreachable from every binary; only their own tests referenced them)

**`internal/spec` — dead visualization/export surface:**
- `GenerateCytoscapeHTML`, `ExportCytoscapeJSON`, `ExportCallGraphCytoscapeJSON`, `GenerateOptimizedCallGraphHTML`, the whole `PaginatedCallGraphServer` (HTTP server + `GenerateServerBasedCytoscapeHTML`), `DrawTrackerTree` (Mermaid), `DrawTrackerTreeCytoscape`, `buildCallPaths`, and the orphaned `server_template.html`/`server_ui.html` copies (diagserver embeds its own).
- The live diag path is untouched: `DrawCallGraphCytoscape`, `DrawTrackerTreeCytoscapeWithMetadata`, `GenerateCallGraphCytoscapeHTML`, `GeneratePaginatedCytoscapeHTML`, `PaginatedCytoscapeData`, ordering/traversal helpers.

**Interface shrink:** `TrackerTreeInterface` → `GetRoots`/`GetMetadata`; `TrackerNodeInterface` loses `GetArgType`/`GetArgIndex`/`GetArgContext`/`GetRootAssignmentMap`. No consumer ever called the removed methods — they existed to satisfy the interface, and the interface carried them for the deleted exporters. Implementations in `LazyTree`/`TrackerTree`/mock go with them, including the per-root `rootAssignmentsFor` computation whose result nothing read, and `TrackerTree`'s never-called `FindNodeByKey`/`TraverseTree`/`GetLimits`/`GetFunctionContext`/`GetNodeCount`/`TraceArgumentOrigin`/`FindVariableNodes`/`GetInterfaceResolutions`.

**Test-only helpers move to test scope:** `mock_tracker.go` → `mock_tracker_test.go` (trimmed to the surviving interface), `test_utils.go` → `test_utils_test.go`.

**`internal/metadata`:** `TypeEdges` (zero references anywhere), `DefaultImportName`, `FrameworkDetector.Add*/GetConfig` mutators, and the orphaned metadata-time external-type collapse (`resolveExternalNamedTypes` + `hasCustomJSONMarshaler`/`hasZeroArgMethod*`). That last one is the machinery the external-type layering fix deliberately disconnected — leaving it in place invites re-wiring the exact bug (metadata-time collapse loses formats). Its live replacement `marshalerKind` gets a direct test in exchange.

**`internal/engine`/`internal/profiler`:** `Engine.GetConfig`; `ProfileFuncWithResult`, `MeasureMemoryUsage`, `MeasureGoroutines`, `NewPerformanceAnalyzerWithThresholds`, `MetricsCollector.{IncrementCounter,RecordHistogram,GetSummary}`.

## Deliberately kept (also main-unreachable, but not dead)

- `generator/` public API and `pkg/patterns` — library surface for external consumers.
- `Engine.GetResolvedCallGraph` + `internal/callgraph` — tracker-redesign step-2 groundwork, documented as "off by default until the summary-based analyses consume it".
- `metadata.GenerateMetadata` and the `io.go` loaders — test-harness entry points (goldens, fixtures).
- `CallGraphSCC.SameComponent`/`InCycle`, `profiler.DefaultProfilerConfig`/`IsProfiling` — assertion helpers used by tests **of live behavior**; deleting them would degrade real tests.

## Verification

- Full suite green, including metadata goldens byte-identical, determinism tests, and all framework fixture structural tests — zero output drift.
- `golangci-lint`, `go vet`, `gofmt` clean.
- deadcode rerun post-cut: no new cascade in `internal/spec`; the remaining flagged entries are exactly the deliberate keeps above.
- Library `-coverpkg` coverage: 82.5% → **83.2%** (denominator shrink is honest — the deleted tests only ever tested the deleted code).

## Summary by CodeRabbit

- **New Features**
  - Improved analysis of external types, including JSON marshaling behavior.
  - Added more consistent handling and identification of inline anonymous structures.
  - Enhanced call-graph visualization metadata with richer call-path details.

- **Changes**
  - Simplified tracker and analysis interfaces by removing obsolete inspection and configuration options.
  - Streamlined export capabilities around HTML-based Cytoscape diagrams; legacy server, JSON, and Mermaid outputs are no longer available.

- **Tests**
  - Refocused coverage on type resolution, visualization metadata, and graceful handling of malformed analysis data.


ehabterra:
Addressing the three CodeRabbit findings (verified each against the code):

**1. `lazytree.go` — "increment nodesBuilt unconditionally / stop using recursion-stack depth": not applied.** The counter already is cumulative — the budget deliberately counts *globally distinct callee keys* (`seenKeys` gate), matching the eager tree's shared-node cap so both engines exhaust the same graph-sized budget on the same inputs (see the inline comment at the increment and CLAUDE.md golden rule 6; `testdata/dense_graph`/`cyclic_graph` guard the bound). Counting every per-path copy instead would change which fixtures hit the cap — a behavior change out of scope for a dead-code PR, and this diff didn't touch the counting semantics at all (it only removed the write-only `rootAssignments` line nearby). Per-path blow-up within a scope is separately bounded by `maxInstancesPerKey`.

**2. License year 2025 → 2026 on the two renamed files: not applied.** Repo convention is the file's *creation* year (per `git log --follow --diff-filter=A`). `mock_tracker_test.go` and `test_utils_test.go` are pure renames of files created in 2025 — git tracks them as renames (100% similarity), so 2025 stays.

**3. `RecoverFromPanic` stack-overflow claim: valid — fixed in 4b5668f.** Correct that Go reports stack exhaustion as a fatal error `recover()` never sees, so that branch could never fire. Removed the branch and the misleading claim; the helper now plainly reports recoverable panics and documents the caveat. (Runaway recursion in the code under test is already prevented by explicit bounds — `maxInstancesPerKey`, node budgets, visited maps.)

PR #132, opened 2026-07-13
perf(spec): intern extraction chain keys — fix quadratic allocation on deep graphs

ehabterra:
## Problem

Profiling spec generation on two large real projects (~100 pkgs / 5.5k call edges and ~40 pkgs / 8k call edges) showed the extraction walk dominated by GC (`runtime.madvise`, `runtime.scanObject*` ≈ 40% of samples). Line-level pprof pinned it to `extractRouteChildren`:

- `childChainKey = chainKey + chainSep + id` — an O(depth) string built for **every visited child** → quadratic total bytes over deep graphs (720ms CPU on this line alone)
- `extractRouteChildren` allocated **1.22GB — 57% of all process allocations**

Introduced after v0.4.0 with the response-candidate chain dedupe (af0fd3c), so this is a regression on deep/dense projects: on the deep-graph project, tracker+map went 3.3s (v0.4.0) → ~3.8s (main).

## Fix

Chain identity is interned to small int handles (`chainInterner`): each recursion step is one map operation, no string building on the hot path. Full chains are materialized only for actual response candidates (rare). The candidate dedupe key becomes a `chainStep{parent handle, callee ID}` struct — **value equality is preserved exactly**, so dedupe semantics are unchanged.

## Results (large real projects, local A/B)

| Metric | before | after |
|---|---|---|
| Process allocations | 2.13GB | 1.23GB (−42%) |
| `extractRouteChildren` alloc (cum) | 1.22GB | 332MB (flat: 1.5MB) |
| `extractRouteChildren` CPU | 20% of samples | out of profile top |
| "spec mapped" stage (deep-graph project) | 2.4–3.8s | 1.8–2.2s |

Combined with the lazy tracker, both projects are now faster than v0.4.0 end-to-end (one 13.7s → ~4s, the other 6.9s → ~5s).

## Zero output drift

- Outputs byte-identical before/after on **both** large projects
- Full suite (goldens, determinism, all framework fixtures) passes untouched
- gofmt / vet / golangci-lint clean

## Also included

CLAUDE.md: new "Profiling & performance" section documenting the existing profiling flags, pprof workflow, `make metrics-generate` / `metrics-view`, per-stage `[engine]` log diagnosis, and the O(depth)-per-child pitfall; plus the HTTP-method precedence chain and coverage-ratchet conventions from recent work.

## Summary by CodeRabbit

* **Performance**
  * Improved internal route analysis efficiency, reducing allocation overhead during complex route extraction.
  * Preserved deterministic response pairing and existing behavior.

* **Documentation**
  * Added contributor guidance for profiling, performance measurement, HTTP-method handling, and testing conventions.
  * Documented coverage expectations and guidance for functions with no test coverage.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -40,6 +40,31 @@ go test ./internal/metadata -run TestGenerateMetadata -update
 scripts/compare-spec.sh                            # regenerate/diff fixture snapshots
 ```
 
+## Profiling & performance
+
+```bash
+./apispec --dir <project> --cpu-profile --mem-profile --profile-dir profiles
+                     # also: --block-profile --mutex-profile --trace-profile
+go tool pprof -top ./apispec profiles/cpu.prof
+go tool pprof -list 'FuncName' -sample_index=alloc_space ./apispec profiles/mem.prof
+make metrics-generate     # per-stage metrics JSON (--custom-metrics --metrics-path)
+make metrics-view         # interactive metrics viewer (scripts/view_metrics.sh)
+```
+
+- First-line diagnosis is the per-stage `[engine]` log lines (loaded /
+  metadata generated / tracker tree / spec mapped). With LazyTree, tree
+  expansion happens *inside* the "spec mapped" stage — a bigger mapping
+  number alone is moved work, not necessarily a regression.
+- Benchmark on a large real project, never on `testdata/` fixtures — they
+  are so small that `go/packages` load noise dominates. A/B by building
+  binaries from two `git worktree`s of the versions under comparison.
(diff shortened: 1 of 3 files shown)


ehabterra:
Added `internal/spec/chain_interner_test.go` in 604160b: table-driven coverage for `push`/`strings` — empty chain (handle 0 → nil), 1-based stable handle reuse, sibling/nested distinctness, root-to-leaf reconstruction (and that reconstruction doesn't disturb interning state), plus the diamond case where the same callee reached through two frames must stay distinct, which is the property the response-pairing dedupe relies on.

PR #135, opened 2026-07-13
feat(engine): layer a scoped net/http config under the detected framework

ehabterra:
## The design question

> net/http is stdlib and never appears in go.mod — how do we detect it? Should we just append its config, and can appending hurt?

**Detection**: framework detection is already import-based (`internal/core/detector.go` parses every file's imports), not go.mod-based — but that doesn't help: `net/http` is imported by virtually every Go web project (for `http.ResponseWriter` alone), so "imports net/http" carries no signal. Detection cannot decide this.

**Appending the full `DefaultHTTPConfig` WOULD hurt.** Two concrete failure cases found by inspection:
- Its `^Handle$` route pattern is **unscoped** (no `RecvTypeRegex`) — merged into another framework it double-matches that router's own `Handle`/`Handle`-shaped methods with net/http extraction semantics (path from arg 0, `MethodFromPath`), producing duplicate verb-less routes.
- Its `(?i)(JSON|String|XML|...)` response catch-all reads **status from arg 0** — merged into fiber it misreads status-less `c.JSON(obj)`, taking the response object for a status.

**The answer: append a *scoped subset*.** `HTTPSecondaryConfig()` contains only patterns that are receiver- or package-scoped (`ServeMux`/`net/http` route patterns, `ResponseWriter`-scoped response patterns, encoder/decoder-scoped body patterns, `http.Header`/`url.Values`/`PathValue` params, scoped security patterns). Scoping makes the merge **inert unless the project actually uses stdlib routing/requests** — no gating heuristic needed. Applied only on the auto-detect path; user-supplied configs are never augmented.

## Merge machinery

`MergeFrameworkConfigs(primary, secondaries...)` — appends each pattern list with first-occurrence-wins dedupe keyed on the pattern's *matching surface* (CallRegex + receiver constraints), so the primary's variant of a shared pattern always beats a secondary's; request-context type regexes/body accessors accumulate uniquely; `Info`/`Defaults`/mappings stay the primary's. This is the general mechanism — future full multi-framework merging (e.g. gin+mux in one repo) can reuse it once each config's unscoped patterns get the same treatment.

## Evidence of no negative effect

- **Full suite passes unchanged** — every framework fixture, metadata goldens, determinism tests, with the merge active for all of them.
- **Two large real projects** (echo/gin-style, ~40 and ~100 packages): diffs are **purely additive** — 0 lines removed, no verb or route changes; the additions are real, previously-missed header parameters read via the raw `*http.Request` inside framework handlers (`X-IAP-JWT-Assertion`, `x-requested-from`, …).

## Fixture

`testdata/mixed_chi_nethttp` — a chi API plus plain default-ServeMux ops endpoints in one binary (the expvar/pprof wiring style):

| path | registration | expected |
|---|---|---|
| `/api/users` | chi `r.Get`/`r.Post` | GET + POST (POST carries `User` body) |
| `/ops/version` | `http.HandleFunc("GET /ops/version", ...)` | GET only |
| `/ops/status` | plain `http.HandleFunc` + `switch r.Method` | GET + DELETE, no POST preset |
| `/api/users` GET | raw `r.Header.Get("X-Request-ID")` in the chi handler | header param documented |

The structural test passes a **nil** fallback config so the engine's own detect+merge path is what's exercised.

## Summary by CodeRabbit

- **New Features**
  - Improved API generation for projects that combine multiple HTTP frameworks.
  - Combined routes, request bodies, parameters, and responses from supported frameworks while avoiding duplicate operations.
  - Added public configuration helpers for safely layering framework-specific API definitions.

- **Bug Fixes**
  - Prevented incorrectly assigned HTTP methods and route patterns when frameworks are combined.
  - Preserved request body and header parameter details in generated API documentation.

- **Tests**
  - Added coverage for mixed Chi/net/http and Gin/mux projects, including route methods, request bodies, headers, and reference validation.


ehabterra:
Good question — as of the commit you linked, a second *recognisable* framework was still invisible: `Detect` stopped at the first framework import, so a gin+mux binary documented only gin. The net/http merge didn't help because mux's patterns weren't in the config at all.

Extended in 9bd7a1e:

- **`DetectAll`** returns every recognised framework in first-seen order. The head is what `Detect` historically returned, so single-framework projects get the exact same primary config as before.
- Each additional framework merges in as a **`SecondaryView`** — a mechanical filter keeping only receiver/package-scoped patterns. That one rule is the whole safety argument: a scoped pattern cannot claim another framework's calls, the unscoped helpers every config repeats (Marshal/Encode/FormValue) already exist in the primary, and the known-dangerous unscoped patterns (net/http's JSON catch-all, its bare `Handle`) are precisely what the filter drops. A unit test sweeps all six shipped configs asserting the views are scoped-only yet never route-less.
- New fixture `testdata/mixed_gin_mux`: gin products API + gorilla/mux admin router in one binary — `/products` GET+POST (gin), `/admin/report` GET and `/admin/users/{id}` DELETE (mux `HandleFunc().Methods(...)` chains + path template), no POST leakage.

Zero-drift evidence re-run: full suite unchanged, and both large real projects are byte-identical to the previous PR head (they're single-framework, so the secondaries loop is a no-op — by design).

Known limitation, documented in `SecondaryView`'s doc comment: a *secondary* framework's unscoped Mount-style patterns are filtered too (e.g. chi's `^Mount$` carries no receiver constraint), so mounts wired through a secondary framework's router may not be traced yet. Scoping those patterns in their home configs is the follow-up that lifts this.

One composition note: verb *constants* in chains (`.Methods(http.MethodGet)`) resolve once #134 lands (constant-value recording in metadata); the fixture uses string literals, which work on this branch alone.

ehabterra:
Both review comments addressed in d44c383:

- `noUnresolvedPlaceholders(t, out)` added to both mixed fixture tests — verified the helper in `testdata_smoke_test.go` and matched the sibling framework tests' convention.
- `MergeFrameworkConfigs` now returns nil on a nil primary, mirroring `SecondaryView`'s guard (both exported via the public `spec` package), with a unit test pinning the behavior.

PR #148, opened 2026-07-16
fix(tracker): don't drop receiver routes on same-name middleware variable collision (#146)

### LINKED ISSUE #146: Lazy tracker drops receiver-registered routes when a middleware reassigns an identically-named variable (regression vs eager)
## Summary

The **default (lazy) tracker** silently drops routes registered directly on a router receiver variable (`r.Get(...)`, nested `r.Group(...)`) when a middleware invoked in the same closure via `r.Use(mw)` reassigns a variable that shares the receiver's name (`r`). The **legacy/eager tracker (`--legacy-tracker`) is not affected** and emits all routes.

This is a regression exposed by making `LazyTree` the default after v0.4.0 (v0.4.0 shipped the eager tracker only).

## Impact

On a realistic idiomatic chi app that uses the canonical `r = r.WithContext(ctx)` idiom inside a middleware, the lazy tracker dropped roughly **half** the operations (e.g. 22 of 50), while `--legacy-tracker` recovered all of them. The dropped set is everything registered directly on the receiver in the affected closure, plus whole nested `r.Group(...)` subtrees. Argument-passed helpers (`mountX(r)`) are unaffected.

The `r = r.WithContext(...)` pattern is extremely common in chi / net/http middleware, so this hits many real projects.

## Reproducer

A chi app where a middleware reassigns the request variable `r` (same name as the router receiver), with a nested group:

```go
func mw(next http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		r = r.WithContext(r.Context()) // reassigns `r` — collides with router receiver
		next.ServeHTTP(w, r)
	})
}

func routes() http.Handler {
	r := chi.NewRouter()
	r.Route("/api/v1", func(r chi.Router) {
		r.Group(func(r chi.Router) {
			r.Use(mw)
			r.Get("/tenant", tenantHandler)  // DROPPED
			mountCaps(r)                     // kept (argument-passed helper)
			r.Group(func(r chi.Router) {     // nested subtree DROPPED
				r.Use(mw)
				mountUsers(r)
			})
		})
	})
	return r
}
```

Expected: `/api/v1/tenant` and the nested-group routes appear. Actual (default lazy tracker): they are missing. With `--legacy-tracker`: all present.

(A single-route toy program does not reproduce — the middleware producer node still gets reached; the drop needs the nested-group density of a real router.)

## Root cause

1. The middleware body's `r = r.WithContext(ctx)` (an `*http.Request` named `r`) is recorded in the `r.Use(mw(...))` **call edge's `AssignmentMap`** under key `r` (metadata builds `AssignmentMap` from the callee body).
2. `Metadata.BuildAssignmentRelationships` reads that `AssignmentMap["r"]` and pairs the callee-body assignment (`ConcreteType: *net/http.Request`, `Func:` the middleware closure) with the **call edge whose caller is the router closure** — a different scope that also has an `r` (the `chi.Router` receiver).
3. `internal/spec/lazytree.go` `buildRelations`, assignment-claim loop:
   ```go
   edges := edgesByRecvVar[recvEdgeKey(getString(meta, rel.Assignment.VariableName), &rel.Edge.Caller)]
   ```
   keys on `rel.Edge.Caller` (the router closure) and so **claims the router closure's `r.Get`/`r.Group` edges**, re-homing them onto the middleware producer node — where they are never re-emitted as routes. The comment there assumes "the assignment's producing edge carries the full identity of the function the assignment lives in," which is false in this cross-scope name-collision case (`rel.Assignment.Func` ≠ `rel.Edge.Caller.Name`).

## Suggested fix

Gate the claim on the assignment actually living in the caller scope whose receiver calls are being claimed:

```go
if getString(meta, rel.Assignment.Func) != getString(meta, rel.Edge.Caller.Name) {
    continue // cross-scope name collision: a callee-body `r` leaked into this
             // edge's AssignmentMap; don't claim the caller's router receiver
}
```

(The `producerByVar` population above the claim already keys correctly by `rel.Assignment.Func`, so only the claim needs the guard.) The type mismatch — a `*net/http.Request` "producer" for a router receiver — is an equally valid discriminator.

Must ship drift-clean against the full golden + fixture suite, with a new `testdata/` fixture pinning the nested-group + `r`-reassigning-middleware reproducer.

## Workaround

Use `--legacy-tracker` until fixed.

ehabterra:
## Summary

Fixes #146 — a v0.5.0 release blocker. The default lazy tracker dropped (or mis-prefixed) routes registered directly on a router receiver variable (`r.Get`, nested `r.Group`) when a middleware installed in the same closure reassigned an identically-named variable. This is the canonical `r = r.WithContext(ctx)` middleware idiom, where the inner `r` is a `*http.Request` colliding with the `r chi.Router` receiver at the registration site.

On a real chi app this lost roughly **half** the routes (22 of 50 ops vs 43 with `--legacy-tracker`). The eager tracker was never affected; the regression surfaced when `LazyTree` became the default after v0.4.0.

## Root cause

1. The middleware body's `r = r.WithContext(ctx)` is recorded in the `r.Use(mw(...))` call edge's `AssignmentMap` (built from the callee body).
2. `Metadata.BuildAssignmentRelationships` pairs that callee-body assignment (`ConcreteType: *net/http.Request`) with the **caller's** `r chi.Router` — same name, different scope.
3. `LazyTree.buildRelations` then **claims** the caller's receiver registrations onto the middleware producer node, where they are re-homed under the wrong path prefix or dropped.

## Fix

Gate the assignment-claim on the assignment genuinely living in the edge's caller scope (`rel.Assignment.Func == rel.Edge.Caller.Name`), which rejects the cross-scope name collision. One guard, in `internal/spec/lazytree.go` `buildRelations`.

## Validation

- New fixture `testdata/chi_middleware_recv_shadow` + `TestTestdata_ChiMiddlewareRecvShadow` — **fails pre-fix** (routes appear as `/api/v1/auth/tenant` / `/api/v1/auth/users/` or missing), passes with the fix.
- Full `go test ./...`: 0 failures (metadata goldens are byte-compared → zero drift).
- `gofmt`, `go vet`, `golangci-lint` clean.
- A/B on a 245-path real project: **byte-identical** output old vs fixed; timing unchanged.

### DIFF
--- a/generator/testdata_chi_middleware_recv_shadow_test.go
+++ b/generator/testdata_chi_middleware_recv_shadow_test.go
@@ -1,0 +1,69 @@
+// Copyright 2026 Ehab Terra
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package generator
+
+import (
+	"testing"
+
+	"github.com/ehabterra/apispec/spec"
+)
+
+// TestTestdata_ChiMiddlewareRecvShadow guards the lazy-tracker regression where
+// a middleware that reassigns the request variable `r` (the canonical
+// `r = r.WithContext(ctx)` idiom) collided with the `r chi.Router` receiver at
+// the registration site. The callee-body `r` leaked into the caller-scope call
+// edge's AssignmentMap, and the tracker claimed the router's own receiver
+// registrations (a direct r.Get, a nested r.Group subtree) onto the middleware
+// producer — dropping them or re-homing them under the wrong path prefix
+// (e.g. /api/v1/tenant surfaced as /api/v1/auth/tenant). Every route must
+// appear at its correct path.
+func TestTestdata_ChiMiddlewareRecvShadow(t *testing.T) {
+	out := loadTestdataWithFixtureConfig(t, "chi_middleware_recv_shadow", spec.DefaultChiConfig())
+	noDanglingRefs(t, out)
+	noUnresolvedPlaceholders(t, out)
+
+	want := map[string][]string{
+		// Receiver-registered routes in the closure that installs the
+		// r-reassigning middleware — these regressed.
+		"/api/v1/tenant": {"GET"},
+		"/api/v1/users/": {"GET", "POST"}, // nested r.Group subtree
+		// Routes that always resolved (argument-passed helpers, sibling mounts).
+		"/api/v1/caps":          {"GET"},
+		"/api/v1/workflows":     {"GET"},
+		"/api/v1/notifications": {"GET"},
+		"/api/v1/auth/login":    {"POST"},
+		"/api/v1/auth/me":       {"GET"},
+	}
+	for path, methods := range want {
+		item, ok := out.Paths[path]
+		if !ok {
+			t.Errorf("path %q missing; have %v", path, mapPathKeys(out.Paths))
+			continue
+		}
+		for _, m := range methods {
+			if opFor(item, m) == nil {
+				t.Errorf("%s %s: expected operation, missing", m, path)
+			}
+		}
+	}
+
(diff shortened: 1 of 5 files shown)


ehabterra:
Closing as a duplicate of #147, which lands the identical fix (same guard in buildRelations) from an external contributor and includes a direct unit test. Validated #147 independently: full suite passes with zero golden drift, a 245-path real project is byte-identical, no perf regression. I'll add my reproducing end-to-end fixture (chi_middleware_recv_shadow, which fails pre-fix) as a follow-up on top of #147 to strengthen coverage.

PR #150, opened 2026-07-16
test(tracker): reproducing end-to-end fixture for the #146 receiver-claim fix

ehabterra:
## Summary

Adds a fixture that **actually reproduces #146 end-to-end**, strengthening the coverage of the fix in #147.

Stacks on **#147** — it contains that fix's commit until #147 merges, after which this reduces to just the fixture. Merge **after** #147.

## Why

#147's fixture (`chi_receiver_name_collision`) does not reproduce the drop: it installs the middleware as a bare func value (`r.Use(requestContext)`), which is too shallow to trigger the mis-claim, so its structural test passes **even without the fix**. #147's unit test guards the fix at the claim-map level, but nothing guards the full pipeline.

`chi_middleware_recv_shadow` closes that gap: it installs the `r`-reassigning middleware as a **call** (`r.Use(auditGuard(reporter()))`) with enough nested-group density that, without the guard, `/api/v1/tenant` and the nested users subtree drop — or surface mis-prefixed as `/api/v1/auth/tenant`. `TestTestdata_ChiMiddlewareRecvShadow` asserts every route at its correct path and **fails on the pre-fix code** (verified by removing the guard).

## Validation

- Passes on #147's base; fails with the guard removed (real regression guard).
- Full `go test ./...` green; zero golden drift.

### DIFF
--- a/generator/testdata_chi_middleware_recv_shadow_test.go
+++ b/generator/testdata_chi_middleware_recv_shadow_test.go
@@ -1,0 +1,69 @@
+// Copyright 2026 Ehab Terra
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


ehabterra:
Superseded — the reproducing fixture (chi_middleware_recv_shadow) is now committed directly onto #147 (commit f7707dc), alongside the fix for the CodeRabbit review comment. No need for a separate stacked PR.

Issue #192, opened 2026-07-18
metadata: Function.ReturnVars records only the max-arity return statement, dropping all others


ehabterra:
## Summary

`Function.ReturnVars` (and `Method.ReturnVars`) is populated to hold **only the return statement with the greatest number of values** — it clears and rebuilds *only* when a later return has strictly more values:

```go
// internal/metadata/metadata.go (~L295-312)
var returnVars []CallArgument
var maxReturnCount int
ast.Inspect(fn.Body, func(n ast.Node) bool {
    ret, ok := n.(*ast.ReturnStmt)
    if !ok { return true }
    if len(ret.Results) > maxReturnCount {          // strictly greater
        maxReturnCount = len(ret.Results)
        returnVars = nil                            // clear + rebuild
        for _, expr := range ret.Results {
            returnVars = append(returnVars, *ExprToCallArgument(expr, ...))
        }
    }
    return true
})
```

So for a function whose returns all have the same arity, `ReturnVars` keeps only the **first** one and silently drops the rest.

## Impact — blocks #187

The write-side status mapper pattern (#187) cannot be resolved:

```go
func MapError(err error) APIError {          // every return is single-valued
    if api, ok := mapAs404(err); ok { return api }   // ReturnVars keeps only THIS one
    if api, ok := mapAs401(err); ok { return api }   // dropped
    ... (403, 422, 409, 400)                          // dropped
    return APIError{http.StatusInternalServerError, "internal error"} // dropped
}
```

`ReturnVars` = `[api]` (the first `return api`). The other six returns — including the concrete `APIError{500, …}` literal — are gone, so the spec layer cannot enumerate the field's `{400,401,403,404,409,422,500}` value set. Observed on a real project: **66 operations** that report errors through such a mapper resolve to a bare `default`.

## Proposed fix

Record **all** return statements. Add a field alongside `ReturnVars` (keeping `ReturnVars`'s max-arity semantics for the existing `analysis.go` return-type-by-index callers, which rely on it):

```go
// Returns holds every return statement's value list, in source order.
Returns [][]CallArgument `yaml:"returns,omitempty"`
```

Populate it in both function and method builders (metadata.go ~L295 and ~L1172), wire serialization (`io.go` `setCallArgumentMeta` over each element), and regenerate the metadata goldens via `-update`.

[...]

Issue #155, opened 2026-07-16
Response status assigned across switch/if branches then passed to an error constructor resolves to `default`


ehabterra:
## Summary

When a shared error mapper computes the status in a **`switch`/`if` across branches** and then hands that variable to an error constructor, the status can't be pinned to one literal and the operation gets a `default` response (and loses the concrete codes the branches clearly set, e.g. `404`).

```go
func writeError(w http.ResponseWriter, err error) {
    var statusCode int
    switch {
    case errors.Is(err, ErrNotFound):
        statusCode = http.StatusNotFound          // 404
    case errors.Is(err, ErrValidation):
        statusCode = http.StatusBadRequest        // 400
    default:
        statusCode = http.StatusInternalServerError // 500
    }
    RespondWithError(w, NewAPIError(err.Error(), statusCode)) // w.WriteHeader(err.Code)
}
```

Every handler that reports errors only through `writeError` shows `default` instead of `{404, 400, 500}`.

## Why it happens

`statusFromConstructorField` (and the parameter-threading path) resolve the constructor's status argument to the `statusCode` **ident**. `MapStatusCode` maps literals/constants, not a variable with multiple branch assignments, so no single code resolves → `default`. This is correct-but-incomplete ("honest over wrong"): the tool declines to guess one of several.

## What a fix requires

Fan **one** `WriteHeader`/constructor-arg out into the **set** of statuses its variable can hold:
1. When the resolved status argument is a local variable, collect its assignments in the enclosing function/method (reusing `callerAssignmentMap` from #154).
2. Keep only integer-constant / `http.Status*` values; if any assignment is non-constant, keep `default` for that residue (honest).
3. Emit one response per distinct resolved code.

This is a one→many change to the response-recording loop in `extractor.go` (today every path sets a single `respInfo.StatusCode`), so it needs its own fixture (switch- and if-assigned status, plus a non-constant branch to prove the honest residue) and zero-drift validation.

## Context

Found validating apispec against a private modular chi gateway (~83 ops): after #154 (constructor-field status through handler-factory closures) recovered 26 `400`s, this pattern accounts for the remaining `default`s on 18 endpoints. It is the concrete instance of the shared-error-mapper caveat noted in `docs/GAP_ANALYSIS.md` §7.3. Related: #144, #154.

PR #193, opened 2026-07-18
feat: resolve response status through an error mapper's struct field (#187, #192)

### LINKED ISSUE #187: Response status is a struct field branch-assigned inside a mapper function resolves to default (MapError pattern)
## Summary

The write-side sibling of #155. When the status handed to `http.Error`/`WriteHeader` is a **struct field** whose value is set across `switch`/`if` branches **inside a different function** (an error mapper), the operation still gets a `default` instead of the concrete `{4xx, 5xx}` the branches set.

```go
type apiError struct { Message string; Status int }

func MapError(err error) apiError {           // the branch set lives HERE
    switch {
    case errors.Is(err, ErrNotFound):   return apiError{"not found", http.StatusNotFound}   // 404
    case errors.Is(err, ErrForbidden):  return apiError{"forbidden", http.StatusForbidden}  // 403
    default:                            return apiError{err.Error(), http.StatusInternalServerError} // 500
    }
}

func writeError(w http.ResponseWriter, r *http.Request, err error, op string) {
    api := MapError(err)
    http.Error(w, api.Message, api.Status)    // status = api.Status (a field of MapError's return)
}
```

Every handler reporting errors through `writeError` shows `default`.

## Difference from #155

#155 (fixed) covers a status **local variable** branch-assigned in the **same function** and passed to an error **constructor** (`statusCode` → `NewAPIError(msg, statusCode)`). This pattern is harder:
- the status is a **struct field** (`api.Status`), not a bare local;
- the branch set is **inside another function** (`MapError`), reached across the call boundary;
- the branches are `return apiError{..., <status>}` composite literals (or field assignments), not assignments to one local.

`statusesFromConstructorField` / `expandVarStatuses` (from #155) don't apply: `constructorFieldParam` expects the field to key to a constructor **parameter**, but here it keys to per-branch **literals inside the mapper's returns**.

## What a fix requires

1. Resolve the status selector (`api.Status`) to its producer function (`MapError`) via the local's assignment (`api := MapError(err)`).
2. In that function, collect the field's value across all `return` composite literals (and/or field assignments across branches), mapping each to a status via `MapStatusCode`.
3. Fan out to one response per distinct code; keep a `default` residue for any non-constant branch (honest, as in #155).
4. Fixture (switch-returning mapper + field access) + zero-drift validation.

## Context

Found validating apispec against a private modular gateway (~66 endpoints route errors through this exact `MapError`/`writeError` shape) — it accounts for the `default` responses that remain after #155. Related: #155, #154, #144. GAP_ANALYSIS.md §7.3 shared-error-mapper caveat.

ehabterra:
Closes #187 and #192. Foundation + feature, shipped together (the metadata change has no consumer on its own).

## The pattern (real-project, ~250 handlers)

```go
api := MapError(err)
http.Error(w, api.Message, api.Status)   // status = a struct FIELD

func MapError(err error) APIError {
    if api, ok := mapAs404(err); ok { return api }   // -> 404 (via helper)
    if api, ok := mapAs400(err); ok { return api }   // -> 400
    return APIError{http.StatusInternalServerError, "internal error"} // -> 500
}
```

`api.Status` resolved to a bare `default`.

## Commit 1 — #192 (metadata foundation)

`Function.ReturnVars` recorded only the **max-arity** return statement, dropping all others. Since `MapError`'s returns are all single-valued, only the first `return api` survived — the concrete `500` literal and the other branches were gone.

Add `Returns [][]CallArgument` (every explicit return, source order) beside `ReturnVars` (kept for the return-by-index callers). Shared `extractReturns` helper; io.go wires meta; goldens regenerated via `-update`. **Pure fact addition — zero spec-output drift** across all testdata + real projects (verified by byte-diff; only the metadata goldens gain `returns:`).

## Commit 2 — #187 (the resolver)

`statusesFromMapperField`: resolves the selector base to the mapper call, then enumerates the concrete statuses the field takes across **every** return of the mapper — recursing through per-status helper mappers (`return api` where `api := mapAsXxx(err)`) and reading struct literals both **keyed** (`APIError{Status: 404}`) and **positional** (`APIError{404, "..."}`, indexed via the type's field order). A non-constant branch keeps an honest `default`; nothing is guessed. Wired as a third resolver after the #155 constructor-field path.

## Additive & honest — verified

A/B byte-diff (this branch vs the metadata-only baseline) across every testdata fixture and all configured real projects:

| | concrete removed | effect |
|---|---|---|
| every testdata fixture | 0 | only the new fixture differs |
| every other real project | 0 | unchanged |
| one real project | **0** | **`default` 66 → 9** (57 resolved), +concrete error-code sets |

**Zero concrete statuses removed anywhere** — the only removals are `default`s that became concrete. The 9 remaining defaults are genuinely non-constant branches (honest residue).

Note: this fans the mapper's full error-code set (e.g. `{400,401,403,404,409,422,500}`) onto every handler that reports through the mapper — the sound over-approximation, since any domain error can map to any of them. If that reads as too broad, a "collapse large shared-body error sets" policy is a separate follow-up, not a correctness issue here.

## Tests

`testdata/mapper_field_status` + structural test; unit tests for every resolver branch (keyed/positional/cross-package/pkg-qualified fields, helper recursion, cycle + depth guards, honest residue). Full suite, `-race`, lint, determinism pass; coverage at the 96% floor.

## Summary by CodeRabbit

- **New Features**
  - Improved detection of concrete HTTP response statuses returned through status-mapping patterns.
  - Added detailed metadata for every explicit function and method return statement.

- **Bug Fixes**
  - Resolved previously unresolved or default response statuses in generated specifications.
  - Improved metadata handling for nested return values.

- **Tests**
  - Added coverage for mapper-based status resolution, return metadata, and updated specification fixtures.

### DIFF
--- a/generator/testdata_mapper_field_status_test.go
+++ b/generator/testdata_mapper_field_status_test.go
@@ -1,0 +1,48 @@
+// Copyright 2026 Ehab Terra
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package generator
+
+import (
+	"sort"
+	"testing"
+
+	"github.com/ehabterra/apispec/spec"
+)
+
+// TestTestdata_MapperFieldStatus covers issue #187: the status handed to
+// http.Error is a struct field (api.Status) whose value is set across the return
+// branches of an error mapper (api := MapError(err)). It must resolve to the
+// concrete branch statuses {400, 404, 500} — recursing through the per-status
+// helper mappers and reading positional struct literals — not a bare default.
+func TestTestdata_MapperFieldStatus(t *testing.T) {
+	out := loadTestdataWithFixtureConfig(t, "mapper_field_status", spec.DefaultHTTPConfig())
+	noDanglingRefs(t, out)
+	noUnresolvedPlaceholders(t, out)
+
+	get := opFor(out.Paths["/thing"], "GET")
+	if get == nil {
+		t.Fatalf("GET /thing missing; have %v", mapPathKeys(out.Paths))
+	}
+	got := keysOf(get.Responses)
+	sort.Strings(got)
+	for _, want := range []string{"400", "404", "500"} {
+		if _, ok := get.Responses[want]; !ok {
+			t.Errorf("GET /thing missing status %s; have %v", want, got)
+		}
+	}
+	if _, ok := get.Responses["default"]; ok {
+		t.Errorf("GET /thing still has an unresolved default; the mapper statuses should be concrete; have %v", got)
+	}
+}
--- a/generator/testdata_mapper_field_status_test.go
+++ b/generator/testdata_mapper_field_status_test.go
@@ -37,6 +37,11 @@ func TestTestdata_MapperFieldStatus(t *testing.T) {
 	}
 	got := keysOf(get.Responses)
 	sort.Strings(got)
+	// The mapper resolves to exactly {400, 404, 500}; any extra status (or a
+	// spurious success code) is a regression.
+	if len(got) != 3 {
+		t.Fatalf("GET /thing responses = %v, want exactly [400 404 500]", got)
+	}
 	for _, want := range []string{"400", "404", "500"} {
 		if _, ok := get.Responses[want]; !ok {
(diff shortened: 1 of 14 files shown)


ehabterra:
Both CodeRabbit findings verified and addressed in 2ed6539:

- **Major — nested closure returns leaking (`metadata.go`).** Confirmed: `extractReturns` used `ast.Inspect`, which descends into `*ast.FuncLit` bodies, so a closure's `return`s were recorded on the enclosing function. Now skips `*ast.FuncLit`. Added `TestExtractReturns_SkipsNestedClosures` (handler-factory shape). Metadata goldens unchanged (no existing fixture had a closure-with-returns) and verified **zero spec drift** across testdata + a large real project — the fix removed pollution nothing consumed.
- **Minor — exact response-set assertion (`testdata_mapper_field_status_test.go`).** The mapper resolves to exactly `{400,404,500}` (no success path on that handler), so added `len(got) != 3` before the membership checks.

Full suite, `-race`, lint, determinism pass; coverage at the 96% floor.

PR #196, opened 2026-07-19
docs+test: document the json.Marshal response over-detection (#195); no gate

### LINKED ISSUE #195: Response over-detection: downstream client request/secondary bodies emitted as spurious `default` responses
## Summary

On a real multi-package project, the majority of remaining `default` responses are **false positives**: response detection walks past the handler's own response writes into **downstream service/client calls** and attributes *their* structs — outbound-request types and secondary bodies that never reach the handler's `w` — as a `default` response of the outer route.

Categorized (21 defaults on the project):

| count | category |
|---|---|
| **11** | a downstream **client request** type detected as a response |
| **6** | a secondary success body (a response struct **field**) |
| 3 | infra / empty (root, metrics, an empty stub handler) |
| 1 | other |

The handlers' *real* responses are already correct (a concrete `200` + the error codes); these `default`s are extra, spurious entries.

## Mechanism (traced)

A handler returns a proper response and delegates work to a use-case interface backed by an HTTP client:

```go
func (h *handler) Get(uc UseCase) func(w, r) {
    return func(w http.ResponseWriter, r *http.Request) {
        v, err := uc.Get(r.Context(), &domainObj)          // -> downstream client
        if err != nil { writeError(w, err); return }
        RespondWithSuccess(w, msg, ResponseDTO{...}, http.StatusOK) // the real 200
    }
}
```

`uc.Get` eventually reaches an HTTP **client** method that has **no `ResponseWriter`**:

```go
func (c *Client) Get(ctx, params *Request) (T, error) {
    payload, _ := json.Marshal(params)                     // OUTBOUND request marshal — jsonMarshalPattern matches this
    rawRes, _  := c.Call(ctx, payload, http.MethodPost, ...) // goes to the wire, not to `w`
    _ = json.Unmarshal(rawRes.Body, &response)             // the client's OWN response
    ...
}
```

`jsonMarshalPattern()` matches `json.Marshal(params)` and treats `params` (the request type) as a response body. The **write-destination gate** (`responseDestResolver.ShouldDrop`, #170/#181) is meant to drop a response whose value doesn't reach the handler's `w` — but here the match occurs in a function with **no writer in scope**, and the gate conservatively **keeps** it (unknown status → `default`) rather than dropping. The 6 secondary-body cases are the same shape from a different angle (a response struct field / secondary body detected downstream).

## Proposed direction: anchor detection on the write sink

The root cause is that `jsonMarshalPattern` matches an **encode call in isolation** — `json.Marshal(v)` — and then tries, after the fact, to decide whether the result reaches `w`. Detection starts from the wrong end. A `json.Marshal(v)` is not evidence of a response; only a **write to the handler's `w`** is.

Invert it: **anchor response detection on the write sink**, then resolve the body schema by tracing *backwards* from the sink to what was written.

- The sinks are the writes to the handler's response writer: `w.Write(b)`, `w.Write([]byte(...))`, `io.Copy(w, ...)`, and an encoder **bound to `w`** (`json.NewEncoder(w).Encode(v)` — already destination-gated via `responseDestResolver`, #170/#181). The writer is identified by provenance to the handler's writer parameter, exactly as `responseDestResolver.reachesWriter` already does.
- From a `w.Write(b)` sink, resolve `b`'s value back to its source: `b, _ := json.Marshal(v)` ⇒ the response body is `v`'s type; `b := []byte(...)` / a string constant ⇒ no JSON schema; `b := someHelper(...)` ⇒ follow the helper's returns (the `Function.Returns` machinery from #192).
- A `json.Marshal(v)` whose result is **never written to a `w`-bound sink** — the downstream client's outbound-request marshal in this bug, a `bytes.Buffer` marshal, a log line — is simply **not a response**. It is excluded structurally, because nothing anchors it, with no marshal-side "is a writer nearby" heuristic.

This makes the write-destination gate the *primary* signal rather than a post-hoc filter, and it subsumes the secondary-body field cases (a struct field marshaled downstream is likewise never sink-anchored). Honest-over-wrong still governs the backward trace: when `b`'s source can't be resolved to a concrete type, keep the response with an unresolved/`default` body rather than dropping a real write.

Needs fixture coverage for: (a) a genuine `w.Write(json.Marshal(v))` and `json.NewEncoder(w).Encode(v)` — body must resolve to `v`; (b) this bug's downstream-client `json.Marshal(params)` with no `w` in the chain — must produce **no** response; (c) `w.Write([]byte("ok"))` — response with no JSON schema, not a spurious body.

### Rejected: writer-in-scope gate (PR #196, reverted)

A `RequireWriterInScope` config gate — drop a `json.Marshal` when its enclosing function has no response-writer parameter — was implemented in PR #196 and **reverted**. It gates on a *proxy* ("is a writer lexically nearby"), not the real signal ("do the marshaled bytes reach `w`"), and exposing it as a config knob institutionalizes an incomplete heuristic; it also introduced its own false negative (a real response marshaled in a helper that receives `w` only indirectly). Per golden rule #9 (fix the root cause or file an issue — never a workaround that trades one wrong answer for another), the over-detection stays as this tracked bug until the sink-anchored model above lands. `testdata/downstream_client_not_response` is a **change-detector** for it (asserts the spurious `default` is still present; flips to fail when this is fixed).

## Reproduction note

A minimal single-hop fixture (handler → client method with `json.Marshal(params)`) does **not** reproduce it — the false attribution needs the deeper use-case-interface → client indirection, so this is a tracker-context precision gap. Filing with anonymized real-project evidence rather than a synthetic repro, per the "fix the root cause or file an issue" rule.

## Not in scope here

The remaining `default`s on the same project are **honest residues** (a `default` alongside the *full* concrete status set, from a genuinely non-constant branch) and **infra/static** handlers (root, metrics, docs) — those are correct/acceptable, not bugs.

ehabterra:
Does **not** close #195 — it documents the over-detection and deliberately does **not** ship a fix.

## What happened

`json.Marshal(v)` returns `[]byte` with no writer argument, so it can't be destination-gated like `json.NewEncoder(w).Encode(v)`. Result: every reachable `json.Marshal` is treated as a response, and a downstream client's **outbound-request** marshal leaks as a spurious `default`.

I first tried a `RequireWriterInScope` gate (drop the marshal when the enclosing function has no response writer). On review that's a **workaround**, for two reasons:

1. It gates on a **proxy** ("is a writer nearby") rather than the real signal ("do the marshaled bytes reach `w`?"). It's a *necessary*, not *sufficient*, condition.
2. Exposing that proxy as a **config knob** institutionalizes a heuristic we already know is incomplete — it introduced its own false negative (a helper marshal whose bytes a different helper writes was dropped). That's trading one wrong answer for another (golden rule #9).

So the gate is **reverted**. The honest state is the tracked bug, not a config-flagged heuristic.

## What this PR keeps

- `json.Marshal` stays ungated; a `NOTE` on `jsonMarshalPattern` points at #195.
- `testdata/downstream_client_not_response` — a **change-detector**: asserts the spurious `default` is still present, and fails when #195's root-cause fix lands.
- CLAUDE.md — the general rule (probe every uncovered shape with a fixture; file an issue if real; keep it as a change-detector when deferred).

No behavior change vs `main` (the gate never merged). The root-cause fix — anchor response detection on the **write sink** (`w.Write(x)` / encoder bound to `w`) and trace the written value's type back — is tracked in **#195**.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -144,6 +144,14 @@ make metrics-view         # interactive metrics viewer (scripts/view_metrics.sh)
 
 ## Testing conventions
 
+- **Probe every uncovered shape with a fixture.** When you hit a code shape no
+  existing fixture covers — a new wiring style, an edge case, a suspected gap —
+  reproduce it as a `testdata/` fixture first, then decide from what it shows:
+  if it exposes a real bug, file an issue (golden rule #9) and, when the fix is
+  deferred, keep the fixture as a change-detector (assert the current behavior
+  with a comment so the test flips when it's fixed); if it already works, the
+  fixture is regression coverage. Never assert a case is (un)covered from
+  reasoning alone — build the fixture and look.
 - **Fixture projects** live in `testdata/<name>/` (a `main.go` + `go.mod`).
   Every fixture is wired into `go test` via structural tests in `generator/`
   (`testdata_*_test.go`): expected routes/methods present, no dangling
--- a/generator/testdata_downstream_client_not_response_test.go
+++ b/generator/testdata_downstream_client_not_response_test.go
@@ -1,0 +1,50 @@
+// Copyright 2026 Ehab Terra
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package generator
+
+import (
+	"testing"
+
+	"github.com/ehabterra/apispec/spec"
+)
+
+// TestTestdata_DownstreamClientNotResponse pins the KNOWN over-detection bug of
+// issue #195: json.Marshal is matched as a response with no write-sink check, so
+// a downstream HTTP client's OUTBOUND-request marshal (in a method with no
+// ResponseWriter) leaks as a spurious `default` response of the outer route.
+// The handler's real responses are the success 200 and the error 500.
+//
+// This is a CHANGE-DETECTOR for the unsolved bug: it asserts the spurious
+// `default` is still present. When #195's root-cause fix lands (anchor response
+// detection on the write to w and trace the written value's type back), the
+// spurious default disappears — this test then fails, prompting the update to
+// require exactly {200, 500}.
+func TestTestdata_DownstreamClientNotResponse(t *testing.T) {
+	out := loadTestdataWithFixtureConfig(t, "downstream_client_not_response", spec.DefaultHTTPConfig())
+	noDanglingRefs(t, out)
+
+	get := opFor(out.Paths["/pk"], "GET")
+	if get == nil {
+		t.Fatalf("GET /pk missing; have %v", mapPathKeys(out.Paths))
+	}
+	for _, want := range []string{"200", "500"} {
+		if _, ok := get.Responses[want]; !ok {
+			t.Errorf("GET /pk missing status %s; have %v", want, keysOf(get.Responses))
+		}
+	}
+	if _, ok := get.Responses["default"]; !ok {
+		t.Errorf("issue #195 appears FIXED (no spurious `default` on GET /pk) — update this test to require exactly {200,500}")
+	}
+}
--- a/internal/spec/config.go
+++ b/internal/spec/config.go
@@ -803,6 +803,12 @@ func jsonMarshalPattern() ResponsePattern {
 		TypeArgIndex: 0,
 		TypeFromArg:  true,
 		Deref:        true,
+		// NOTE: json.Marshal(v) returns []byte with no writer argument, so it is
+		// not destination-gated. It therefore over-detects — a Marshal reachable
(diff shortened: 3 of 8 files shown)


ehabterra:
Fixed the critical determinism finding in `callerSignatureParams`: it now iterates `Files`/`Types` in sorted order and filters candidate methods by the edge's caller receiver (so two types sharing a method name resolve deterministically to the right one). Verified: full suite + determinism guards pass, and a real multi-package project generates byte-identical output across two runs.

ehabterra:
**Known limitation documented (issue #197).** As discussed, `RequireWriterInScope` is a *sound necessary condition*, not the actual response signal ("do the marshaled bytes reach `w`?"), so it has a real false negative: when the `json.Marshal` carrying the response type is in a helper with no writer and the write is in a *different* helper, the schema degrades to `[]byte`.

- It did **not** occur in the real projects tested (0 kept-response body changes), and this PR is net-positive there (removed 20 spurious client-request `default`s, degraded nothing).
- Added `testdata/marshal_helper_written_elsewhere` as a **change-detector** (asserts the current `[]byte`; flips to fail when #197 is fixed) and CLAUDE.md golden rule #11.
- Root-cause fix (tracked in #197): anchor detection on the write sink (`w.Write(x)` / encoder bound to `w`) and trace the written value's type back — explicitly **not** a second heuristic on top of this one.

Shipping this as the documented interim; #197 tracks the proper fix.

ehabterra:
Superseded by #198, which implements the actual root-cause fix (anchor response detection on the write sink and trace the written value back, retiring the marshal-anywhere pattern). This PR was docs-only and did not close #195; #198 does.

PR #215, opened 2026-07-25
fix(spec): scope framework-specific patterns so secondaries keep them (#211)

### LINKED ISSUE #211: Secondary frameworks silently lose their unscoped patterns (gin loses all params and request bodies)
In a mixed-framework project, every framework except the primary is filtered through `SecondaryView`, which keeps only patterns carrying a receiver or package constraint. The unscoped ones are dropped — and for some frameworks that is most of what makes them work.

Measured across all six configs (`len(cfg.Framework.X)` vs `len(SecondaryView(cfg).Framework.X)`):

| framework as secondary | routes | mounts | params | request bodies | responses |
|---|---|---|---|---|---|
| **gin** | 1/1 | 1/1 | **0/4** | **0/3** | 5/7 |
| **echo** | 1/1 | 1/1 | **1/4** | 3/3 | 8/8 |
| fiber | 1/1 | **2/3** | 4/4 | 3/3 | 10/10 |
| chi | 3/3 | 2/2 | **5/6** | 3/3 | 8/8 |
| mux | 6/6 | 2/2 | 1/1 | 2/2 | 6/6 |
| net/http | 1/2 | 1/1 | 3/5 | 0/2 | 5/7 |

A secondary gin keeps its route pattern — so its endpoints are *found* — but loses every param and every request-body pattern, so those endpoints are documented empty.

## Reproduce

Two files, one module. The chi file is deliberately tiny; the gin API is the real one:

```go
// a_admin.go
import "github.com/go-chi/chi/v5"
func adminRouter() *chi.Mux { r := chi.NewRouter(); r.Get("/admin/ping", ping); return r }
```
```go
// z_api.go
func ginCreate(c *gin.Context) {
    var req CreateReq
    _ = c.ShouldBindJSON(&req)
    _ = c.Query("filter")
    c.JSON(http.StatusCreated, req)
}
```

### Actual

```yaml
/gin/items:
  post:
    operationId: ginsec2.ginCreate
    responses:
      default:
        description: Default response (no response found)
        content: {application/json: {schema: {type: object}}}
components: {}
```

No `filter` parameter, no request body, no response schema, no summary — an empty operation.

### Expected

What the identical handler produces when gin is primary: the `filter` query parameter, a `$ref` to `CreateReq`, and the 201 response schema.

This is precisely the scenario the code advertises as supported — `engine.go`'s merge comment cites "a gin API next to a gorilla/mux admin router, half-migrated projects".

## Root cause

`SecondaryView` (`internal/spec/config_merge.go:120`) keeps a pattern only when `RecvType != "" || RecvTypeRegex != ""`. The rule is sound in itself — an unscoped pattern transplanted into another framework can claim calls that are not its own — but the fix was never applied on the other side: the frameworks' own configs still leave many patterns unscoped, so the filter throws them away.

The doc comment on `SecondaryView` acknowledges the mount case; the parameter and request-body losses above are the same defect with a wider blast radius.

## Suggested fix

Scope the patterns in their home configs, so they survive the filter for the right reason — the same move #138 applied to chi's `Mount`/`Route`:

- gin: `Param`/`Query`/`DefaultQuery`/`GetHeader` → `^github\.com/gin-gonic/gin\.\*?Context$`; `BindJSON`/`ShouldBindJSON`/… likewise.
- echo: the three unscoped param patterns → echo's `Context`.
- fiber: the remaining unscoped mount pattern.
- chi: the remaining unscoped param pattern.

Each with a zero-drift check for single-framework projects, since scoping a pattern can only narrow what it matches.

## Related: net/http's parallel config

net/http is the one framework not filtered from its primary config — `HTTPSecondaryConfig()` (`internal/spec/config_merge.go:37`) is a separate, hand-maintained subset. That keeps it working, but the two can silently diverge: while fixing #138 a new `MountPattern` added to `DefaultHTTPConfig` had no effect in mixed projects until it was added to `HTTPSecondaryConfig` as well, which is exactly the case the fix was about.

Worth deciding whether that parallel config should remain, or whether scoping net/http's own patterns would let it go through `SecondaryView` like everything else.

## Note

This compounds with the primary-selection problem filed separately: which framework ends up secondary is decided by file order, so a project's dominant framework can be the one that gets stripped.

### LINKED ISSUE #212: Primary framework is chosen by file order, so renaming a file changes the generated spec
In a multi-framework project the *primary* framework decides which config is used in full; every other framework is reduced to its merge-safe subset. That choice is currently made by whichever framework is imported by the first file the detector happens to parse.

`internal/core/detector.go:50` `DetectAll` walks `CollectGoFiles(dir)` in order, appending each framework the first time it sees an import, and `internal/engine/engine.go:576` takes `framework := frameworks[0]` as primary — the rest go through `SecondaryView`.

Nothing weighs how much each framework is actually used.

## Reproduce

One module, two files: a 12-line chi admin router and the real gin API. **Only the filenames differ between the two runs.**

```
a_admin.go (chi) + z_api.go (gin)   ->  chi primary, gin secondary
z_admin.go (chi) + a_api.go (gin)   ->  gin primary, chi secondary
```

Same `ginCreate` handler in both:

```go
func ginCreate(c *gin.Context) {
    var req CreateReq
    _ = c.ShouldBindJSON(&req)
    _ = c.Query("filter")
    c.JSON(http.StatusCreated, req)
}
```

### chi sorts first — gin is secondary

```yaml
/gin/items:
  post:
    operationId: ginsec2.ginCreate
    responses:
      default: {description: "Default response (no response found)"}
components: {}
```

### gin sorts first — gin is primary

```yaml
/gin/items:
  post:
    parameters:
      - {name: filter, in: query, schema: {type: string}}
    requestBody:
      content: {application/json: {schema: {$ref: '#/components/schemas/ginsec2_CreateReq'}}}
    responses:
      "201":
        content: {application/json: {schema: {$ref: '#/components/schemas/ginsec2_CreateReq'}}}
```

Renaming a file changes the API documentation. Nothing in the output indicates which framework was treated as primary or that anything was dropped.

## Why it matters

On its own, an arbitrary primary would be tolerable if every framework kept its capabilities as a secondary. It does not — see #211, where a secondary gin loses all four parameter patterns and all three request-body patterns. The two defects compose: **the dominant framework of a project can silently become the secondary one and lose most of its extraction.**

It also makes the tool non-deterministic in spirit: the output depends on a filename, which is not a property of the API.

## Possible directions

1. **Rank by usage** — count route registrations (or framework-typed receivers) per framework in the metadata and make the busiest one primary. Most faithful to intent; needs the count before config selection, which currently happens first.
2. **Rank by a fixed precedence** — deterministic and trivial, but still arbitrary when a project genuinely leads with the "lower" framework.
3. **Make the merge symmetric** — if every framework's patterns were properly scoped (#211), primary-vs-secondary would stop mattering for extraction, and this reduces to picking defaults (Info, response content type). This is the most robust outcome and makes (1) unnecessary.

(3) plus #211 looks like the real fix; (1) is a narrower stopgap.

## Acceptance

- A fixture whose files are ordered so the dominant framework does *not* sort first, asserting its endpoints are documented in full.
- Renaming files in a fixture must not change the generated spec — worth a test, since this is the property that broke.

ehabterra:
Closes #211.

Every framework except the primary is reduced to its receiver-scoped subset by `SecondaryView`. The rule is sound — an unscoped pattern transplanted into another framework can claim calls that are not its own — but the frameworks' own configs never held up their end, so the filter threw away most of what makes them work.

A secondary gin kept its route pattern, so its endpoints were still *discovered*, and then documented them with no parameters, no request body and no response schema. Same handler, same code; the only thing deciding it was which framework happened to be primary.

## The fix: scope the calls that were always receiver-bound

| framework | patterns scoped | receiver |
|---|---|---|
| gin | `Param`/`Query`/`DefaultQuery`/`GetHeader`, the `Bind*` family, the status-carrying `JSON`/`String`/`XML`/… responder | `gin.*Context` |
| echo | `Param`/`FormValue`/`Cookie` | echo `Context` |
| fiber | `Mount` | fiber `App`/`Router`/`Group` |
| chi | `FormValue` | `*http.Request` |
| net/http | `FormValue`/`Cookie` | `*http.Request` |

Retention when secondary, before → after:

| | before | after |
|---|---|---|
| gin params | 0/4 | **4/4** |
| gin responses | 5/7 | **6/7** |
| echo params | 1/4 | **4/4** |
| fiber mounts | 2/3 | **3/3** |
| chi params | 5/6 | **6/6** |
| net/http params | 3/5 | **5/5** |

The gin responder is the pattern `SecondaryView`'s comment singles out as unsafe to merge, because it reads the status from arg 0 and would misread a status-less call like fiber's `c.JSON(obj)`. Scoping it to gin's `Context` is what makes it safe — rather than dropping it.

`HTTPSecondaryConfig` can now carry `FormValue`/`Cookie`: its comment deferred them "until they gain receiver scoping", and that is this change.

## Deliberately *not* scoped

The generic `Decode`/`Unmarshal`/`Encode` helpers every config repeats. A secondary losing those costs nothing, because the primary supplies its own copies — which is why gin's request bodies read 1/3 rather than 3/3 above and nothing is actually missing from the output.

## This also fixes #212 for extraction

#212 (primary framework chosen by file order) proposes three directions and calls (3) — *make the merge symmetric, so primary-vs-secondary stops mattering* — the real fix. That is what this change turns out to deliver.

Measured on a two-framework project (gin API + gorilla/mux admin router, framework setup in separate files) where **only the filenames differ**, so the sort order flips which framework is primary:

| | mux sorts first → gin secondary | gin sorts first → gin primary |
|---|---|---|
| `main` | `/products`: no query param, `default` response, no `Product` component. `/products/{kind}`: `x-warning` placeholder param, no `requestBody` | full params + `requestBody` + `201` schema |
| this branch | identical to the best case | identical to the best case |

`diff` of the two orderings on this branch is **byte-identical**, and equal to `main`'s best-case output. #212 is not closed here — the primary still decides `Info`/`Defaults`/default response content type, and it still asks for a rename-invariance test — but the part that silently dropped most of a framework's extraction is gone.

## Tests

`testdata/secondary_framework_params` — a chi admin router plus the real gin API, **ordered so gin is not primary** (the chi file sorts first, which is what makes gin secondary). Renaming the files would mask the bug, so the ordering *is* the test. The structural test asserts the secondary gin endpoint keeps its query param, path param, request body `$ref` and concrete response status.

Zero drift across the suite (goldens, determinism, every framework fixture) and on both large real projects.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved secondary-framework route and parameter analysis across Chi, Echo, Fiber, Gin, and net/http.
  * Fixed receiver scoping for form values and cookies (and related request/response extraction) so mixed framework scenarios interpret parameters from the correct source.
* **Tests**
  * Added new test coverage for secondary Gin and net/http parameter bindings, including request/response `$ref` validation.
  * Added merge verification to ensure scoped secondary patterns propagate into primary configurations correctly.


ehabterra:
Valid finding — addressed in 1edcde3.

The two patterns really had no coverage of their own: the suite would have stayed green if the merge had dropped them again. Both are only reachable *through* the merge (the net/http surface is layered under every non-net/http framework), so the coverage is split:

- **Fixture** — `testdata/secondary_framework_params` gains `GET /admin/search`, reading `r.FormValue("q")` and `r.Cookie("session")`. The cookie is what isolates the merge: chi (the fixture's primary) has a `FormValue` pattern of its own but **no `Cookie` pattern at all**, so a resolved `session` cookie parameter can only have come from the layered net/http config. Structural assertions live in a new `generator/testdata_secondary_http_params_test.go`.
- **Merge layer** — `TestMergeFrameworkConfigs` asserts the whole path for both patterns: declared in `HTTPSecondaryConfig`, scoped to `net/http.*Request` (an unscoped pattern would be filtered out again by `SecondaryView`), kept by `SecondaryView`, and present after merging into a **gin** primary that declares neither. It `t.Fatal`s if gin ever grows its own form/cookie read, since that would silently stop isolating the merged one.

Both fail against `main`'s config (where neither pattern exists), so they are real change-detectors rather than decoration. Full suite green, lint 0 issues.

Issue #143, opened 2026-07-14
Routes unreachable through composite-literal function fields (urfave/cli Action pattern — gitea)


ehabterra:
## Symptom

Running apispec on gitea produces **0 routes**. Reduced fixture: `testdata/cli_action_routes` (added in the fixtures PR) — plain chi registrations, but reached like this:

```go
app := &App{Commands: []*Command{{Name: "web", Action: runWeb}}}
_ = app.Run("web")          // Run dispatches c.Action()
```

## Diagnosis (verified on gitea itself)

- The metadata **does** record the registration edges (`(*web.Router).Post` from `routers/web` was found in gitea's call graph and string pool).
- Tracker-tree roots are **`main` functions only** (`NewLazyTree`, same for the eager tree). The chain `main → app.Run(...)` → `cli.Command{Action: runWeb}` has no static call edge from main's subtree into `runWeb`: the function value crosses through a composite-literal **field** and is invoked by the library dispatcher (`c.Action()`).
- Receiver-scoped route patterns are NOT the blocker: extending them to gitea's wrapper router (`code.gitea.io/gitea/modules/web.*Router`) still yields 0 paths, because the tree never reaches the registration functions at all.

## Possible directions

1. **Function-value field tracing**: record `Command{Action: runWeb}` as an assignment-like fact (metadata already records composite-literal args) and let tree expansion follow `c.Action()`-style dynamic calls to the recorded candidates — honest-over-wrong applies when several candidates exist.
2. **Re-rooting**: functions containing route registrations that are unreachable from any root become additional tree roots (possibly behind a flag). Simpler, but loses mount/prefix context that flows down from main, and can double-count.
3. Combination: re-root only registration clusters whose enclosing function is stored in a composite-literal function field (detectable from metadata).

The fixture's structural test currently **pins the gap** (asserts 0 paths, with a pointer here) so it fails loudly the day the capability lands and the expectations must flip to `/users` GET+POST.

Related: receiver-scope extension for wrapper routers is a separate, config-only concern (works today via `--config`).

Issue #216, opened 2026-07-25
Closure handlers get an operationId containing the absolute source path, so the spec is not reproducible across machines


ehabterra:
A route whose handler is a closure gets an `operationId` built from the function literal's **absolute** source position, so the generated spec embeds local filesystem paths and is not reproducible across machines or checkouts.

## Reproduce

Any inline handler — here `testdata/secondary_framework_params`:

```go
r.Get("/admin/health", func(w http.ResponseWriter, r *http.Request) {
	_ = json.NewEncoder(w).Encode(Health{OK: true})
})
```

### Actual

```yaml
/admin/health:
  get:
    operationId: github.com/ehabterra/apispec/testdata/secondary_framework_params.FuncLit:/Users/me/src/apispec/testdata/secondary_framework_params/a_admin_chi.go:20:25
```

### Expected

An identifier that is a property of the code, not of the machine — the package-relative position at most:

```yaml
    operationId: github.com/ehabterra/apispec/testdata/secondary_framework_params.FuncLit:a_admin_chi.go:20:25
```

## Why it matters

- **The spec leaks absolute paths** — home directories, CI runner working dirs — into a document that usually gets published or committed.
- **Same code, different spec.** Two checkouts at different paths (or a dev machine vs CI) produce different `operationId`s, so any diff-the-spec workflow reports spurious changes and nothing anchors the operation across runs. This is the same class of defect as the map-iteration nondeterminism guarded by `TestGenerateDeterministic`, except the varying input is the *checkout path* rather than iteration order — the determinism tests all run from one directory, so they cannot see it.
- Renaming or moving a file changes the `operationId` of every closure handler in it.

## Root cause

Closure "names" are minted from `getPosition`, which is `fset.Position(pos).String()` — an absolute `file:line:col` when packages are loaded from an absolute directory (the CLI's `--dir` case):

- `internal/metadata/helpers.go:72` — `getPosition`
- `internal/metadata/analysis.go:82`, `analysis.go:287`, `internal/metadata/expression.go:80` — `fmt.Sprintf("FuncLit:%s", position)`

That name is a metadata *fact* used as an identity key throughout the tracker, so the fix is not simply to rewrite the string at the spec boundary: the position needs to be recorded module- or package-relative at the point it is created (or rendered relative only where it reaches output, if any consumer depends on the absolute form).

## Notes

- Only closure handlers are affected; named handlers produce `pkg.handlerName`.
[...]

PR #219, opened 2026-07-25
feat(spec): follow route registration through func-typed struct fields (#143)

ehabterra:
**Partially addresses #143** — do not auto-close it. Branched from `main`, independent of the #215/#217/#218 stack.

This fixes func-field dispatch when the dispatcher is **in the analysed module** (proved by four fixture shapes across two engines). It does **not** fix gitea, which is the issue's headline symptom, because gitea's dispatcher is inside urfave/cli — measured and explained under "Measured on gitea" below.

apispec reported **zero routes** for gitea. The registration edges were in metadata the whole time; nothing reached them. Tracker roots are main functions, and in the urfave/cli wiring style no static call edge crosses the dispatcher hop — the registering function goes into a composite-literal field and the library calls it back as `c.Action()`, so expansion stopped at `app.Run(...)`.

| | before | after |
|---|---|---|
| `testdata/cli_action_routes` | **0 paths** | `/users` GET+POST, `/admin/stats` GET, `/metrics` GET — with request bodies, statuses and schemas |

## Approach: direction (1) from the issue

A func-typed field is a dispatch point in exactly the way an interface method is, so it is resolved the same way: collect the values recorded for it and expand into all of them. `funcFieldDispatch` (new `internal/spec/func_field_dispatch.go`) indexes `"pkg.Type.Field" -> function base keys`.

Three sources, because the same Go code is recorded three different ways depending on where the literal sits:

| source | shape it covers |
|---|---|
| `File.StructInstances` | `var CmdWeb = &cli.Command{Action: runWeb}` — **gitea's own form**; a package-level initializer is not kept as a `CallArgument` anywhere |
| assignment values | `app := &App{Commands: []*Command{{Action: runWeb}}}` — the nested literal survives in full in the CallArgument tree |
| call arguments | `app.Register(&Command{Action: runWeb})` |

I verified the nesting is genuinely present before building on it (`ExprToCallArgument` keeps `Action: runWeb` intact) — the metadata *YAML* truncates it, which is what made this look like a fact-layer gap at first glance. It isn't: nothing needed to change in `internal/metadata`.

## Both engines

- **LazyTree** — one `expandKey` loop beside the existing interface fan-out in `buildPlan`.
- **Eager tree** — beside its `ImplementedBy` attachment. This side additionally has to build the attached nodes' **argument** children itself: a node builds its own callees, but its arguments are the parent loop's job. Without that the routes appear and document *nothing* — which is exactly what my first eager version did, and why the parity check below is in the PR rather than assumed.

Both engines produce **byte-identical** specs for the fixture (checked with `--legacy-tracker`).

## Honest by construction

- A value resolves only when it **names** a function expansion can follow (`meta.Callers` / `meta.ParentFunctions`, which is also what makes an inline closure `Action` work — closures are callers in their own right). A field holding a *variable* contributes nothing rather than a guess.
- A field-name match is never enough on its own: the field must be func-typed, so `Name: something` can't be mistaken for a dispatch point.
- **Several values for one field are all kept.** Each is genuinely invoked for its own subcommand, so their union is the surface the binary serves — a stronger statement than picking one, and the reason this isn't the ambiguity golden rule #7 tells you to refuse. The fixture has three commands registering three different route sets and asserts all three appear.
- Known limitation, deliberately not guessed at: the union is per `(type, field)`, not per *instance*. If two instances of the same struct type hold different functions and are called at different mount prefixes, both expand at both sites. Narrowing that needs the call site's receiver traced back to a specific literal — value resolution, a separate piece of work. No fixture in the suite exhibits it (zero drift), and for CLI dispatch the union is the correct reading.

## Tests

- `testdata/cli_action_routes` stops pinning the gap: it now carries all three shapes and `TestTestdata_CLIActionRoutes` asserts the routes **and their handler bodies** (a route found but documented empty is barely better than a route missed). The old known-gap assertion is removed from `testdata_known_gaps_test.go`.
- `TestBuildFuncFieldDispatch` covers what the index refuses — a variable-valued field, a non-func field, a function with no recorded calls — plus multi-value ordering (sorted; this index decides which routes exist, so its order reaches the output) and positional literals. Built against the real `ExprToCallArgument` rather than hand-assembled arguments, since the nested-literal shape is the thing under test.
- `keysFor` unit-tested across pointer/qualified receivers and the four non-matches (different member, type, package, no receiver).

## Verification

Full suite green (goldens, determinism, all framework fixtures), `golangci-lint` 0 issues, `gofmt`/`vet` clean. **Zero drift**: I generated all 60 fixtures with a `main` binary and a branch binary — `cli_action_routes` is the only one that changes.

## Measured on gitea

Run against a real gitea checkout with both binaries. **Still 0 routes**, and the reason is structural rather than a tuning problem:

```
DBG index entries=18                 # the index does build, on gitea's own types
DBG edges-calling-Action=0           # ...but nothing ever CALLS Action
DBG func-typed-fields-declared=42
```

Three blockers stack up, all of them about the dispatcher being third-party:

1. **No dispatch edge exists.** `cmd.Action(ctx, cmd)` is executed inside `urfave/cli`, and external packages are never loaded (360 packages either way; `--include-package` only filters what the loader already returns). With no edge whose callee is `Action`, the expansion hook has nothing to fire on — which is why the index being correctly populated changes nothing.
2. **`cli.Command` is external**, so its field list is not in metadata and the type cannot be looked up.
3. **`Action` is typed `cli.ActionFunc`** — a *named* func type, not a literal `func(...)`, so even with the declaration present the func-typed-field test would reject it.

My fixtures hid all three by putting the dispatcher in the module, which is exactly the shape the issue warned about ("invoked by the library dispatcher"). Filed as its own issue with this evidence; the tractable fix is the issue's own direction (2)/(3) — re-root a registration cluster whose enclosing function is stored in a composite-literal field, which needs no external analysis because the *assignment* is in gitea's own code — plus dropping the field-type test in favour of "the value names a known function". That is a behaviour-risk change (new tree roots) and wants its own PR and a fixture that depends on the real `urfave/cli`, so it cannot be faked by an in-module dispatcher a second time.

## Measured performance

No regression. Warm, alternating runs on gitea:

| | main | this branch |
|---|---|---|
| wall clock | 46.5s / 46.5s | 47.0s / 49.8s |
| metadata generated | 33.3s | 33.7s |
| spec mapped | 8.4s | 9.6s |
| `buildRelations` (cum, CPU profile) | 7.04s | 7.27s |
| **`buildFuncFieldDispatch` (cum)** | — | **0.04s (0.075%)** |

The index build does not appear in the profile above the default threshold; at full resolution it is 40ms. The run-to-run spread in "spec mapped" is machine noise, not this change.

A doubling *was* observed initially (1:50 vs ~46s) and reproduced — it is the **cold first run**: 536s user time at 608% CPU, i.e. `go/packages` type-checking gitea and writing export data. Warm runs of either binary land at ~46-47s.

## What still holds

Full suite green, `golangci-lint` 0 issues, coverage 96.1% (floor 96). Zero drift across the other 59 fixtures, normalising the absolute path #216 leaks into closure operationIds.

### DIFF
--- a/generator/func_field_parity_test.go
+++ b/generator/func_field_parity_test.go
@@ -1,0 +1,86 @@
+// Copyright 2026 Ehab Terra
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package generator
+
+import (
+	"sort"
+	"testing"
+)
+
+// TestFuncFieldEngineParity runs the func-field dispatch fixtures (#143) through
+// BOTH tracker engines and requires the same routes, with the same operations and
+// response statuses, from each.
+//
+// The two engines add children in different places — the lazy tree expands by
+// key, the eager tree materializes nodes — so a capability landing in one is
+// silently absent from the other. That is not hypothetical here: the eager side
+// initially found the routes and documented them *empty*, because a node builds
+// its own callees while its ARGUMENTS are built by the parent loop, and the new
+// attachment point had to do that itself. Comparing statuses per operation is
+// what catches that class of divergence; comparing path names alone would not.
+func TestFuncFieldEngineParity(t *testing.T) {
+	for _, fixture := range []string{"cli_action_routes", "cli_action_cross_package"} {
+		t.Run(fixture, func(t *testing.T) {
+			// nil config on purpose: these fixtures rely on framework detection
+			// and the multi-framework merge, which an explicit config bypasses.
+			lazy := generateWithTracker(t, fixture, nil, true)
+			eager := generateWithTracker(t, fixture, nil, false)
+
+			lazyPaths, eagerPaths := mapPathKeys(lazy.Paths), mapPathKeys(eager.Paths)
+			sort.Strings(lazyPaths)
+			sort.Strings(eagerPaths)
+			if len(lazyPaths) == 0 {
+				t.Fatalf("lazy documented no paths at all — the fixture stopped exercising #143")
+			}
+			if len(lazyPaths) != len(eagerPaths) {
+				t.Fatalf("paths differ between engines:\n lazy:  %v\n eager: %v", lazyPaths, eagerPaths)
+			}
+			for i := range lazyPaths {
+				if lazyPaths[i] != eagerPaths[i] {
+					t.Fatalf("paths differ between engines:\n lazy:  %v\n eager: %v", lazyPaths, eagerPaths)
+				}
+			}
+
+			for _, path := range lazyPaths {
+				for _, method := range []string{"GET", "POST", "PUT", "PATCH", "DELETE"} {
+					lazyOp := opFor(lazy.Paths[path], method)
+					eagerOp := opFor(eager.Paths[path], method)
+					if (lazyOp == nil) != (eagerOp == nil) {
+						t.Errorf("%s %s: lazy=%v eager=%v", method, path, lazyOp != nil, eagerOp != nil)
+						continue
+					}
+					if lazyOp == nil {
+						continue
+					}
+					// Statuses, not schemas: the question is whether the handler
+					// body was reached at all, which is what diverged.
+					if len(lazyOp.Responses) != len(eagerOp.Responses) {
+						t.Errorf("%s %s: response count lazy=%d eager=%d",
+							method, path, len(lazyOp.Responses), len(eagerOp.Responses))
+					}
+					for status := range lazyOp.Responses {
+						if _, ok := eagerOp.Responses[status]; !ok {
+							t.Errorf("%s %s: status %q in lazy but not eager", method, path, status)
+						}
+					}
+					if (lazyOp.RequestBody == nil) != (eagerOp.RequestBody == nil) {
+						t.Errorf("%s %s: requestBody lazy=%v eager=%v",
+							method, path, lazyOp.RequestBody != nil, eagerOp.RequestBody != nil)
+					}
+				}
+			}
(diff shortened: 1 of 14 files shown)


ehabterra:
Ran this against a real gitea checkout with profiling. Two findings, one good and one that changes this PR's scope.

**Performance: no regression.** Warm, alternating runs, both binaries built from worktrees:

| | main | this branch |
|---|---|---|
| wall clock | 46.5s / 46.5s | 47.0s / 49.8s |
| metadata generated | 33.3s | 33.7s |
| spec mapped | 8.4s | 9.6s |
| `buildRelations` (cum, CPU profile) | 7.04s | 7.27s |
| **`buildFuncFieldDispatch` (cum)** | — | **0.04s (0.075%)** |

The index build does not even clear pprof's default threshold; at full resolution it is 40ms on 76,429 call edges. The spread in “spec mapped” is run-to-run noise.

The ~2x slowdown reported against this branch reproduces, but it is the **cold first run**, not the change: cold main took **1:50 at 608% CPU and 536s user time** — `go/packages` type-checking gitea and writing export data — while every warm run of either binary lands at 46-47s. Re-timing main cold would show the same doubling.

**gitea still yields 0 routes, and this PR cannot fix that.** Instrumented run:

```
DBG index entries=18                 # index builds fine over gitea's own types
DBG edges-calling-Action=0           # nothing in the graph ever calls Action
DBG func-typed-fields-declared=42
```

`cmd.Action(ctx, cmd)` executes inside `urfave/cli`, and external packages are never loaded (360 packages with or without `--include-package "github.com/urfave/cli/**"` — that flag filters the loader's output rather than extending it). So the edge the expansion hook keys on does not exist. On top of that `cli.Command` is external (no field list in metadata) and `Action` is `cli.ActionFunc`, a *named* func type that the func-typed-field test rejects.

My fixtures hid all three by keeping the dispatcher in-module — exactly the case #143 warned about. So:

- the PR body no longer says “Closes #143”; it is scoped to in-module dispatch, which the fixtures do prove on both engines;
- the real blocker is filed as #220 with this evidence and two candidate directions (re-rooting field-stored functions, which needs no external analysis, vs. loading the dispatcher's package).

Happy to attempt the re-rooting fix next — it is the tractable one, but it adds tree roots, so it wants its own PR plus a fixture that depends on the real `urfave/cli` so an in-module dispatcher cannot satisfy it by accident.

Issue #221, opened 2026-07-26
Wrapper routers: the inner delegate double-registers with unresolved path/handler, and a verb carried in an argument can't be configured


ehabterra:
Projects that wrap a router in their own type — gitea's `modules/web.Router` around `*chi.Mux` is the canonical example — can be documented today by adding receiver-scoped route patterns for the wrapper. But two defects make that config path unusable as-is, and one shape cannot be expressed at all.

## What already works (measured)

A probe replicating gitea's `modules/web/router.go` — verb methods delegating through a shared `Methods(methods, pattern string, h ...any)`, and a `Group(pattern, fn)` that accumulates the prefix in a **field** and applies it in `getPattern` rather than mounting a sub-router:

```go
func (r *Router) Methods(methods, pattern string, h ...any) {
	// ...unwrap h[len(h)-1] into an http.HandlerFunc...
	r.chiRouter.Method(methods, r.getPattern(pattern), handlerFunc)
}
func (r *Router) Get(pattern string, h ...any) { r.Methods("GET", pattern, h...) }
func (r *Router) Group(pattern string, fn func()) { prev := r.curGroupPrefix; r.curGroupPrefix += pattern; fn(); r.curGroupPrefix = prev }
```

With a config carrying **only** a wrapper verb pattern (`^(?i)(Get|Post|Put|Delete|Patch|Head)$` scoped to the wrapper receiver, `pathFromArg`/`handlerFromArg`), the output is fully correct:

```yaml
/users:      get -> listUsers  (200, []User)     post -> createUser (requestBody User, 201 User)
/api/items:  get -> getItem    (200, User)        # group prefix from the FIELD, resolved
```

Handlers passed as plain funcs into the variadic `...any` resolve to their bodies, so request bodies, statuses and schemas all come through. That is worth stating plainly: **the wrapper-router shape itself is supported.** No new capability is needed for it.

## Defect 1 — the inner delegate double-registers, and wins (on the DEFAULT path)

This is not an artifact of hand-writing a config. With plain auto-detection — chi as primary, net/http merged as a scoped secondary, exactly as the multi-framework machinery intends — the probe above produces **one junk route and none of the three real ones**:

```yaml
/{path}:
  post:
    operationId: example.com/wrap.net/http.HandlerFunc
    parameters: [{$ref: '#/components/parameters/PathParam'}]   # synthesized, #34
    responses: {default: {...}}                                  # no body
```

[...]

Issue #220, opened 2026-07-26
Func-field route registration is unreachable when the dispatcher is third-party (gitea/urfave-cli still yields 0 routes)


ehabterra:
Follow-up to #143. PR #219 makes apispec follow a function value out of a composite-literal field into the registering function — but only when the **dispatcher** is in the analysed module. gitea, the project #143 was filed about, still generates **0 routes**, and the reason is structural.

## Measured on a real gitea checkout (PR #219 build)

```
DBG index entries=18                 # the func-field index builds fine, over gitea's own types
DBG edges-calling-Action=0           # ...but NOTHING in the graph ever calls Action
DBG func-typed-fields-declared=42
```

The index is populated and correct (`git.RunOpts.PipelineFunc`, `doctor.Check.Run`, `migration.ReleaseAsset.DownloadFunc`, …). It changes nothing, because the call it would answer does not exist.

## Why

gitea wires routes exactly as urfave/cli intends:

```go
// cmd/web.go
var CmdWeb = &cli.Command{
	Name:   "web",
	Action: runWeb,          // runWeb registers every HTTP route
}
```

Three separate blockers, all of them "the dispatcher is third-party":

1. **No dispatch edge.** `cmd.Action(ctx, cmd)` runs inside `github.com/urfave/cli/v3`. External packages are never loaded — 360 packages are analysed with or without `--include-package "github.com/urfave/cli/**"`, because that flag filters what the loader already returned rather than adding to it. So there is no edge whose callee is `Action`, and any expansion hook keyed on such an edge is unreachable by construction.
2. **`cli.Command` is external**, so its field list is absent from metadata: the type cannot be looked up to learn that `Action` is a function-typed field.
3. **`Action` is `cli.ActionFunc`** — a *named* func type, not a literal `func(...)`. Even with the declaration in hand, a "is this field func-typed?" test that keys on the rendered type string rejects it.

The single-module fixtures in #219 (`testdata/cli_action_routes`, `testdata/cli_action_cross_package`) hide all three, because their dispatcher is in the module. That is the shape #143 explicitly warned about ("invoked by the library dispatcher"), and it is worth being blunt that the fixtures were not evidence about gitea.

## Directions

[...]

PR #222, opened 2026-07-27
feat(spec): resolve routes behind a CLI dispatcher via entrypoint patterns (#220)

ehabterra:
**Refs #220** (leaves it open — gitea needs #221 too). Stacked on **#219**: base is `feature/143-function-value-fields`, so the diff here is just this commit.

Measured on the three real projects:

| project | before | after | remaining blocker |
|---|---|---|---|
| **photoprism** | 0 paths | **107 paths** | group prefix comes from `conf.BaseUri(…)`, a runtime call, so paths are unprefixed |
| **gitea** | 0 paths | 2 paths (`/debug/fgprof` real, `/string` junk) | **#221** — its wrapper router |
| every existing fixture | — | **zero drift** (61 fixtures) | — |

## The problem

A server-shaped Go program starts its router from a subcommand, and the call that invokes that subcommand is inside the CLI library:

```go
var CmdWeb = &cli.Command{Action: runWeb}   // gitea, photoprism
```

apispec never analyses external packages, so **no edge in the graph has `Action` as its callee** — #219's "follow a func-typed field to its value" has nothing to follow. The registration subtree is unreachable from main and the project documents nothing.

## The mechanism: `EntrypointPattern`

> *a function stored in this field is invoked by something outside this module — root it when nothing else reaches it*

The enabling fact is that the **literal** is written in the analysed module, so metadata records its type (`github.com/urfave/cli/v2.Command`) and the field's value even though urfave/cli is never loaded. **Nothing from the external declaration is needed** — which is what makes this possible at all, since `cli.Command` has no recorded field list and `Action` is the named type `cli.ActionFunc` rather than a literal `func(...)`. Both of those blocked the earlier approach.

Config-driven with import-keyed presets (golden rule #5), riding the same preset merge as the auth libraries and the scoped net/http secondary:

| library | fields |
|---|---|
| `urfave/cli` v1 / v2 / v3 (+ its `codegangsta` name) | `Action`, `Before`, `After` on `Command`/`App` |
| `spf13/cobra` | `Run`, `RunE`, `PreRun(E)`, `PostRun(E)`, `PersistentPreRun(E)` |
| `peterbourgon/ff` (ffcli) | `Exec` |

A house dispatcher declares its own field in config and needs no code change; a project importing none of these pays nothing. `mitchellh/cli` (factory map) and `kong`/`subcommands` (interface methods) are documented as deliberately out of scope in the preset file — the latter already resolve through the existing interface fan-out.

## Two gates, both load-bearing

1. **Already reachable → left alone.** Otherwise its routes document twice.
2. **Must be able to reach a route registration**, answered by a bottom-up pass over the SCC condensation — the same shape `Extractor.reachSet` already uses for param accessors and middleware.

Gate 2 is what keeps the cost proportional to routing code rather than to subcommand count, and the numbers show why it matters:

```
photoprism:  53 declared, 1 rooted  (0 already reachable, 52 register no routes)
gitea:       49 declared, 1 rooted  (0 already reachable, 48 register no routes)
```

**The gate honours each route pattern's receiver scope, not just its call regex.** I measured name-only first: gin's `^(?i)(GET|POST|…)$` matches `Get` on a cache or a config, which rooted **24 of photoprism's 53** entrypoints. Honouring the receiver roots exactly 1, with byte-identical output. One trap worth recording: the receiver has to be built the way the extractor's matchers build it (`pkg + "." + recvType`) — comparing the raw pooled string rejected every real router call and quietly produced 0 paths.

## One non-obvious fix

`LazyTree.GetRoots()` now calls `buildRelations()`. Entrypoint roots come out of the same walk that builds the func-field index, and **the extractor asks for roots before it expands anything** — so deferring left the extra roots invisible: the tree had them, nobody ever saw them. It is memoized, so this only moves work every expansion path already pays.

## Diagnostics

`--verbose` reports `Entrypoints: N declared, M rooted (X already reachable, Y register no routes)`. Bounded is fine; silent is not.

## Tests

- **`testdata/cli_entrypoint_routes` depends on the real `github.com/urfave/cli/v3`.** That is the point: an in-module dispatcher cannot satisfy it by accident, which is exactly how `testdata/cli_action_routes` passed while gitea stayed at zero. It also carries a route-less second command as the control for gate 2.
- `TestEntrypointRoots` covers both gates (unreachable+registers → root; registers but reachable → skipped; unreachable but route-less → skipped; unknown key → skipped).
- `TestRouteRegistrationMatcher` pins the receiver check, including that `cache.Get` must **not** count, plus exact-`RecvType`, mount patterns, the unscoped name-only fallback, and nil inputs.
- `TestApplyEntrypointPresets` covers every preset by import (cobra included — it has no fixture because the dependency is not vendored), idempotence, and that a user's own pattern keeps precedence.

## Cost

photoprism 25s → 47s, gitea 46s → 70s. That is the price of expanding a route subtree that previously did not exist — photoprism's mapping stage goes 0.4s → 25s and yields the 107 paths. On gitea the same expansion currently yields little because of #221; once that lands the same cost produces its full surface. Projects with no entrypoint imports are unchanged (zero drift, and the presets are inert).

## Not covered

photoprism's `/api/v1` prefix: the group prefix is `conf.BaseUri(config.ApiUri)`, resolved at runtime, so paths come out unprefixed. Worth its own issue — the const half (`config.ApiUri = "/api/v1"`) is statically knowable and could be recovered by argument tracing.

## Summary by CodeRabbit

- **New Features**
  - Improved route extraction for CLI/dispatcher-based entrypoints using configurable entrypoint patterns and automatic framework preset application.
  - Framework detection/composition now uses composed multi-framework configs with primary selection stability; API detection returns an ordered list and generation honors it.
  - Added entrypoint-based rooting for both lazy and eager extraction modes.
- **Bug Fixes**
  - Prevented unintended route expansion from additional CLI commands.
  - More accurate route-registration matching with receiver- and mount-scoped behavior.
- **Tests**
  - Added/updated regression tests for CLI/cobra routing, preset application, root selection, and dispatch matching.
- **Documentation**
  - Updated README and added expanded analysis documentation.


ehabterra:
Both outstanding CodeRabbit findings were valid and are fixed — both were mine, and both are cheap.

**1. Duplicate `typeStringOf` on the identifier path** (4175541). It was called twice back-to-back for the same type — once for the prefix trim, once for the `"untyped nil"` check. Even a cache hit costs an RLock plus a map lookup, and this is the hottest path in the walk. Hoisting is also strictly fewer calls than before: two when `typeStr` was empty, one otherwise; now always one.

**2. Framework detection running twice per run** (84d3d5a). The engine called `DetectAll` to report the detected set, then `ComposeFrameworkConfig` detected again to compose from it — an import scan of every Go file, twice. That was a consequence of writing the composition helper for the UI (which has no prior detection) and reusing it wholesale in the engine. `ComposeFrameworkConfigFrom` is now the composition without the detection, so the engine composes from the list it already has and the UI keeps the detecting variant. The composition itself is unchanged, so the UI/engine divergence this PR fixed stays fixed.

Verified: output **byte-identical** on both large projects after each change; wall clock 10.5s → 10.1s and 38.5s → 37.6s. Full suite green, lint 0 issues, coverage 96.0%.

For context on the earlier finding in this PR (the unconditional `break` in the eager entrypoint-root loop): also valid, fixed in the rewritten branch history — a nil `NewTrackerNode` no longer drops the entrypoint and every route below it.

PR #223, opened 2026-07-27
fix(ui): compose the same multi-framework config the engine does

ehabterra:
**Stacked on #222** (`feature/220-entrypoint-dispatch`) — the diff here is just this commit. Fixes a UI bug that is independent of, and older than, that work.

## The symptom

apispecui documented **zero routes** for photoprism while the CLI documented **107** from the same code, same branch, same machine.

## The cause

`buildAPISpecConfig` built the config from the detected **primary alone**:

```go
if req.Framework == "" { req.Framework = "net/http" }   // "unspecified" meant stdlib
cfg := defaultConfigForFramework(req.Framework)          // ONE framework
```

and handed it to the engine — which, given an explicit config, skips detection, the scoped-secondary merge and the net/http layer entirely.

This is **issue #212's failure mode reaching the UI**. photoprism's primary detects as `mux` because one file — `docker/dummy/oidc/app/server.go`, a dummy OIDC app — imports `gorilla/mux` and sorts first in file-walk order. Every route photoprism actually serves is gin. The CLI is immune because it merges gin back in as a receiver-scoped secondary (#211/#212); the UI, holding a mux-only config, could not recognise a single one.

Measured:

| config the UI sent | paths |
|---|---|
| `mux` only (what it built) | **0** |
| `net/http` only (what an untouched selector produced) | **0** |
| composed: primary + scoped secondaries + net/http | **107** |

## The fix

`ComposeFrameworkConfig(dir)` now owns the composition, and **both** the engine and the UI call it. One path, one answer, no way to drift apart again.

`ComposeFrameworkConfigWithPrimary(dir, primary)` serves the framework selector: the user picks which framework *leads*, and the others still merge under it — picking one must not discard the rest. An explicit choice detection never saw is added rather than rejected, and the returned list always leads with the primary in effect, since the UI reads `frameworks[0]` as the effective one.

`DetectResponse` gains `detectedFrameworks`, so the UI can show the whole set instead of a misleading single name. An unspecified framework now means *detect it* rather than net/http.

## Verified through the UI's own API

Not just the library — the actual HTTP path the browser uses:

```
POST /api/generate {"info":{"title":"photoprism","version":"1.0.0"}}
  ->  {"ok":true,"framework":"mux","pathCount":107}      # was pathCount: 0
GET  /api/detect
  ->  {"detectedFramework":"mux","detectedFrameworks":["mux","gin"]}
```

## Tests

`TestComposeFrameworkConfig` asserts a mixed gin+mux project composes patterns that can recognise **both** frameworks plus net/http — the property whose absence caused this. `TestComposeFrameworkConfigWithPrimary` covers the selector: for each of `gin`, `mux` and `chi` as primary (including a framework the project does not use), both detected frameworks stay recognisable and the primary leads. `TestComposeFrameworkConfigSingleFramework` pins that the ordinary one-framework case is unchanged.

Full suite green, `golangci-lint` 0 issues.

## Note

This is why #222 looked broken when tested in the UI: photoprism resolves 107 paths there only once *both* changes are present — #222 makes the CLI-dispatched routes reachable, and this makes the UI analyse them with the right framework patterns.

### DIFF
--- a/cmd/apispecui/main.go
+++ b/cmd/apispecui/main.go
@@ -34,7 +34,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/ehabterra/apispec/internal/core"
 	"github.com/ehabterra/apispec/internal/diagserver"
 	"github.com/ehabterra/apispec/internal/engine"
 	"github.com/ehabterra/apispec/internal/insight"
@@ -118,10 +117,16 @@ type ServerConfig struct {
 // DetectResponse is what GET /api/detect returns: information the UI needs
 // to pre-fill the configuration form.
 type DetectResponse struct {
-	InputDir            string                         `json:"inputDir"`
-	ModuleRoot          string                         `json:"moduleRoot"`
-	ModulePath          string                         `json:"modulePath"`
-	DetectedFramework   string                         `json:"detectedFramework"`
+	InputDir          string `json:"inputDir"`
+	ModuleRoot        string `json:"moduleRoot"`
+	ModulePath        string `json:"modulePath"`
+	DetectedFramework string `json:"detectedFramework"`
+	// DetectedFrameworks is every framework found, in first-seen order. The
+	// first is the primary; the rest are merged as receiver-scoped views. The
+	// UI shows the whole set because a project's primary is decided by
+	// file-walk order (issue #212), so naming only the primary is misleading —
+	// photoprism reports `mux` while serving gin.
+	DetectedFrameworks  []string                       `json:"detectedFrameworks,omitempty"`
 	SupportedFrameworks []string                       `json:"supportedFrameworks"`
 	OpenAPIVersion      string                         `json:"openapiVersion"`
 	Info                spec.Info                      `json:"info"`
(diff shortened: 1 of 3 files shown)


ehabterra:
Two more gaps, both surfaced by testing in the actual browser rather than via curl — pushed in 814015f.

**1. `/api/default-framework` still handed back ONE framework's patterns.** This is the one that kept photoprism at zero even with the first commit: the UI posts whatever is in its pattern editors straight back, and those *replace* the composed lists server-side, so touching the framework selector silently undid the merge. It now composes for the current project with the chosen framework leading, exactly like `/api/detect` and the engine.

Verified through the endpoints the browser actually calls:

```
GET  /api/default-framework?framework=mux
  -> patterns=9, recognises gin AND mux
POST /api/generate {framework:"mux", frameworkConfig:<that>}
  -> {"framework":"mux","pathCount":107}     # was 0
```

**2. The label.** The form said `② Framework (detected): mux` for a project that serves gin — it only ever showed the primary, which is decided by file-walk order (#212), so it reads as a claim about the project. It now lists the rest underneath: *"Also detected: gin — merged in, so their routes are documented too. The selection only decides which one leads."*

Worth noting what this means for gitea: `/api/detect` reports `["chi"]` there, which is correct — its remaining zero is #221 (the `web.Router` wrapper), not this.

PR #239, opened 2026-07-28
feat(spec): the declared caller/callee pattern filters actually filter (#238)

### LINKED ISSUE #238: Four declared pattern filters are dead config surface: CallerPkgPatterns, CallerRecvTypePatterns, CalleePkgPatterns, CalleeRecvTypePatterns
Split out of #221, which is otherwise finished (#232).

Every pattern type in `internal/spec/config.go` declares four filter fields:

```go
CallerPkgPatterns      []string
CallerRecvTypePatterns []string
CalleePkgPatterns      []string
CalleeRecvTypePatterns []string
```

They are documented as "Package/type filtering". **No code reads them** — `grep -rn "CallerRecvTypePatterns" --include=*.go .` returns only the struct declarations and the tests/fixtures that set them. A user configuring a pattern reaches for exactly this field, sets it, and gets silence: no filtering, no warning, no error.

`cmd/apispecui/config_parity_test.go` currently exempts them from the config-editor parity check on that ground ("never read by any matcher"), which is an honest exemption but a standing one.

#221 needed a way to say "ignore the delegate registration inside the wrapper"; honouring these filters was direction 1 there. It shipped via direction 2 instead — `completesSameRegistration` prefers the outermost registration on a chain, so the wrapper case needs no config at all. That makes this a config-surface question rather than a blocker, but it is still four documented knobs that do nothing.

## Directions

1. **Implement them** in the pattern matchers: caller filters test the enclosing function's package / receiver type, callee filters the call target's. This is the option that keeps the documented behaviour and gives a genuine escape hatch when the automatic outermost-registration rule guesses wrong.
2. **Delete them** and say so in the README's config reference. Cheapest, and honest, but removes the escape hatch and is a breaking change for any config that sets them (they parse today and are silently ignored, so nothing behaves differently — only validation would newly reject them).

(1) looks right: the fields are already in every pattern struct and in the UI's config model, and the machinery to answer "what is the caller's receiver type" exists (`param_origin.go`, `receiverOriginTypes`).

## Acceptance

- Either the four fields filter matches — with a fixture where a pattern matches only because of a caller/callee filter, and one where it is excluded only because of it — or they are removed from the schema, the UI model and the docs, and the parity-test exemption goes with them.
- Whichever way it goes, `config_parity_test.go` must no longer need the exemption.

ehabterra:
Closes #238.

Every pattern type has declared `callerPkgPatterns`, `callerRecvTypePatterns`, `calleePkgPatterns` and `calleeRecvTypePatterns` since the config was written, documented as "Package/type filtering". **No code read them.** A user set one, got no filtering, no warning and no error, and had no way to tell.

Direction 1 of the two the issue listed: implement them, rather than delete them.

## Semantics

All six matchers (route, request body, response, parameter, mount, security) now apply them:

- each list is a set of **alternatives** — any entry admits the call;
- the four fields are **ANDed** with each other, so a callee filter cannot undo a caller filter;
- an empty list constrains nothing (every built-in framework config leaves all four unset — zero output drift across all 71 fixtures);
- a regex that fails to compile matches **nothing** rather than everything: a typo should narrow a filter to nothing visible, not silently widen it;
- a receiver-less function is addressed by its package, the same convention `recvTypeRegex` already uses.

The callee side overlaps with `recvTypeRegex` on purpose — it is the list form of the same question, for admitting several unrelated owners without one alternation. The **caller** side is the capability that had no other expression.

## Why the fixture looks like that

`testdata/pattern_caller_scope` registers routes from two packages through the same chi router, with the identical calls. Nothing about the call separates them; only where it is made does.

Both directions are asserted against that one fixture — first that the unfiltered config documents everything, then that adding the filter drops exactly the operator routes. An assertion that something is absent proves nothing on its own, and the two runs differ in one field, so whatever the second drops was dropped by the filter. Verified the test fails without the implementation.

```yaml
routePatterns:
  - callRegex: ^(?i)(Get|Post|Put|Delete)$
    recvTypeRegex: ^github\.com/go-chi/chi(/v\d)?\.\*?(Router|Mux)$
    callerPkgPatterns: [/internal/api$]   # operator endpoints stay out of the spec
```

## One thing worth stating

They are **include** filters. RE2 has no negative lookahead, so there is no "everything except"; a pattern that must avoid one caller names the ones it wants. That is why #221 could not have been solved this way and took the outermost-registration rule instead — this PR does not reopen that.

## Also

- The UI offers all four on every pattern that declares them, one regex per line, and `config_parity_test.go` no longer needs its exemption. Verified in the browser end to end: typing `/internal/api$` into one route pattern's caller filter takes the fixture from 2 paths to 1, dropping `/debug/stats` and keeping `/users` with its bodies.
- Documented in `docs/CONFIGURATION.md` (new subsection with the field table) and summarised in the README.

## Verification

- New: `TestMatchCallScope` (13 cases incl. OR/AND semantics, uncompilable regex, empty entry, no-edge, receiver-less owner), `TestCallScopeIsAppliedByEveryMatcher` (one matcher left unwired is exactly this issue's failure mode), `TestTestdata_PatternCallerScope`.
- Full suite green; **zero drift** across all 71 fixtures A/B against `main`; `golangci-lint` 0 issues; `go vet` clean; coverage 96.0% (floor 96%).

## Summary by CodeRabbit

* **New Features**
  * Added caller and callee scope filters for framework detection patterns.
  * Filters can target caller or callee packages and receiver types using regex lists.
  * Updated the configuration editor to support multi-line scope filter entries.
* **Documentation**
  * Added configuration guidance, matching behavior details, and YAML examples for scoped patterns.
* **Bug Fixes**
  * Framework matches now respect configured call-site scope constraints.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@
 - [How It Works](#how-it-works)
   - [The pipeline, step by step](#the-pipeline-step-by-step)
 - [Configuration](#configuration)
+  - [Scoping a pattern to where the call is made](#scoping-a-pattern-to-where-the-call-is-made)
   - [Automatic wrapper detection](#automatic-wrapper-detection)
 - [Programmatic Usage](#programmatic-usage)
 - [Performance & Limits](#performance--limits)
@@ -726,6 +727,31 @@ framework:
     - callRegex: ^GetHeader$
       paramIn: header
 ```
+
+### Scoping a pattern to where the call is made
+
+Every pattern type also accepts `callerPkgPatterns`, `callerRecvTypePatterns`,
+`calleePkgPatterns` and `calleeRecvTypePatterns` — lists of regexes narrowing a
+pattern by *where a call is made*, not only by what is called. Two packages can
+register routes with the identical call, and that is the only fact separating
+them:
+
+```yaml
+framework:
+  routePatterns:
+    - callRegex: ^(?i)(Get|Post|Put|Delete)$
+      recvTypeRegex: ^github\.com/go-chi/chi(/v\d)?\.\*?(Router|Mux)$
+      methodFromCall: true
+      pathFromArg: true
+      handlerFromArg: true
+      handlerArgIndex: 1
+      callerPkgPatterns:
+        - /internal/api$      # operator endpoints stay out of the spec
+```
+
+Any entry in a list admits the call, an empty list constrains nothing, and all
+four are include filters. See
+[`docs/CONFIGURATION.md`](docs/CONFIGURATION.md#scoping-a-pattern-to-where-the-call-is-made).
 
 ### Entrypoints (CLI-dispatched services)
 
--- a/cmd/apispecui/assets/js/config.js
+++ b/cmd/apispecui/assets/js/config.js
@@ -602,9 +602,7 @@ function fieldEditor(p, f, set) {
     // Stored as a list, edited one entry per line — an empty box means "no
     // constraint", which is why blank lines are dropped rather than kept.
     return html`<div class="field">
-      ${lbl}<textarea class="input" style="min-height:54px" onInput=${(e) => set(toLines(e.target.value))}>
-${lines(val)}</textarea
-      >
+      ${lbl}<textarea class="input" style="min-height:54px" value=${lines(val)} onInput=${(e) => set(toLines(e.target.value))}></textarea>
     </div>`;
   }
   if (type.startsWith("select:")) {
--- a/cmd/apispecui/assets/js/config.js
+++ b/cmd/apispecui/assets/js/config.js
@@ -494,6 +494,16 @@ const COMMON_MATCH = [
   ["functionNameRegex", "Function name regex", "text", "Optional — match by the ENCLOSING function's name instead of the call (e.g. only routes registered inside RegisterRoutes)."],
 ];
 
+// Where a call is MADE, as opposed to what is called. One regex per line; any
+// one of them admits the call, an empty box constrains nothing, and all four are
+// include filters (there is no 'everything except').
+const SCOPE_FILTERS = [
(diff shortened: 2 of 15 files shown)


ehabterra:
Both addressed in `850594b`, though the first one's stated mechanism did not hold up.

**Controlled textarea** — applied, but the reported failure does not reproduce. I tested the scenario that would expose stale state: seeded two route patterns with distinct `callerPkgPatterns` (`AAA_first_pattern`, `BBB_second_pattern`), then deleted the first so the second's values shift into the first's DOM node. The surviving pattern showed `^SECOND$` / `BBB_second_pattern` — the updated values, not the stale ones. Preact maps textarea children to the value and re-renders them, so external state changes were already reflected (a "Reset to framework defaults" cleared a seeded box too).

Applied anyway, on a different rationale: `value=` is what the other three branches of `fieldEditor` use, so the branch now matches its neighbours, and it removes a real if minor divergence — `toLines` trims and drops entries, so an uncontrolled box can display text that is not in the config.

Worth noting the pre-existing `area()` helper in the same file uses the children form for the request/response context and handler-interface-method editors. It is not changed here; by the evidence above it is not broken either.

**Duplicated `fqRecvType` blocks** — good catch, and there were **seven**, not six: `edgeMatchesAccessor` builds the same string from raw metadata rather than through the context provider. `fqOwner` now delegates to a string-level `qualifyOwner` so both callers share one implementation, and all seven sites use it.

Re-verified after both: full suite green, **zero drift** across all 71 fixtures (diffed against the pre-refactor snapshots, not just against `main`), `golangci-lint` 0 issues, coverage 96.0%. The UI was re-checked in the browser — seeded values render, and the delete-a-pattern case still updates correctly.

Issue #264, opened 2026-07-29
The node budget is global and spent depth-first, so improving the call graph makes the spec worse


ehabterra:
Split out of #224. The gates on #248 and #249 say they are "blocked on #224", but #224 is about a *different* limit (`MaxInstancesPerKey`, the per-scope copy cap) and its stated mechanism does not reproduce — see https://github.com/ehabterra/apispec/issues/224#issuecomment-5120599736. This is the limit they are actually blocked on.

## The defect

`MaxNodesPerTree` is **one budget for the whole walk**, spent depth-first, with no notion of which subtree leads to a route registration. Whatever is expanded first consumes it.

On a ~900-route project that means the run ends with:

```
[engine] expansion truncated at the 50000-node limit — the spec is incomplete
[engine] spec mapped (12 paths)
```

**12 paths out of ~900.** The budget is gone inside `modules/setting` and `modules/log` before the walk reaches most routing code.

## Why it blocks everything downstream

The failure is not "the budget is too small". It is that the budget's *unit* and *order* make call-graph accuracy actively harmful — and it bites from **both** directions:

| change | direction | measured effect |
| --- | --- | --- |
| #248 — stop recording builtins (`pkg.make`, `pkg.append`) as callees | **removes** junk edges | mapping 10s → 37s, peak RSS 3.0GB → 6.7GB, **12 paths → 1** |
| #249 — resolve cross-package receivers correctly (aliases, inline interfaces) | **adds** real edges | a 100k-node run 2m06 → **>10 min** |

Opposite changes to the graph, same outcome. #248 is the clearest illustration: builtins are cheap leaf *keys*, `MaxNodesPerTree` counts keys, so deleting them frees budget the walk immediately spends expanding roughly four times as many real nodes — in the wrong place.

Both fixes are correct and both are currently `const … = false` in the source for exactly this reason. So is any future improvement to the graph, including making `--resolve-call-graph` the default: measured, resolution nearly doubles the documented surface (+76% spec content) *at the same 12 paths*, and it would pay off far more if those 12 were 900.

#247 made the two quantities honest (`NodesBuilt` = distinct callee keys, `NodesMaterialized` = nodes actually created; a 246-route service builds 208,394 nodes across 20,198 keys) but deliberately left the budget on keys, because switching it to nodes — though truthful — starves route discovery even harder: that project documents ~900 paths at a raised key budget and **1** path at a node budget of fifteen million.

## What the fix is not

[...]

Issue #224, opened 2026-07-27
The instance cap counts per argument-ancestor scope, not per route: an argument node above many handlers makes them share one allowance


ehabterra:
On a large chi service (~330 routes) that registers routes inside `r.Route("/x", func(r chi.Router){ … })` group closures and writes every response through one house helper, **274 of 350 success responses (78%) come out with no schema** — `content: {application/json: {}}` — while the error responses on the very same operations resolve correctly.

Raising one hardcoded constant fixes all of them:

| `maxInstancesPerKey` (internal/spec/lazytree.go) | 2xx **with** schema | 2xx **missing** | wall clock |
|---|---|---|---|
| 10 (current) | 76 | **274** | 14.2s |
| 400 | **390** | **0** | 20.5s |

## Why

`instanceScope()` walks up to the nearest **argument** node:

```go
func (n *LazyNode) instanceScope() string {
	for cur := n; cur != nil; cur = cur.parent {
		if cur.isArgument {
			return cur.key
		}
	}
	return ""
}
```

`maxInstancesPerKey`'s comment describes the intent as *"node copies of the same callee WITHIN one instance scope (the subtree of the nearest argument-node ancestor — approximately 'per handler')"*, and reasons about the trade-off on that basis: *"a response helper shared by every handler legitimately needs one copy per route for per-route value tracing"*.

With **group-closure routing that assumption breaks**. The closure passed to `r.Route(...)` is itself an argument node, so it — not the handler — becomes the scope for every route inside the group. A helper reached by all of them is therefore materialised **10 times per group**, not 10 times per route, and every route after roughly the tenth in its group silently loses its response body. The exact thing the comment says must keep working is what stops working.

That matches the observations:

- the operations that *do* resolve are the ones whose payload flows through **their own closure argument** (`h.serve(w, r, func(ctx …) (any, error) { return h.svc.Cashflow(…) })`) — each closure is its own scope, so each gets its own budget;
- error responses resolve everywhere, because their payload is a literal at the immediate call site and needs no deep node;
- the failures are spread across groups rather than concentrated in one shape.

## What it is NOT

Worth recording, because I chased both first:

[...]

PR #267, opened 2026-07-29
fix(spec): a group must not open a per-route budget scope — main is currently dropping routes

ehabterra:
**Urgent.** `main` currently loses most routes on any project that groups them with `chi.Route` / `Mount`. #266 was merged at `fce6cd1`, which is *before* this fix commit, so the defect is live on `main`.

| | paths | components |
| --- | --- | --- |
| v0.5.4 | 334 | 216 |
| **`main` today** | **35** | **24** |
| this PR | 334 | 203 |

A second project: 246 paths on v0.5.4, **32** on `main`, 246 with this PR.

## Cause

The per-route budget opened its scope using `RouteRegistrationMatcher`, which deliberately matches mounts and groups as well as routes — *"does this subtree lead to a route?"* has to say yes for `r.Route("/api", …)`, or the whole group is written off.

Reusing that predicate to open a **budget scope** meant every route inside one group shared a single 20,000-node allowance. It was spent inside the first few routes and the rest were never discovered. The diagnostic said so plainly:

```
per-route limit (20000) truncated 69 of 6 route subtrees
  (first at github.com/go-chi/chi/v5.Mux.Route@…/router.go:259:2)
```

`Mux.Route` — a group, not a route. This is the #224 mistake in a new place: a limit meant for a route, applied to a group.

## Fix

Two questions, two predicates. `RouteRegistrationMatcher` keeps mounts, for reachability and the entrypoint gate. A new `TerminalRouteMatcher` matches only a single route registration, and only that opens a scope.

`TestGroupCallDoesNotOpenARouteScope` is the regression guard.

Also fixes the report, which read `truncated 69 of 6 route subtrees`: a truncated subtree is re-entered many times as the walk unwinds, so the count has to be per scope, not per blocked attempt.

## Revised numbers for #264 overall

The ~900-route project moves 12 → **181** (not the 210 reported on #266 — part of that was itself an artefact of scoping on mounts). The 246-route and 334-route projects are unchanged from v0.5.4 on paths.

## Why the fixture suite did not catch it

All 74 fixtures passed while three real projects lost 85–90% of their paths: the fixtures are too small for any budget to fire. That gap is real and I am adding a fixture with enough routes in one group to exhaust a low `--max-nodes-per-route`, asserted at a cap where the unfixed code starves. Filed as a follow-up rather than held here, because `main` is broken now.

## Verification

```
go test ./...          # clean, zero fixture drift
go vet ./...           # clean
golangci-lint run      # 0 issues
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected route matching so grouped or mounted routes are distinguished from individual route registrations.
  * Improved route budget tracking to apply independently to each registered route.
  * Fixed route truncation counts so repeated processing of the same route is counted only once.
  * Improved identification of the first route affected by truncation.

### DIFF
--- a/internal/engine/engine.go
+++ b/internal/engine/engine.go
@@ -795,7 +795,8 @@ func (e *Engine) GenerateOpenAPI() (*spec.OpenAPISpec, error) {
 		tree = intspec.NewLazyTree(meta, limits,
 			intspec.WithHandlerInterfaceMethods(apispecConfig.Framework.HandlerInterfaceMethods),
 			intspec.WithEntrypoints(apispecConfig.Framework.EntrypointPatterns,
-				intspec.RouteRegistrationMatcher(apispecConfig, meta), NewVerboseLogger(e.config.Verbose)))
+				intspec.RouteRegistrationMatcher(apispecConfig, meta), NewVerboseLogger(e.config.Verbose)),
+			intspec.WithTerminalRouteMatcher(intspec.TerminalRouteMatcher(apispecConfig, meta)))
 		e.reportPhase("tracker tree ready (lazy)", time.Since(tTree))
 	} else {
 		tree = intspec.NewTrackerTree(meta, limits, NewVerboseLogger(e.config.Verbose),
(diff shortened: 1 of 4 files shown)


ehabterra:
Superseded by #268. The group-scope fix here is correct, but two further defects make part 2 unshippable as designed: the per-route default silently deletes response schemas on projects that were previously fine, and — decisively — part 2 can produce the WRONG request-body DTO, which raising the cap does not fix. Reverting is the right call; the findings move to #264.

Issue #275, opened 2026-07-31
A Mount performed inside a helper function loses its prefix


ehabterra:
`r.Mount(prefix, server)` applies the prefix correctly when it is written at the registration site, and silently loses it when the same call is made one function deeper. The routes are still documented — at the wrong paths, without their prefix.

This matters for any project that centralises mounting in a small helper, which is the normal shape once a codebase has more than a couple of modules (and is what a DI-wired project generates).

## Reproduction

Same fixture in both cases; only the location of the `Mount` call differs.

**Works** — mounted at the call site:

```go
func main() {
	root := chi.NewRouter()
	root.Mount("/api", Server())
	_ = http.ListenAndServe(":8080", root)
}
```
→ `/api/things` ✓

**Loses the prefix** — mounted inside a helper:

```go
func MountServer(r *chi.Mux, server http.Handler) bool {
	r.Mount("/api", server)          // literal prefix, right here
	return true
}

func main() {
	root := chi.NewRouter()
	MountServer(root, Server())
	_ = http.ListenAndServe(":8080", root)
}
```
→ `/things` ✗

## What it is not

The prefix is **literal and local** in the failing case, so this is not about resolving a value:

| helper signature | prefix | documented |
| --- | --- | --- |
| mounted directly in `main` | `"/api"` literal | `/api/things` ✓ |
| `MountServer(r, server)` | `"/api"` literal inside the helper | `/things` ✗ |
| `MountServer(prefix string, r, server)` | plain `string` parameter | `/things` ✗ |
| `MountServer(prefix RoutePrefix, r, server)` | named type, `r.Mount(string(prefix), server)` | `/things` ✗ |

All three helper forms fail identically, including the one with no parameter involved — so the trigger is the **Mount being one call deeper**, not the prefix's kind, a named-type conversion, or parameter resolution.

The mounted handler itself resolves fine in every case: the route and its method are found, only the prefix is missing.

## Related

#138 (closed) covered a mount prefix lost across frameworks; this is the same class of loss with a single framework and an ordinary helper.

## Acceptance

- A `testdata/` fixture mounting through a helper, asserting the prefix reaches the mounted routes.
- Covers the literal-inside-helper form and the prefix-as-parameter form, since they fail for what may be different reasons.

Issue #34, opened 2025-10-19
how to debug routes parser


sattellite:
> @sattellite It should be fixed now in v0.3.1. Thank you for helping me address this issue, and I'm looking forward to your feedback!

_Originally posted by @ehabterra in https://github.com/ehabterra/apispec/issues/31#issuecomment-3419215452_

Everything is working fine with the router in the testdata. And it even found a few routes in the real project, but not all of them. I don't understand what's going wrong. Could you please explain how I can do the debugging to provide you with more information for future fixes?

ehabterra:
First of all, try adding these parameters as they might solve the issue:

```bash
--analyze-framework-dependencies=false
--auto-include-framework-packages=false
--auto-exclude-tests=false
--auto-exclude-mocks=false
```

For more detailed output, add:
```bash
--verbose
```

If the previous step didn't succeed in showing missing routes, there are major debugging steps you can check to see if the failure is in one of them. Here are the specific breakpoints and values to examine:

1. **Metadata Analysis:**
   - **Breakpoint:** [`internal/metadata/metadata.go:414`](https://github.com/ehabterra/apispec/blob/main/internal/metadata/metadata.go#L414)
   - **Values to check:** The `metadata` object being returned to see if it contains the expected call graph data

2. **Tracker Tree Processing:**
   - **Breakpoint:** [`internal/spec/tracker.go:428`](https://github.com/ehabterra/apispec/blob/main/internal/spec/tracker.go#L428)
   - **Values to check:** 
     - [`internal/spec/tracker.go:240`](https://github.com/ehabterra/apispec/blob/main/internal/spec/tracker.go#L240) - `roots` slice to see if route nodes are being identified

3. **Route Extraction:**
   - **Breakpoint:** [`internal/spec/mapper.go:91`](https://github.com/ehabterra/apispec/blob/main/internal/spec/mapper.go#L91)
   - **Values to check:** [`internal/spec/mapper.go:88`](https://github.com/ehabterra/apispec/blob/main/internal/spec/mapper.go#L88) - the `routes` variable to see what routes are being extracted

**Helper tools for debugging:**
- **(coming soon)** Use `apidiag` with the `--diagram-type=tracker-tree` flag to visualize the tracker tree data instead of the call graph:
  ```bash
  apidiag --diagram-type=tracker-tree --verbose 
  ```
- The `apidiag` tool is available at: https://github.com/ehabterra/apispec/tree/main/cmd/apidiag

[...]

PR #277, opened 2026-08-02
fix: fold a concatenated registration path instead of losing it (#274)

### LINKED ISSUE #274: Route path built by string concatenation is dropped: `r.Post(opts.BaseURL+"/things", h)` documents no path
A registration whose path argument is a **binary expression** loses the path entirely. The route is still discovered — it is documented at the mount prefix alone, or at `/` when there is no mount.

This is not an exotic shape: it is what **oapi-codegen emits for every chi server it generates**, so a project whose HTTP layer is generated documents no usable paths at all.

## Reproduction

```go
type ChiServerOptions struct {
	BaseURL    string
	BaseRouter chi.Router
}

func HandlerWithOptions(si ServerInterface, options ChiServerOptions) http.Handler {
	r := options.BaseRouter
	if r == nil {
		r = chi.NewRouter()
	}
	wrapper := ServerInterfaceWrapper{Handler: si}
	r.Group(func(r chi.Router) {
		r.Post(options.BaseURL+"/things", wrapper.CreateThing)   // <-- concatenated path
	})
	return r
}

func main() {
	root := chi.NewRouter()
	root.Mount("/api", HandlerWithOptions(impl{}, ChiServerOptions{}))
	_ = http.ListenAndServe(":8080", root)
}
```

| path argument | documented |
| --- | --- |
| `options.BaseURL+"/things"` | `/api/` ✗ |
| `"/things"` | `/api/things` ✓ |

Isolated against the same fixture with every other ingredient held constant — router from a struct field, the `r == nil` fallback, the `r.Group(...)` closure, and a method-value handler that dispatches through an interface all resolve correctly. **The concatenation is the only thing that breaks it.**

## Where

`BasePatternMatcher.resolvePathArg` (`internal/spec/pattern_matchers.go`) resolves `KindLiteral` and const idents, and synthesizes a `{placeholder}` for `KindCall` (issue #34). `KindBinary` falls through to `GetArgumentInfo`, which yields nothing usable. The existing doc comment names this as a known follow-up:

> All other kinds fall through to GetArgumentInfo for backwards compatibility — handling KindIdent (non-const variable) and KindBinary (`prefix + "/x"`) similarly is a possible follow-up but is out of scope for the initial fix.

## Suggested handling

Concatenation of parts that are individually resolvable should fold to the joined string: a const/literal operand contributes its value, and an empty-valued operand (a zero-value struct field like `BaseURL`) contributes nothing — `"" + "/things"` is `/things`. When an operand cannot be resolved, the honest fallback is the existing `{placeholder}` treatment rather than dropping the segment, so the route is still addressable.

Worth covering both operand orders and a chain of three (`base + version + "/things"`).

## Acceptance

- A `testdata/` fixture with a generated-style server registering through `opts.BaseURL+"/path"`, asserting the full path is documented.
- The unresolvable case degrades to a placeholder, not to an empty segment.

### LINKED ISSUE #276: oapi-codegen-generated servers document no usable paths (umbrella)
A project whose HTTP layer is generated by **oapi-codegen** and wired with **google/wire** produces a spec with no usable paths — measured on a 46-package service: `spec mapped (0 paths)`.

Nothing in that chain is exotic; it is what both tools emit by default. This is an umbrella for the two defects that cause it, filed separately with isolated reproductions:

- **#274** — the registration path is a concatenation, `r.Post(options.BaseURL+"/things", wrapper.CreateThing)`, and the path segment is dropped.
- **#275** — the mount is performed inside a helper, `MountServer(prefix, r, server)`, and the prefix is lost.

Together they leave every route at `/` or at the bare mount prefix.

## The shape

Faithful multi-package reproduction (four packages, mirroring generated code + DI wiring):

```go
// package gen — what oapi-codegen emits
func HandlerWithOptions(si ServerInterface, options ChiServerOptions) http.Handler {
	r := options.BaseRouter
	if r == nil {
		r = chi.NewRouter()
	}
	wrapper := ServerInterfaceWrapper{Handler: si}
	r.Group(func(r chi.Router) {
		r.Post(options.BaseURL+"/things", wrapper.CreateThing)
	})
	return r
}

// package carts — the module server
func Server() http.Handler { return gen.Handler(&Carts{}) }

// package providers — the mount helper
func MountServer(prefix RoutePrefix, r *chi.Mux, server http.Handler) bool {
	r.Mount(string(prefix), server)
	return true
}

// package injectors — what wire generates
func MountCartsServer(prefix providers.RoutePrefix, r *chi.Mux) bool {
	handler := carts.Server()
	return providers.MountServer(prefix, r, handler)
}

func main() {
	r := chi.NewRouter()
	injectors.MountCartsServer("/api", r)
	_ = http.ListenAndServe(":8080", r)
}
```

Result: **one path, `/`** — the route is discovered, and both its own path and its mount prefix are gone. Expected `/api/things`.

## What already works, and is worth keeping covered

Isolated one ingredient at a time against the same fixture, all of these resolve correctly on their own:

- the router taken from a struct field (`r := options.BaseRouter`) with the `if r == nil` fallback;
- the `r.Group(func(r chi.Router) { ... })` closure;
- a method-value handler on a local wrapper struct (`wrapper.CreateThing`);
- that wrapper dispatching through an interface (`siw.Handler.CreateThing(w, r)`) to the concrete implementation;
- the handler being produced by a function call passed as an `http.Handler` argument.

So the generated shape is *nearly* supported — two path-composition defects account for the whole loss.

## Acceptance

A `testdata/` fixture of the generated shape end to end, asserting `/api/things` with its request body, so this combination is covered as a whole and not only as two separate unit fixtures. Closing depends on #274 and #275.

ehabterra:
Fixes #274.

## The defect

A registration whose path argument is a binary expression resolved to **nothing**, so the route was documented at its mount prefix alone — or at `/` when there was none. The literal part of the path, the only part actually written in the source, was the part that disappeared.

It is what oapi-codegen emits for every chi server it generates:

```go
r.Post(options.BaseURL+"/things", wrapper.CreateThing)
```

`resolvePathArg` already handled literals, const idents and call expressions; `KindBinary` fell through to `GetArgumentInfo`, and its own doc comment named this as a known follow-up.

## What it does now

A `+` chain flattens and each operand resolves on its own:

| operand | resolves via |
| --- | --- |
| literal / const (incl. another package) | `ConstantValue` |
| parameter | the caller's argument |
| field of a struct the caller passed | the composite literal — a field it does not set has its **zero value** |
| anything else | `{placeholder}`, same as a call expression (#34) |

The zero-value rule is what makes the generated shape land exactly: `ChiServerOptions{}` means `BaseURL` is `""`, which *contributes nothing* rather than being unknown.

Two resolution paths had to be added, both narrow on purpose:

- **`paramArgOfEnclosingFunc`** — a call inside a closure has the func literal as its caller and no edge calls that literal by name, so `argViaParent`'s frame check can never match. Generated servers register inside `r.Group(func(r chi.Router){...})`, so the parameter is captured from the function that *defines* the closure.
- **`uniqueCallerArg`** — a registration made on a router *parameter* is re-homed under that router's producer, so the call binding the prefix is nowhere on the ancestor path. It resolves from the function's call sites, and **only when they agree**: two callers passing different prefixes leave the value genuinely ambiguous, where the honest answer is the placeholder, not whichever literal comes first (golden rule #7).

An unresolvable operand degrades to a placeholder rather than vanishing. A shortened path claims the handler answers somewhere it does not; an obviously incomplete one stays addressable and honest.

## Deliberate behaviour change

`testdata/wrapper_router` registers through a helper as `m.Delete(base+"/{id}", h)`, with the caller passing `"/users"`. That was documented at `/` — the root, which no handler there serves. It is now `/users/{id}`, and the two assertions that pinned the old path are updated to say why.

## Verification

- **4 OSS projects**: path sets byte-identical.
- **A large real service**: exactly one change, `/` → `/scheduler*`, which matches `router.Handle(queue.SchedulerRootPath+"*", ui)` in its source — a cross-package const operand, now resolved. Route count unchanged (68).
- Full suite, `go vet`, `gofmt`, `golangci-lint` clean.

## Tests

- `testdata/concat_path` + structural test — one route per operand class, plus the placeholder degradation and the request body surviving the fold.
- `internal/spec/concat_path_test.go` — unit tests for the folding itself: multi-operand order, empty operand, placeholder naming, non-`+` operators left alone, plain literal paths unchanged.

## Scope

This does **not** close #276. The generated-server project that prompted it still documents 0 paths: its mounts go through a helper, which is #275, and #276 notes a further gap beyond those two. This PR fixes the path composition only, verified by fixtures rather than by that project.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved route and mount path resolution for concatenated paths.
  * Preserved unresolved path segments as placeholders instead of dropping them.
  * Correctly resolves caller parameters and composite-literal values in generated paths.
  * Fixed generated helper routes to retain their literal base paths.

* **Tests**
  * Added coverage for nested, empty, unresolved, and unsupported path expressions.


ehabterra:
Reviewed all six comments against the code. Both **actionable** findings were real and are fixed in 8e98125; three nitpicks applied; one declined with reasoning.

## Fixed

**Only the first placeholder name was reported** (major, valid). Each name becomes a declared path parameter via `appendDynamicParamRefs` (mapper.go:359) and the component registration at mapper.go:462, so `a() + b() + "/x"` emitted two `{name}` segments with only one declared — an invalid path template. `resolvePathArg` now returns the full list, deduped in source order, and both call sites append all of them. Pinned by a unit test asserting `[baseOf versionOf]`.

**A positional composite literal was read as all-zero fields** (major, valid — and the sharpest catch in the review). `structFieldValue` concluded "absent from the keys ⇒ zero value", which only holds for a *keyed* literal. `ChiServerOptions{"/api", nil}` names no fields, so it silently shortened the path — exactly the defect this PR exists to prevent. A non-key-value element now reports the operand as unresolved, so it degrades to `{BaseURL}`. I did not implement positional resolution (matching element index against the struct's field order); that is a real capability, but guessing it here would be worse than reporting the operand as unknown.

**Stale `/` route** (nitpick): both `wrapper_router` tests now assert `DELETE /` is absent, not merely that `/users/{id}` exists — they could otherwise coexist.

**Non-deterministic `/dyn` lookup** (nitpick): collected and sorted, so a failure names the same path every run.

**Unit tests with a tracker node** (nitpick): added `TestResolveConcatenatedPathWithNode` covering the paths only the fixture reached — a parameter resolved from its caller, a set field, an unset field, the positional literal, and two unresolvable operands.

## Declined

**Exposing metadata through the `ContextProvider` interface.** The type assertion mirrors what `traceVariable` and `findAssignmentFunction` already do in this same file, and what the extractor does in `request_provenance.go`. Widening the interface would touch every implementation and mock — a reasonable refactor, but its own change, not part of a path-resolution fix.

## Verification after the fixes

4 OSS projects byte-identical; the large real service unchanged from the pre-review run (68 paths, the single `/` → `/scheduler*` improvement); full suite, `go vet`, `gofmt` and `golangci-lint` clean.

Issue #270, opened 2026-07-30
CallArgument allocation is 21.6% of the run: one argument tree is rebuilt per call site, never shared


ehabterra:
Measured with `--mem-profile` on a ~900-route project. `alloc_space` totals 5,666 MB for the run, of which:

| | allocated | share |
| --- | --- | --- |
| `metadata.NewCallArgument` | **1,223 MB** | **21.6%** |
| `spec.(*chainInterner).push` | 640 MB | 11.3% |
| `spec.(*LazyTree).newNode` | 604 MB | 10.7% |
| `slices.Sorted[…].AppendSeq` | 299 MB | 5.3% |

`NewCallArgument` is the single largest allocator in the program, by a wide margin.

## What it is not

**Not field layout.** `CallArgument` is 216 bytes and `fieldalignment` reports only a *pointer-byte* improvement for it (216 → 120), with no size claim — its footprint is already minimal. Reordering would reduce GC scan surface, not bytes, and it is yaml-serialised in declaration order so it would churn every metadata golden. Measured for the two structs that *can* shrink (`LazyNode` 104→96, `TrackerNode` 128→120): no detectable change in peak RSS over three A/B repeats, because the difference is below this machine's ±8% run-to-run noise.

So the volume is the problem, not the size of each one.

## Where the volume comes from

`ExprToCallArgument` builds a **fresh tree per call site**, recursively, with one `CallArgument` per expression node — and it is called for every argument of every recorded call (`processCallExpression` → the arg loop), plus again from assignment processing (`processAssignment` is 1,440 MB cumulative, 25%). On a project with 82k call edges that is millions of nodes.

Two things make it worse than it needs to be:

1. **Identical expressions are rebuilt every time they appear.** `http.StatusOK`, `"application/json"`, `err`, `ctx` — the same literal or ident appears at thousands of call sites and gets a fresh 216-byte node with a fresh subtree each time. Every scalar field is already an interned string-pool index, so two nodes for the same expression are usually *byte-identical*.
2. **Every node is heap-allocated individually.** `newNode`-style slab allocation was applied to `LazyNode` for exactly this reason and helped there; `CallArgument` has no equivalent.

## Directions, roughly in order of expected value per unit of risk

[...]

Issue #269, opened 2026-07-29
Two routes sharing a body helper can resolve to each other's DTO when expansion goes deeper


ehabterra:
Latent today, real, and it will be hit by the next change that expands more. Found while reverting #264 (#268), which is the only reason it surfaced.

## The defect

Two routes registered on the same router that pass their request bodies through a **shared decode helper** can be documented with **each other's** DTO. Not a missing schema — a confidently wrong one.

Observed on a 68-route service, four endpoints affected. One example:

| | `POST /cart/api/v1/item` requestBody |
| --- | --- |
| correct (v0.5.4 and current `main`) | `AddItemRequest` |
| when expansion runs deeper | `UpdateItemRequest` |

The other three swapped `LoadCartRequest` ↔ `DeliveryChargeDetailsRequest` and `LoadCartRequest` ↔ `CalculateFinalTotalRequest` — in every case a DTO belonging to a *sibling route on the same router*, which is the tell: the argument resolution is reaching a neighbouring registration's body instead of its own.

## Why it is not reproducible on `main` today

It needs deeper expansion than any shipped version performs. It was exposed by #264 part 2, which stopped the global node budget from applying below a route registration — route subtrees then expanded past any depth previously reached, and the wrong match won.

Bisected precisely:

| | result |
| --- | --- |
| v0.5.4 | correct |
| `main` before #264 | correct |
| \+ #264 part 1 (child ordering only) | correct |
| \+ #264 part 2 (per-route budget) | **wrong** |

Raising `--max-nodes-per-route` did **not** fix it, which is what rules out the budget value and points at resolution rather than truncation: more depth simply gives the wrong candidate more chances to be selected.

## Why it matters despite being unreachable now

Every direction currently in flight expands more than today does:

- #264, whenever it is retried (the point of it is to reach code the budget currently cuts off)
- `--resolve-call-graph` becoming the default (#244, #245, #253) — resolution adds real edges
- #248 / #249, gated on expansion being affordable (#264)

So this is not a curiosity: it is a wall that each of those will hit, and it fails in the worst way — silently, with a plausible schema.

## Root cause, as far as it is understood

Unknown, and deliberately not guessed at. What is known:

- it is a **resolution precision** problem, not a budget or truncation problem (raising the cap does not change it)
- the wrong DTO always belongs to a **sibling route on the same router**
- it involves a **shared helper** — the routes that resolve correctly at all depths are the ones whose body is decoded at their own call site

[...]

Issue #253, opened 2026-07-29
Indirect-call resolution doubles the mapping stage on a large project, because it resolves calls the spec never asks about


ehabterra:
Reported from real use: a run with `--resolve-call-graph` (via apispecui) was killed after **12+ minutes**. Reproduced and measured.

## Measured on gitea

| `--max-nodes` | mapping off | mapping on | total off | total on | components off / on |
|---|---|---|---|---|---|
| 50,000 (default) | 10s | 11s | 46s | 55s | 1 / 15 |
| 100,000 | 1m30s | **2m56s** | 2m06s | 3m41s | 41 / **87** |
| 300,000 | 4m04s | not run | 4m41s (7.8GB) | — | — |

The cost is **not** the flat +19% I documented from the default-limit run. It scales with the node budget, and at a raised budget — which is exactly what a large project needs to reach all its routes — the mapping stage roughly doubles. Extrapolating the 300k row lands past 8 minutes before SSA overhead, which matches the report.

## Why

Building SSA and running VTA is only ~6s of it. The rest is the feature working: 7,370 rewritten call sites (2,880 interface, 4,490 promoted) point at real function bodies the syntactic graph left pointing at an interface or at a type that only embeds the declaring one. More reachable program = more tree to expand.

And it is not wasted work — at the 100k budget, 340 of 389 operations gained response or request detail, and components went 41 -> 87.

## The actual defect

**All 7,370 sites are resolved, but only the ones the spec asks about matter.** Resolution is applied to the whole call graph before anything knows which calls will be consulted by a route, response, request-body, param, mount or security pattern. Most of gitea's resolved calls are business logic that no pattern will ever match, and every one of them widens the tree.

## Directions

1. **Resolve on demand.** Keep the `CalleesAt()` index and rewrite a call the first time something asks about it, rather than rewriting the graph up front. Fits the lazy tracker's design, and the cost becomes proportional to what the spec actually reads.
2. **Resolve only what a pattern can reach.** Restrict rewrites to call sites whose callee (before or after resolution) could match a configured pattern, or that sit inside a handler subtree. Cheaper to implement than (1), but needs care: the reachability experiment in #244's sibling work showed that a pattern-based filter over the call graph is not conservative (it deleted 124 of 312 operations on one project), so this must gate *rewrites*, not *expansion*.
3. **At minimum, warn.** apispecui should say something when resolution is enabled together with a raised node budget, since that combination is what produces a multi-minute run with no progress indication.

[...]

Issue #249, opened 2026-07-29
metadata loses the receiver on some cross-package method calls, recording them as functions of the CALLER's package


ehabterra:
Found by classifying the 10,182 gitea call sites where the SSA+VTA graph names a different callee than the syntactic one (PR #241). 1,240 of them (12%) fit no explanation — neither an interface resolved to an implementation nor a method promoted through embedding. I treated them as untrustworthy joins and left them alone, then went looking at what they actually are.

They are not mis-joins. They are calls metadata recorded wrongly.

## Two shapes, on a 246-route service

**The receiver is lost, and the callee is attributed to the caller's package:**

```
recorded:  internal/http.ListTeacherSubjects        (no receiver, http package)
resolved:  internal/curriculum.Service.ListTeacherSubjects
```

Same for `AddTeacherSubject`, `RemoveTeacherSubject`. The call is `deps.CurriculumService.ListTeacherSubjects(...)` — a method on a service held in a struct field, in another package. Metadata kept the method name, dropped the receiver, and filed it under the package doing the calling.

**The receiver is recorded as the literal string `interface`:**

```
recorded:  cmd/server.interface.SignedGetURL
resolved:  internal/storage.s3driver.SignedGetURL
```

An inline/anonymous interface type renders as `interface`, so the owner is a type name that cannot be looked up, and nothing downstream can scope a pattern to it.

Counts on that service: 3 with no receiver, 7 same-name-different-owner, 0 other — i.e. **all ten** are one of these two shapes.

## Why it matters

- A pattern scoped by `recvTypeRegex` cannot match a call whose receiver was dropped. This is the same class as golden rule #10 (a cross-package callee's name lives in `Fun.Sel`), which is exactly the kind of bug that is invisible in single-package fixtures.
- It corrupts the call graph for every consumer: reachability, the tree, diagrams, and the resolved-graph join.
- It inflates the "unexplained" bucket that the resolved-callee rewrite deliberately refuses to act on, so fixing it also **raises the share of resolvable call sites** (72% actionable today).

## Direction

Both shapes are receiver-resolution failures at metadata time, not spec-layer decisions. Reproduce each as a `testdata/` fixture first (golden rule: probe every uncovered shape with a fixture, never assert from reasoning):

1. a method on a service held in a struct field of another package, called from a handler;
2. a call through a field whose declared type is an inline `interface{ ... }`.

Then fix receiver resolution for both, rather than special-casing the rendered name.

## Acceptance

[...]

PR #291, opened 2026-08-06
fix(spec): bound expansion per route, so finding routes and detailing them stop competing (#264)

### LINKED ISSUE #264: The node budget is global and spent depth-first, so improving the call graph makes the spec worse

ehabterra:
Fixes #264.

The node budget was one allowance for the whole walk, spent depth-first, with no notion of which subtree leads to a route registration. On a ~900-route project it was gone inside configuration and logging packages before the walk reached the routers: **12 paths documented out of ~900**, and every improvement to the call graph — in either direction — made that worse, which is why #248, #249 and #253 are all gated on this.

This restores the two reverted parts (#265, #266 — reverted by #268 because of #269, now fixed by #271/#273), plus the group-scope fix, and then fixes the defect that made the split ineffective.

## The split was only half done

Bounding a route's detail separately is the fix, but the keys a route's own subtree discovered were still counted into the single global `nodesBuilt` that `MaxNodesPerTree` reads. Expanding one route in more detail therefore spent the allowance the walk needed to **find** the next route, so raising the per-route budget made the spec **smaller**:

| `--max-nodes-per-route` | 20,000 | 200,000 | 1,000,000 |
|---|---|---|---|
| paths (before this commit) | 181 | 163 | **103** |
| paths (after) | 491 | 581 | **640** |

That first row is #264's own "improving expansion makes the spec worse", reproduced one level down from where the issue found it. Only keys the wiring walk reaches may exhaust `MaxNodesPerTree`; `seenKeys` carries a bit set rather than a bool so the second count costs no second map — it is the largest allocation in the run.

The wiring bit is set on its own occasion, not only when a key is first seen at all: a key reached below a route first and by the wiring walk later still costs the walk, or a traversal that happens to descend early understates its own cost without bound.

## The default is set by what it must not cost

A per-route cap is a **new** restriction. Before it, a route's detail was bounded only by the global budget, so every project that never hit that budget was effectively unbounded below its registrations — and any useful ceiling is a silent regression for them, since a truncated route loses a schema, not a path, so the spec still looks complete.

Measured against per-project snapshots: **20,000** lost detail on three real projects; **200,000** cleared all but one endpoint, whose four responses need somewhere past 400,000; **1,000,000** restores exact parity everywhere. It is cheap because deep routes are rare — on the project holding that endpoint a million costs nothing measurable (9.0s either way).

## Results

| | before | after |
|---|---|---|
| ~900-route project | 12 paths, 46s, 2.9GB | **640 paths**, 166s, 7.3GB |
| 10 other real projects | — | paths identical, wall clock within noise (worst +1.1s) |

## Gate

`scripts/compare-spec.sh` over all 74 fixtures and 11 real projects — the harness that caught all three traps recorded on #264 while the fixture suite stayed green. **Zero drift**, except 12 CHANGED enum values on one project that `main` produces identically (verified by diffing the two runs). Wall clock and path counts unchanged on every project that was fine before.

Full suite, `-race`, `golangci-lint`, `gofmt`, `go vet` all clean.

## Tests

Both fixtures #264 required before a retry:

- **`TestGroupClosurePerRouteBudgetIsScopedPerRoute`** — trap 1. 15 routes in one group closure, budget driven to 5 so every subtree truncates; asserts all 15 routes are still **discovered** and that **15** scopes are opened, not 1. A group-scoped budget fails it immediately. (`RouteRegistrationMatcher` must keep matching groups for reachability; only `TerminalRouteMatcher` may open a scope.)
- **`testdata/shared_body_helper`** (the #269 fixture) now also sweeps `--max-nodes-per-route`, including 2,000,000 — the direction that exposed the wrong-DTO regression in the first place. Depth is a starvation setting in reverse, so the same assertion has to hold with the budget raised far past the default.
- **`TestWiringBudgetIsNotChargedForRouteDetail`** / **`TestWiringKeyCountedWhenReachedLater`** — the accounting invariant. Mutation-checked: both fail against the broken accounting.

## Note for reviewers

`--max-nodes` changes meaning: it now bounds only the walk that finds route registrations, and the two limits fail differently — `--max-nodes` spent means routes are **missing**, `--max-nodes-per-route` spent means one **named** route is less detailed and no other route is affected. README, `docs/DEBUGGING.md` and the flag help are updated.

Unblocks #248, #249 and #270; #245 next.

## Summary by CodeRabbit

* **New Features**
  * Added the `--max-nodes-per-route` CLI option to independently limit analysis detail for each route.
  * Route-level analysis budgets are scoped independently, helping preserve route discovery while controlling detailed expansion.
  * Analysis results now report route truncation counts, limits, affected routes, and the first truncated route.

* **Documentation**
  * Updated CLI, debugging, and limits documentation to explain global versus per-route limits and related warnings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,8 @@ apispec --output openapi.yaml --skip-cgo
 | `--diagram`                 | `-g`      | Write call-graph HTML to this path                     | `""`                            |
 | `--paginated-diagram`       | `-pd`     | Use paginated rendering for the diagram                | `false`                         |
 | `--diagram-page-size`       | `-dps`    | Nodes per page in paginated diagram (50–500)           | `100`                           |
-| `--max-nodes`               | `-mn`     | Max nodes in the call graph                            | `50000`                         |
+| `--max-nodes`               | `-mn`     | Max nodes in the walk that finds route registrations   | `50000`                         |
+| `--max-nodes-per-route`     |           | Max nodes expanded below one route registration (lazy engine) | `1000000`                |
 | `--max-children`            | `-mc`     | Max children per node                                  | `500`                           |
 | `--max-args`                | `-ma`     | Max arguments per function                             | `100`                           |
 | `--max-nested-args`         | `-md`     | Max depth for nested arguments                         | `100`                           |
@@ -652,7 +653,7 @@ APISpec turns Go source into an OpenAPI document through a fixed sequence of sta
 - *Importance:* Nothing downstream touches the raw AST again — metadata is the substrate. String-pooling plus sorted iteration at every boundary make the output **deterministic** (clean release diffs and reliable golden tests), and `--write-metadata` dumps this model so a missed route can be debugged.
 
 **6. Build the tracker tree**
-- *Role:* Starting from each route-registration call site, expand the call graph down to the actual handler and the calls made inside it — through wrappers, groups, mounts, handler factories, and helper functions — bounded by engine-specific limits (see [Performance & Limits](#performance--limits)). The default **lazy** tree expands subtrees on demand and is bounded by `--max-nodes`/`--max-children`/`--max-args`/`--max-instances-per-key`; the eager tree (`--legacy-tracker`) materializes them up front and additionally honors `--max-recursion-depth` and `--max-nested-args`.
+- *Role:* Starting from each route-registration call site, expand the call graph down to the actual handler and the calls made inside it — through wrappers, groups, mounts, handler factories, and helper functions — bounded by engine-specific limits (see [Performance & Limits](#performance--limits)). The default **lazy** tree expands subtrees on demand and is bounded by `--max-nodes` (finding registrations) / `--max-nodes-per-route` (detail below one) / `--max-children`/`--max-args`/`--max-instances-per-key`; the eager tree (`--legacy-tracker`) materializes them up front and additionally honors `--max-recursion-depth` and `--max-nested-args`.
 - *Purpose:* Connect a route to the concrete code that actually serves it, following real control flow rather than assuming the handler lives where the route is declared.
 - *Importance:* In real codebases the handler is rarely at the registration site — it's behind middleware, a group closure, a mounted sub-router, or a factory. This traversal is what makes detection work across those styles. The bounds are the safety brake that turns a pathological (deep or cyclic) call graph into a truncation warning instead of a hang or out-of-memory.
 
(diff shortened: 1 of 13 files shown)


ehabterra:
All three addressed in 6f9bd56, each verified against the code first.

**1. `--max-nodes-per-route` is lazy-only — valid.** `grep -rn MaxNodesPerRoute --include='*.go'` confirms it reaches `internal/spec/lazytree.go` and nowhere else; the eager tree has no notion of a route scope and bounds a route with `MaxRecursionDepth` instead. So it is accepted under `--legacy-tracker` and silently does nothing. Noted in the flag help, and also on `engine.EngineConfig.MaxNodesPerRoute` and `metadata.TrackerLimits.MaxNodesPerRoute` — where a caller sets it, not only where it is read.

**2. The "shipped default" claim — valid, and worse than stale.** 20,000 is exactly the value `DefaultMaxNodesPerRoute`'s own doc comment records as *rejected*, because three real projects lost detail at it. A test calling it what ships is a signpost back to the rejected value. Rather than rewrite the number, the sweep now reads `intspec.DefaultMaxNodesPerRoute`, so it re-runs at whatever actually ships and cannot go stale again if the default moves.

**3. Positional `m.CallGraph[N]` — taking it.** Fair point: an inserted edge shifts the indices, and the tests keep compiling and may keep passing while asserting nothing about a route registration. Replaced with a name-based `edgeTo` helper across all four tests.

Full suite, `-race`, `golangci-lint`, `gofmt`, `go vet` clean. No output-affecting change in this commit, so the cross-project harness result in the description still stands.

PR #293, opened 2026-08-07
fix(spec): stop deciding which response body survives from its type NAME (#287)

### LINKED ISSUE #287: Error responses are identified by the substring "error" in the type name, so ProblemDetails is missed and ErrorBudgetReport is demoted
## What

`internal/spec/extractor.go:1583` decides whether a response DTO is an error type by looking for the letters "error" in its name:

```go
// isErrorBodyType reports whether a body type name looks like an error DTO
// (e.g. ErrorResponse, APIError). Used only as a tie-break for the default slot.
func isErrorBodyType(bodyType string) bool {
	return strings.Contains(strings.ToLower(bodyType), "error")
}
```

It is the tie-break in `preferResponse` (`:1560`) that picks which of two competing responses fills the `default` slot.

## Why it's wrong

Golden rule #9 forbids exactly this shape — a type-name test standing in for a fact the lower layers can answer. It is wrong in both directions:

- **Misses** the most common error DTOs that don't spell it out: RFC 7807 `ProblemDetails`, and the usual `Failure`, `Fault`, `Problem`, `Rejection`, `ValidationResult`.
- **Claims** ordinary domain types: `ErrorRateReport`, `ErrorBudget` (SLO tooling), `TerrorAlert`, `MirrorConfig` — any of which, matched here, is demoted below a genuine error response and loses the `default` slot to it.

Because it only breaks ties, the failure is quiet: the spec still has a `default` response, just the wrong schema in it, with nothing to indicate a choice was made on a substring.

## The fact is available

Metadata already knows things that actually answer the question, none of which this consults:

- whether the type implements `error` (`Type.Implements`, subject to the stdlib-interface gap in #178);
- whether the response is written on a non-2xx status branch — `mapper_field_status.go` and `response_dest.go` already reason about the status a body is written with;
- whether the type is the one the framework's error path returns (the config layer knows each framework's error responder).

Any of those is evidence. The name is not.

## Suggested fix

Replace the name test with a status-based one: prefer the response whose resolved status is 2xx for the success slot and let the non-2xx one take `default`. Where no status is resolvable for either candidate, keep both orderings deterministic by the existing `BodyType` comparison (`:1571`) and drop the error tie-break entirely — an arbitrary-but-stable choice is more honest than a wrong-but-confident one.

Fixture: a handler returning `ErrorBudgetReport` on 200 and `ProblemDetails` on 400, asserting the 200 type is not demoted and the 400 type fills `default`.

ehabterra:
Fixes #287.

`isErrorBodyType` tested for the letters `error` anywhere in a body type name and used the answer to break ties for the `default` slot. **Losing that tie means being dropped from the spec**, so a substring silently replaced a handler's real response with another one — and the operation still had a response, just the wrong schema, with nothing to indicate a choice had been made on a name.

## Removed, not replaced

The issue proposed a status-based tie-break instead. That turns out not to be available: every candidate at this point has an unresolved status **by construction** — the mapper collapses onto `default` only for `StatusCode < 0` (`mapper.go:715`). The status a body was written with is exactly what is missing there.

The other option was to widen the name test to whole camelCase words, the way handler-name verb inference does. That is the trade golden rule #9 forbids: it drops `MirrorConfig` and `TerrorAlert` while still claiming `ErrorBudgetReport` (leading word) and still missing `ProblemDetails` — a false positive swapped for a false negative, not progress.

So the guess is gone. What remains is concrete-beats-generic, then a stable `BodyType` comparison: deterministic, and claiming nothing it cannot support.

## It decided nothing on any measured code

I instrumented the branch before touching it and ran the whole corpus:

| project | times the tie-break line was reached | decided by the name |
|---|---|---|
| 68-route service | 7,909 | **0** |
| 350-route service | 1,574 | **0** |
| 265-route service | 864 | **0** |
| 8 other projects + 74 fixtures | — | **0** |

The two competitors have the **same `BodyType` every single time** — 10,347 for 10,347 — so the entire tail after the concreteness check was comparing a type with itself. This was a trap waiting for the first handler that writes two differently-named bodies with unresolvable statuses, not a working heuristic. It also means removal is provably output-neutral on everything measurable, which the gate confirms.

## Gate

`scripts/compare-spec.sh` over 74 fixtures and 11 real projects: **zero drift**, as predicted. The only reported difference is the pre-existing enum drift on one project that `main` produces identically (fixed separately in #292). Full suite, `-race`, `golangci-lint`, `gofmt`, `go vet` clean.

## Tests

`testdata/error_named_domain_type` holds all three shapes in one fixture:

- **`ErrorBudgetReport`** — the *success* body. An error budget is a quantity, not a failure; demoting it drops the handler's real 200.
- **`ProblemDetails`** — the genuine RFC 7807 error body, which no name test can recognise.
- **`MirrorConfig`** — `error` as a bare substring, no word boundary near it.

**Honest scope, stated in the test's doc comment:** this fixture does not reach `preferResponseInfo` and passed before the change as well as after — both its statuses resolve, so the bodies land in different slots and never compete. Reaching the tie-break would require making *both* statuses unresolvable, at which point the fixture pins the behaviour of a guess rather than of a fact. It is a change-detector that no future error classifier may quietly demote a domain type again; the direct evidence is the unit coverage in `extractor_sweep_test.go` and the measurement above, recorded in the comment on `preferResponseInfo`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved generated API response mappings by selecting schemas consistently based on their type definitions rather than names containing “error.”
  * Ensured concrete response schemas continue to take precedence over generic schemas.
  * Improved consistency regardless of response processing order.

* **Tests**
  * Added regression coverage for endpoints returning error-named domain types.
  * Added validation for response mappings across successful and error status codes.
  * Expanded coverage for ordinary domain types whose names include “Error.”

### DIFF
--- a/generator/testdata_error_named_domain_type_test.go
+++ b/generator/testdata_error_named_domain_type_test.go
@@ -54,6 +54,9 @@ func TestTestdata_ErrorNamedDomainType(t *testing.T) {
 		t.Fatalf("GenerateOpenAPI: %v", err)
 	}
 
+	noDanglingRefs(t, out)
+	noUnresolvedPlaceholders(t, out)
+
 	for _, tc := range []struct {
 		path   string
 		status string
@@ -82,8 +85,17 @@ func TestTestdata_ErrorNamedDomainType(t *testing.T) {
 			}
 			ref := refOf(resp.Content)
 			if !strings.HasSuffix(ref, "_"+tc.want) {
-				t.Errorf("%s %s documents %q, want %s — %s",
+				t.Fatalf("%s %s documents %q, want %s — %s",
 					tc.path, tc.status, ref, tc.want, tc.why)
+			}
+			// Naming the right type is only half of it: the $ref has to land on a
+			// component that exists, with the fields the Go type declares.
+			comp := componentByName(out, "_"+tc.want)
+			if comp == nil {
+				t.Fatalf("%s resolves to no component schema", ref)
+			}
+			if len(comp.Properties) == 0 {
+				t.Errorf("component for %s has no properties — the type was named but never resolved", tc.want)
 			}
 		})
 	}
--- a/generator/testdata_error_named_domain_type_test.go
+++ b/generator/testdata_error_named_domain_type_test.go
@@ -1,0 +1,99 @@
+// Copyright 2026 Ehab Terra
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package generator
+
+import (
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/ehabterra/apispec/internal/engine"
+	intspec "github.com/ehabterra/apispec/internal/spec"
+)
+
+// TestTestdata_ErrorNamedDomainType pins that a type's NAME never decides which
+// response body survives (issue #287).
+//
+// The fixture holds the two shapes a substring test gets backwards: an ordinary
+// SLO domain type whose name contains "error" (ErrorBudgetReport, the success
+// body), and the genuine RFC 7807 error body whose name contains nothing an
+// "error" test can see (ProblemDetails) — plus MirrorConfig, where "error" is a
+// substring with no word boundary near it.
+//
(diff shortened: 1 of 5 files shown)


ehabterra:
Valid — fixed in the commit just pushed.

The fixture asserted the right types on the right operations but nothing about whether those `$ref`s resolve, which the convention in CLAUDE.md requires and I had skipped. Naming the right type is only half of it: a `$ref` that lands nowhere documents a type the client cannot resolve.

Added `noDanglingRefs` and `noUnresolvedPlaceholders` over the whole document (reusing the shared helpers in `testdata_smoke_test.go` rather than adding new ones), plus a per-case check that each response `$ref` resolves to a component that actually has properties — otherwise a component could exist as an empty shell and still pass. The suffix assertions are now `t.Fatalf`, so a wrong type does not go on to report a confusing second failure about a component it never named.

Suite, lint, vet clean. No production-code change, so the zero-drift harness result in the description still stands.

PR #299, opened 2026-08-08
fix(spec): read a map literal's constant keys instead of documenting an opaque map (#295)

### LINKED ISSUE #295: A map composite literal with constant string keys degrades to additionalProperties, losing both the keys and the element type
Reproduces on `main` with the **default** config.

## What

`map[string]any{"cost_codes": items}` — the standard Go idiom for a named envelope around a slice — is emitted as an opaque map:

```yaml
type: object
additionalProperties:
  type: object
```

Both facts are statically present at the literal: the key is the constant `"cost_codes"`, and the element type is `[]CostCode`. Neither survives, and `CostCode` is not emitted as a component at all.

## Repro

```go
type CostCode struct {
	ID   string `json:"id"`
	Name string `json:"name"`
}

func listCodes(w http.ResponseWriter, r *http.Request) {
	items := []CostCode{{ID: "1", Name: "a"}}
	w.WriteHeader(http.StatusOK)
	_ = json.NewEncoder(w).Encode(map[string]any{"cost_codes": items})   // degraded
}

// control: the same slice, no envelope
func listTyped(w http.ResponseWriter, r *http.Request) {
	items := []CostCode{{ID: "1", Name: "a"}}
	w.WriteHeader(http.StatusOK)
	_ = json.NewEncoder(w).Encode(items)                                  // resolves
}
// r.Get("/codes", listCodes); r.Get("/typed", listTyped)
```

The control is the point: `/typed` resolves to `{type: array, items: {$ref: …CostCode}}` and emits the component. `/codes` returns the same slice, wrapped, and gets `additionalProperties: {type: object}`. So the element type is resolvable — it is being discarded by the map branch, not missing.

**Expected** for `/codes`:

```yaml
type: object
properties:
  cost_codes:
    type: array
    items:
      $ref: "#/components/schemas/…CostCode"
```

## Why it matters more than it looks

This is a type-fidelity loss that an "untyped response" check does **not** catch, because the schema is non-empty — just useless. Downstream it becomes `Record<string, Record<string, unknown>>`, so a consumer gating on `schema: {}` sees a clean run while the operation is effectively undocumented. Measured on a ~500-route service: 9 operations in this shape, plus an unknown number of `200` blocks.

## Suggested fix

When the map value reaching a response write is a composite literal whose keys are all constant strings, emit `properties` from those keys and infer each value's type through the existing resolution, falling back to `additionalProperties` only for genuinely dynamic maps (a variable key, a range-built map, a non-literal value).

Mixed cases should stay honest (golden rule #7): if some keys are constant and others are not, emitting the known `properties` alongside `additionalProperties` for the rest is still strictly more information than dropping everything.

Note the mapper's map branch is one of the two deliberate string-surgery islands recorded in `docs/TYPE_MODEL.md`, so a fix there should stay inside that documented boundary.

ehabterra:
Fixes #295.

`map[string]any{"cost_codes": items}` — the standard Go idiom for a named envelope around a payload — was documented as an opaque map:

```yaml
type: object
additionalProperties: { type: object }
```

That is all the **type** can say, because mapping ran from `map[string]any` alone. Both facts the caller wants are at the literal: the key is a constant, and the value carries its own resolved type. Reading them produces the object it actually is, and — because the values go through `mapGoTypeToOpenAPISchema` like any other — the payload keeps its component and its `$ref`:

```yaml
type: object
properties:
  cost_codes:
    type: array
    items: { $ref: "#/components/schemas/…CostCode" }
```

The fixture's `/typed` control makes the point: the same slice returned unwrapped already resolved correctly, so the element type was never missing — it was being discarded by the map branch.

## Why it is worth fixing rather than tolerating

This is a fidelity loss that an "untyped response" check **cannot** see. The schema is non-empty, just useless, so a consumer gating on `schema: {}` gets a clean run while the operation is effectively undocumented. Downstream it becomes `Record<string, Record<string, unknown>>`.

## Only what the literal says

- A **non-constant key** leaves `additionalProperties` alongside the keys that did resolve. Naming the known keys is strictly more information, and denying the unknown ones would be worse than saying nothing (golden rule #7).
- A map with **no literal** at the write (built at runtime) or with **non-string keys** is untouched — `additionalProperties` is the correct answer there, and the fixture pins both so this cannot degenerate into "map literals become objects".

## The part the fixture caught

The per-value type comes from `resolveArgType`, **extracted** from `ExtractResponse` rather than re-derived. Reading `CallArgument.GetType()` directly was the obvious implementation and it was wrong twice, both caught by building the fixture before believing the happy path:

| value | raw `GetType()` | consequence |
|---|---|---|
| `meta` (same-package struct ident) | `Meta` — **bare** | keyed a component under a second naming convention in the same document |
| `"abc"` (literal) | *empty* | emitted `cursor: {}` |

The extracted sequence — literal → const-ident → call-return → origin trace → deref → generic normalisation — fixes both. Each step exists for a recorded reason, so a second copy of it for envelope values would drift from this one the first time any step is corrected; that is why it is a shared method and not a parallel implementation.

## Gate

`main` vs this branch over all 74 fixtures and 11 real projects, compared semantically (path sets, per-operation status sets, emptied schemas, component sets):

**Zero paths lost, zero status sets lost, zero schemas emptied, zero components lost. 18 components gained** on two projects.

One methodology note for reviewers: run against the committed `openapi-7.68.yaml` snapshots, the harness reports large drift (305 MISSING keys, 32 status changes). That is not this change — those snapshots predate #291, #292 and #293. 244 of the 305 MISSING keys are the `additionalProperties` entries this PR replaces, every status change reads "gained", and the remainder disappears entirely when the comparison is made against `main` instead of the stale baseline. **The snapshots are due a regeneration.**

Full suite, `-race`, `golangci-lint`, `gofmt`, `go vet` clean. The structural test was verified to fail against `main`'s logic.

## Tests

`testdata/map_literal_envelope` covers five shapes in one fixture: all-constant keys with three value shapes, a mixed literal with one computed key, a runtime-built map, a non-string-keyed map, and the unwrapped control.

## Summary by CodeRabbit

- **New Features**
  - Added support for generating OpenAPI response schemas from map literals with constant string keys.
  - Constant keys are represented as individual object properties with inferred value types.
  - Variable or runtime-generated keys continue to use flexible additional-property schemas.

- **Bug Fixes**
  - Improved handling of escaped string keys and preserved existing behavior for unsupported map types.

- **Documentation**
  - Documented map-literal response envelope support and added a related fixture reference.


ehabterra:
Both addressed, one applied and one filed — each verified against the code first.

**1. Resolution-layer unit tests — valid, added in 8aab7dd.** The change shipped with a fixture and a structural test but nothing exercising `constantKeyValue`/`mapLiteralSchema` directly, which the testing conventions require for resolution logic.

The new tests cover what the fixture cannot reach cheaply: an escaped key (`"a\"b"`), an empty-string key, a numeric literal key, a computed key, a value whose type did not resolve, a non-string-keyed map, an empty literal, and a literal where *every* key is unreadable. The escaped-key case earns its place — it pins that keys are **unquoted** rather than trimmed, and trimming would leave the backslash in the property name.

The fall-through group is the important half: each of those must return `nil` so the type-driven mapping still decides. That is what stops this becoming a blanket "map literals are objects", and it is not visible from the happy path.

**2. `typemodel.TypeRef` for deref and generic normalisation — correct in principle, but not this PR's code.** Filed as #301.

The flagged lines are pre-existing and unchanged; this PR only *moved* them out of `ExtractResponse` into `resolveArgType` so map-envelope values could reuse the same sequence. The diff shows them as added lines purely because the block relocated:

```console
$ git show main:internal/spec/extractor.go | grep -c "normalizeGenericInstanceName(bodyType)"
1
```

Converting them is a heavy lift that touches every response body in every spec, and its acceptance test is zero output drift across the fixture suite *and* `scripts/compare-spec.sh` over the external projects. Doing that inside a behaviour fix would put two things in one reviewable change and make the drift gate here meaningless — this PR's gate is currently "purely additive vs main", which is only a useful signal while the change is small. #301 has the scope, the reason it is identity-affecting rather than cosmetic, and a note to check it against the two deliberate string-surgery islands in `docs/TYPE_MODEL.md` so the fix does not just relocate the surgery into a third one.

Full suite, `-race`, `golangci-lint`, `gofmt`, `go vet` clean. No production-code change in 8aab7dd, so the A/B gate in the description still stands.

Issue #182, opened 2026-07-17
Consolidate the overlapping variable / assignment / return-value resolution mechanisms in internal/spec


ehabterra:
### Summary

The spec layer resolves "where did this value come from?" through several overlapping, ad-hoc mechanisms, and code picks a different subset depending on when it was written. There is no single canonical value-provenance API, so callers hand-roll combinations.

**Assignment / origin (variables):**

1. `edge.AssignmentMap[name]` — raw, callee-body-visible assignments on the call edge.
2. `callerAssignmentMap(impl, edge, name)` (`extractor.go`) — the enclosing function's / parent-function's / method-receiver's assignment map (added for status-via-constructor).
3. `metadata.TraceVariableOrigin(...)` — cross-scope variable origin tracing.
4. `latestAssignment(cp, edge, name)` (`extractor.go`) — the edge-map → function-scope "latest RHS" lookup extracted during #181; sits *alongside* (1)–(3) rather than being the one entry point.

`(1)` alone misses assignments made in the handler body (they live on the `Function`, not the Decode/Encode call edge); `(2)`/`(4)` recover those; `(3)` overlaps both for origin tracing.

**Parameter / return / field (across the call boundary) — the fourth family, surfaced by #155/#187:**

5. `resolveArgThroughParams` / `argViaParent` — a helper parameter → the caller's actual argument, multi-hop.
6. `constructorFieldParam` — a return composite-literal field → the constructor parameter it's assigned from.
7. `constructorArgForParam` / `findCallEdge` — the constructor call edge (matched by call-site position) → the argument bound to a parameter.
8. `statusFromConstructorField` / `statusesFromConstructorField` — glue over (2)/(4)/(5)/(6)/(7) to reach a value across handler → helper → constructor/mapper.

These all reach across the handler → helper → constructor/mapper boundary using the same underlying primitives, but as parallel bespoke resolvers.

### Evidence

- PR #181 needed "latest assignment RHS, edge-map then function-scope" in **both** the request-body source resolver (`src := r.Body`) and the response-destination resolver (`dst := w`, `lw := &loggingWriter{w}`) — first duplicated byte-for-byte, then extracted to `latestAssignment` (the immediate de-dup only).
- PR #186 (#155) added (6)/(7)/(8) for "status through an error constructor," a fourth family of the same shape.
- #187 will extend (6)/(7)/(8) to a struct field branch-assigned inside a **returning** function (`api := MapError(err); … api.Status`).

### Proposal

[...]

PR #304, opened 2026-08-08
fix(spec): a Mount written inside a helper keeps its prefix (#275)

### LINKED ISSUE #275: A Mount performed inside a helper function loses its prefix

ehabterra:
Fixes #275 (partially — see *What is not fixed*).

`r.Mount(prefix, server)` applied its prefix at the registration site and lost it silently one function deeper, documenting the routes at paths nothing serves.

## The cause is tree position, not resolution

A mount prefix reaches nested routes by **tree containment**. With the Mount written in a helper, the sub-router was built at the **call site**, so its routes hang under the caller's argument while the Mount node sits in the helper's body — siblings, not ancestor and descendant. The prefix had nothing to apply to.

That explains the part of the issue that looked strange: all three helper forms failed identically, *including* the one with a literal prefix and no parameter anywhere. The prefix was never the problem.

```
main
└── mountLiteral (call)
    ├── arg: Server()          → chi.NewRouter, r.Get("/things")   ← the routes
    └── body: r.Mount("/api", server)                              ← the prefix
```

Resolving the mounted router back through its parameter to the caller's argument gives the subtree the prefix belongs to.

The plain traversal still reaches that subtree unprefixed, which is what would make a duplicate appear — `dropSubsumedMountPrefixes` already exists for exactly that (a route reached through several contexts keeps only its most-mounted form), and the fixture asserts the un-prefixed `/things` does **not** survive alongside `/api/things`.

`handleRouterAssignment`'s body becomes `mountRouterSubtree`, now shared by the two ways a mounted router can live outside the mount node's subtree: assigned to a variable, and passed in as a parameter.

## What is not fixed

A prefix that is itself a **parameter** (`mountParam(prefix string, r, server)`) or a **named-type conversion** (`r.Mount(string(prefix), server)`). Those are path resolution rather than tree position: `resolvePathArg` does not trace parameters.

The fixture asserts them at today's wrong behaviour, so the test **fails and forces an update** when they are fixed rather than the gap going unnoticed.

For whoever picks that up: `resolvePathOperand` already has `callerArgFor` for precisely this case, but it is only wired into *concatenated* paths, not plain path arguments. I tried extending it inside this change and reverted — unwrapping the conversion made the named form resolve to its **type** name (`/github.com/…/RoutePrefix/things`), trading one wrong path for another, which is not progress.

## Verification

- fixture `testdata/mount_via_helper` covers all four forms: mounted at the call site (control), literal prefix in a helper (the fix), and the two parameter forms (change detectors)
- **A/B across 11 real projects: zero paths, statuses, schemas or components differ.** None of them mounts through a helper, so the fixture is the coverage — worth knowing when weighing the gate
- full suite, `-race`, `golangci-lint`, `gofmt`, `go vet` clean

## Summary by CodeRabbit

* **New Features**
  * OpenAPI generation now applies route prefixes when mounts are declared inside helper functions.
  * Routes inherited through helper-provided routers retain applicable tags, parameters, and middleware.

* **Bug Fixes**
  * Improved handling of mounted sub-router routes to prevent missing prefixed routes and incorrect unprefixed duplicates.

* **Tests**
  * Added integration coverage for helper-based mounts, generated references, placeholders, and related routing behavior.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -96,3 +96,6 @@ testdata/enum_split_const_blocks/enum_split_const_blocks
 
 # a `go build` inside the map-envelope fixture drops its binary beside main.go
 testdata/map_literal_envelope/map_literal_envelope
+
+# a `go build` inside the mount-via-helper fixture drops its binary beside main.go
+testdata/mount_via_helper/mount_via_helper
--- a/generator/testdata_mount_via_helper_test.go
+++ b/generator/testdata_mount_via_helper_test.go
@@ -1,0 +1,78 @@
+// Copyright 2026 Ehab Terra
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package generator
+
+import (
+	"path/filepath"
+	"testing"
+
+	"github.com/ehabterra/apispec/internal/engine"
+)
+
+// TestTestdata_MountViaHelper pins that a Mount written inside a helper still
+// applies its prefix (issue #275).
+//
+// A mount prefix reaches nested routes by tree containment. With the Mount one
+// function deeper the sub-router was built at the CALL SITE, so its routes hang
+// under the caller's argument while the Mount node sits in the helper's body —
+// siblings, not ancestor and descendant. The routes were documented at the wrong
+// paths, with nothing to indicate the path was incomplete.
+func TestTestdata_MountViaHelper(t *testing.T) {
+	cfg := engine.DefaultEngineConfig()
+	cfg.InputDir = filepath.Join("..", "testdata", "mount_via_helper")
+
+	out, err := engine.NewEngine(cfg).GenerateOpenAPI()
+	if err != nil {
+		t.Fatalf("GenerateOpenAPI: %v", err)
+	}
+
+	noDanglingRefs(t, out)
+	noUnresolvedPlaceholders(t, out)
+
+	for _, want := range []string{
+		"/direct/things", // control: mounted at the call site
+		"/api/things",    // literal prefix inside a helper — the fix
+	} {
+		if _, ok := out.Paths[want]; !ok {
+			got := make([]string, 0, len(out.Paths))
+			for p := range out.Paths {
+				got = append(got, p)
+			}
+			t.Errorf("missing %s; documented %v — a Mount in a helper must still apply its prefix", want, got)
+		}
+	}
+
+	// The un-prefixed form must NOT survive alongside the prefixed one: the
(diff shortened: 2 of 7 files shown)


ehabterra:
Valid — added in the commit just pushed.

The change shipped with a fixture and a structural test but nothing exercising `mountRouterSubtree` directly, which the testing conventions require for resolution logic.

The new unit tests pin what the mount context actually carries:

- **the prefix reaches the subtree** — and does *not* also claim the un-prefixed context, which is the duplicate `dropSubsumedMountPrefixes` then has to clean up
- **tags** — a mount with a path names its own, one without inherits what it was given; getting that backwards silently retags every route under a mount
- **dynamic params and middleware** thread through rather than being reset
- an empty target is a no-op

Plus a guard in the other direction, which is the one I would want most in review: `handleMountNode` must **not** take the new parameter hop when the mounted router is an ordinary local. `resolveArgThroughParams` returns the same argument there, and re-walking the mount's own subtree at its own prefix would duplicate every route under it.

One thing I did not do, deliberately: the parameter-resolution hop itself stays fixture-covered. Binding a parameter to a caller's argument in synthetic metadata reproduces the *shape* but not the resolution — the same reason the value-resolution work in #182/#188/#190 uses real fixtures for this class of gap. A unit test there would assert that my hand-built metadata is self-consistent, not that the hop works.

Suite, `-race`, lint, vet clean. No production-code change, so the A/B gate in the description still stands.

PR #307, opened 2026-08-08
docs: document the released binaries, and stop the release rewriting GO_VERSION

ehabterra:
`docs/INSTALLATION.md` described three ways to install, **all of which require a Go toolchain**, and mentioned the pre-built binaries only in a closing paragraph with no commands. Every release publishes a binary per platform; a user without Go had to work that out from the Releases page.

## Corrections, each verified

| claim | reality |
|---|---|
| `brew install ehabterra/tap/apispec` — the **first** macOS instruction | `github.com/ehabterra/homebrew-tap` returns **404**. It fails for everyone. Removed, with a note so it does not come back from memory. |
| "Go 1.24 or later" prerequisite | `go.mod` declares **`go 1.26.0`** — and it is not a prerequisite at all for the binaries |
| `go install …@latest@latest` | typo in the update instructions |
| install script listed as an option | all three of its modes (`go-install`, `source-local`, `source-system`) **build from source**, so it does not help the audience the binaries are for — now stated |

## Found by running the commands as written

The obvious form — `curl -L -o apispec …` — makes the checksum step **fail**, because the published `.sha256` files name the asset:

```
shasum: apispec-darwin-arm64: No such file or directory
```

So the instructions download under the asset's original name and install with `install -m 0755 … /usr/local/bin/apispec`. Verified end to end: `apispec-darwin-amd64: OK`, binary runs, reports `0.5.6`. The macOS arch split and the Gatekeeper `xattr` note came out of the same run (a `darwin-arm64` binary on an x86_64 host fails with "bad CPU type", which is correct behaviour — I checked `file` on both assets to be sure the release was not mislabelled).

Adds the real asset table (the six names actually published), how to pin a tag, and a note that only `apispec` ships as a binary — `apispecui` and `apidiag` are built from source.

## The release bug this turned up

`apispec --version` from **v0.5.6** prints `Go version: 0.5.6`. Root cause is in this workflow:

```bash
sed -i "s/VERSION=\"[^\"]*\"/VERSION=\"$VERSION\"/g" scripts/release.sh
```

`VERSION="` is a substring of `GO_VERSION="`, so the unanchored `/g` also rewrote line 12 — `GO_VERSION="${GO_VERSION:-$(go version …)}"` — baking the release version into the Go-version field. Anchored to `^VERSION=` and dropped the `/g`; confirmed the anchored form leaves `GO_VERSION=` untouched.

The `Makefile` and `cmd/apispec/main.go` were never at fault, which is worth saying because both look like the obvious suspects.

**This cannot be verified without cutting a release** — the next tag will show a real Go version. The documented `--version` example now shows what the binary actually prints, including that the version carries no `v` prefix.

Docs and CI config only; no Go code changes.

## Summary by CodeRabbit

* **Documentation**
  * Updated installation guidance with pre-built binaries as the recommended Go-free option.
  * Added platform-specific download commands, checksum verification, version pinning, architecture details, and troubleshooting guidance.
  * Clarified Go version requirements and installation limitations.
  * Updated instructions for updating, uninstalling, and selecting release assets.

* **Bug Fixes**
  * Improved release automation to update the intended version value without affecting other version settings.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -64,33 +64,47 @@ jobs:
           echo "go_version=$GO_VERSION" >> $GITHUB_OUTPUT
 
       - name: Update version in Makefile
+        # VERSION comes from the pushed tag, so it is untrusted input. It is passed
+        # through env rather than interpolated into the script with ${{ }}, or a tag
+        # named `v1.0.0"; rm -rf /; #` would execute on the runner.
+        env:
+          VERSION: ${{ steps.version.outputs.version }}
         run: |
-          # Update VERSION in Makefile
-          sed -i "s/^VERSION = .*/VERSION = ${{ steps.version.outputs.version }}/" Makefile
+          # `|` as the sed delimiter, not `/`: a git tag may legally contain a
+          # slash, which would otherwise end the substitution early.
+          sed -i "s|^VERSION = .*|VERSION = $VERSION|" Makefile
           
           # Update VERSION in scripts/release.sh if it exists
           if [ -f "scripts/release.sh" ]; then
             # ^VERSION= anchored: "VERSION=" is a substring of "GO_VERSION=", and an
             # unanchored /g rewrote scripts/release.sh line 12 too — which is why
             # v0.5.6 ships reporting `Go version: 0.5.6`.
-            sed -i "s/^VERSION=\"[^\"]*\"/VERSION=\"${{ steps.version.outputs.version }}\"/" scripts/release.sh
+            sed -i "s|^VERSION=\"[^\"]*\"|VERSION=\"$VERSION\"|" scripts/release.sh
           fi
 
       - name: Update version constants in main.go
+        env:
+          VERSION: ${{ steps.version.outputs.version }}
+          COMMIT: ${{ steps.version.outputs.commit }}
+          BUILD_DATE: ${{ steps.version.outputs.build_date }}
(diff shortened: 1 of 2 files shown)


ehabterra:
Valid, and on a line this PR already touched — fixed in the commit just pushed.

`steps.version.outputs.version` is derived from `GITHUB_REF`, i.e. the pushed tag, so it is untrusted input, and it was interpolated straight into `run:` blocks with `${{ }}`. GitHub substitutes the raw string before the shell sees it, so a tag named

```
v1.0.0"; touch /tmp/pwned; #
```

closes the sed expression and runs arbitrary commands on the runner with the release job's token. Anyone who can push a tag can do it.

All four values now arrive through `env:` and are read as ordinary shell variables. Verified rather than assumed — running the patched sed with that tag:

```
VERSION = v1.0.0"; touch /tmp/pwned; #
ls: /tmp/pwned: No such file or directory
```

It lands as literal text and nothing executes.

Two things fixed while in there:

- **sed delimiter `/` → `|`.** A git tag may legally contain a slash, which would otherwise end the substitution early. That is a correctness bug that outlives the injection one, and it would have produced a corrupt Makefile rather than an obvious failure.
- **The build step no longer builds a command line out of these values at all.** The Makefile declares them with `?=`, so the environment already wins — `make release` is enough.

Left alone deliberately: the `${{ }}` on the asset list is inside an action's `with:`, not a shell, so it is not a script-injection vector.

YAML re-validated. No Go code changes in this PR.

PR #309, opened 2026-08-09
feat: Homebrew tap, and install instructions that actually install

ehabterra:
Following the macOS instructions left `./apispec-darwin-amd64` sitting in the working directory — not on PATH, not named `apispec`.

## Why that happened

The recommended **"Verify before installing"** block was a *separate* sequence: it downloaded the asset and then ended in **prose** — "then install it under the name you want, as the commands above do". Anyone reading top to bottom ran that block and stopped. The install command was in a different block, above, for a different architecture.

## The fix

Each platform is now **one copy-paste block** that:

1. detects the architecture (`uname -m` → `amd64`/`arm64`) — no "replace `darwin-arm64` with `darwin-amd64`" step to get wrong
2. downloads the asset **and** its checksum
3. verifies, and says to stop if it fails
4. `install -m 0755` to `/usr/local/bin/apispec`
5. removes the download
6. prints `apispec --version`

Tested as written on darwin/amd64. Also adds a no-`sudo` `~/.local/bin` variant, and a real PowerShell equivalent — the old Windows snippet had **no checksum step at all**.

## Homebrew

```bash
brew install ehabterra/tap/apispec
```

Now the first-listed method. The formula installs the **pre-built release binary**, so it needs no Go toolchain and takes seconds.

Homebrew requires a tap to live in a repo named `homebrew-<name>`, so the formula users install from cannot live here. `release.yml` renders it and pushes to `ehabterra/homebrew-tap` on every tag, with **checksums computed from the artifacts that job just built** — the formula cannot point at a digest the release does not have. The step is `if:`-guarded on `HOMEBREW_TAP_TOKEN`, so releases and forks keep working before the tap exists.

### Drift protection

The `.tmpl` is what ships; the `.rb` is what a human reads and runs `brew audit` against — so they can drift. `internal/packaging` tests that:

- the `.rb` is **exactly** what the template renders
- no placeholder was dropped — a dropped one does not fail the release, it silently ships the **previous** version's digest
- the referenced asset names are the ones the release actually publishes, since Homebrew surfaces a wrong URL only as a 404 on the user's machine

Verified against the real v0.5.6 assets: the template renders **byte-identical** to the committed formula.

## Still needs you (documented in `packaging/homebrew/README.md`)

1. Create a **public** repo `ehabterra/homebrew-tap`
2. Fine-grained PAT, **Contents: read/write** on that repo only
3. Add it here as the secret `HOMEBREW_TAP_TOKEN`
4. Seed `Formula/apispec.rb` once (one `cp` + commit, command in the README)

I have not created the repo — that publishes something new under your account.

Also gitignores downloaded release binaries; one 14MB asset nearly went in with this commit.

## Summary by CodeRabbit

* **New Features**
  * Added Homebrew installation support for macOS and Linux on ARM and Intel.
  * Added checksum-verified binary installation instructions for macOS, Linux, and Windows.
  * Added upgrade, uninstall, and multi-installation guidance.

* **Documentation**
  * Expanded Homebrew setup, verification, troubleshooting, and platform-specific installation guidance.

* **Bug Fixes**
  * Release publishing can automatically update the Homebrew formula with current versions and checksums.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,6 +11,12 @@ permissions:
 jobs:
   release:
     runs-on: ubuntu-latest
+    # Hoisted to the JOB so the tap step can test it in an `if:`. The `secrets`
+    # context is not available in a step-level `if` — written that way the step is
+    # skipped even when the secret IS set, which fails silently and looks like the
+    # tap simply never updates.
+    env:
+      HOMEBREW_TAP_TOKEN: ${{ secrets.HOMEBREW_TAP_TOKEN }}
     permissions:
       contents: write  # Required to create releases and upload assets
       pull-requests: read
@@ -120,10 +126,9 @@ jobs:
       - name: Update the Homebrew tap
         # Skipped automatically when the tap token is not configured, so the
         # release still succeeds on a fork or before the tap exists.
-        if: ${{ secrets.HOMEBREW_TAP_TOKEN != '' }}
+        if: env.HOMEBREW_TAP_TOKEN != ''
         env:
           VERSION: ${{ steps.version.outputs.version }}
-          TAP_TOKEN: ${{ secrets.HOMEBREW_TAP_TOKEN }}
         run: |
           set -euo pipefail
           # Checksums come from the artifacts THIS job just built, so the formula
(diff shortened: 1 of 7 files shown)


ehabterra:
Valid — and it was already fixed in `ad6a1e0`, pushed before this review landed. The comment is against the earlier commit.

I hit the same thing while writing the setup instructions, which was lucky timing: had the token existed first, the step would have been skipped with no error and it would have looked like Homebrew was broken rather than the guard being wrong.

**Fixed differently from the suggestion, deliberately.** The secret is bound at **job** level and the step tests `env.HOMEBREW_TAP_TOKEN != ""`. actionlint's own message lists the allowed contexts for `if` — `"env", "github", "inputs", "job", "matrix", "needs", "runner", "steps", "strategy", "vars"` — so `env` is permitted where `secrets` is not.

Verified with the same linter that flagged it:

```console
$ actionlint .github/workflows/release.yml
# no `context "secrets" is not allowed` error
# remaining output is pre-existing shellcheck info on lines 36/56/112/157 — other steps, untouched here
```

The reason for preferring it over an in-script `exit 0`: with the `if`, the step renders as **skipped** in the Actions UI. Exiting 0 early renders as **passed**, with the real reason buried in the log — so a release that quietly did not update the tap looks identical to one that did. The guard exists precisely for the case where nobody is watching.

Same behaviour otherwise: absent secret → no tap push, release still succeeds, forks unaffected.

Also in this branch since the review: an at-a-glance install/update/uninstall table in `docs/INSTALLATION.md`, so the three commands for each method sit on one row instead of in three separate sections.

PR #310, opened 2026-08-09
docs(homebrew): core-ready formula, and the submission is one-time not per release

ehabterra:
Two things, one of which is a gap I created.

## The core-ready formula never reached main

`43cc1e4` — the source-building `apispec-homebrew-core.rb` and the
"Getting into homebrew-core" docs — was pushed to the #309 branch **after** that
PR was merged, so it stayed on the branch. Cherry-picked here. It is the same
commit, verified before: `brew install --build-from-source` compiles it,
`brew test` passes, `brew audit --new --strict` is clean.

## The submission commands read as a per-release ritual

They are not — that sequence is for the **initial** new-formula PR, once. Now
documented, with evidence rather than assertion:

- **Autobump is the default.** `Formula#autobump?` is true unless a formula calls
  `no_autobump!` (`Library/Homebrew/formula.rb:284`), so an accepted apispec is on
  the autobump list and BrewTestBot opens the version-bump PR itself.
- **Detection needs no `livecheck` block.** Checked against the candidate formula:
  `brew livecheck` already resolves `apispec: 0.5.6 ==> 0.5.6` — it finds the
  current release and agrees. A 0.5.7 tag becomes `0.5.6 ==> 0.5.7`.
- **To push a release in the same day** instead of waiting for the bot:
  `brew bump-formula-pr --version=X.Y.Z apispec` — one command that forks,
  branches, updates `url`/`sha256` and opens the PR.

Net: the ongoing cost of core is close to zero — the one-time submission, plus
reviewing the occasional bot PR when something breaks.

Docs and one new formula file; no Go changes.

## Summary by CodeRabbit

- **New Features**
  - Added a Homebrew Core candidate formula for installing `apispec`.
  - Homebrew installation builds now verify the installed version and test OpenAPI generation.

- **Documentation**
  - Expanded Homebrew guidance with eligibility requirements, source-build details, submission steps, and validation commands.
  - Documented release versioning workflows, including automatic updates and manual bumping.
  - Clarified the outcomes and next steps following Homebrew Core acceptance or rejection.

### DIFF
--- a/packaging/homebrew/README.md
+++ b/packaging/homebrew/README.md
@@ -113,9 +113,66 @@ brew audit --strict --formula ./packaging/homebrew/apispec.rb
 `brew audit` is worth running before a release: it catches a stale `version`,
 an unreachable `url`, and a `sha256` that does not match what the URL serves.
 
-## Why not homebrew-core?
+## Getting into homebrew-core (the formulae.brew.sh search)
 
-homebrew-core has notability requirements (roughly: a meaningful number of
-stars/forks/watchers, or demonstrable wide use). A tap has none and can be
-submitted to core later without changing the install command for anyone who
-already used the tap.
+<https://formulae.brew.sh/formula/> indexes **homebrew-core only**. A tap is never
+listed there, however it is configured — so appearing in that search means getting
+the formula accepted into core.
+
+### Where apispec stands
+
+| requirement | status |
+|---|---|
+| Notable enough — Homebrew's heuristic is 75 stars **or** 30 forks **or** 30 watchers | **83 stars** — clears it |
+| Open-source with an OSI licence | Apache-2.0 |
+| Stable, versioned releases | v0.5.6, tagged and signed |
+| Maintained, not a duplicate of an existing formula | yes |
+| **Builds from source** | this is the work — see below |
+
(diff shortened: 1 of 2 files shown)


ehabterra:
Valid — fixed in the commit just pushed.

Homebrew installs from its API by default, so `brew install --build-from-source apispec` inside the `homebrew/core` checkout ignores the local file. The verification step I documented would have verified **something other than the formula being submitted** — green, and meaningless.

Confirmed in Homebrew's own docs before changing anything; the FAQ is blunt about it:

> if you are editing a core formula or cask you must set `HOMEBREW_NO_INSTALL_FROM_API=1` before using `brew install` or `brew update` otherwise they will ignore your local changes and default to the API

— `docs/FAQ.md:180`, and the same instruction appears in `Adding-Software-to-Homebrew.md:68` and `Formula-Cookbook.md:45`.

**Worth recording why my own testing missed it.** I verified the formula from a *scratch tap*, and a third-party tap is always read from disk — so that run was genuinely valid, and the core flow it was standing in for was not. One test did not cover both, and the formula header now says so instead of implying it did.

Scoped to the core section only: the `ehabterra/tap` instructions do not need the variable, for the same reason.

Issue #282, opened 2026-08-03
dependency_analyzer guesses the project root by byte-wise common prefix and hyphen heuristics, while go/packages reports Module.Path


ehabterra:
## What

`internal/metadata/dependency_analyzer.go` decides which packages belong to the analysed project by guessing, in a chain of four fallbacks, while `go/packages` hands it the answer.

It holds `map[string]*packages.Package` — every element of which carries `pkg.Module.Path`, the module path straight out of `go.mod`. It never reads it. Instead:

```go
// :621 detectProjectRoot — longest common prefix of all package paths, BYTE-wise
commonPrefix = fd.findCommonPrefix(commonPrefix, path)
...
// :648 — if the guess "looks like a domain", guess differently
if len(commonPrefix) < 3 || strings.Contains(commonPrefix, ".") {
	for _, path := range packagePaths {
		parts := strings.Split(path, "/")
		if len(parts) >= 2 && !strings.Contains(parts[0], ".") {
			return parts[0]          // "the first path segment without a dot"
		}
	}
}
```

```go
// :698 fallbackProjectPackageDetection
// "If it contains hyphens or underscores, it's likely a project package"
if strings.Contains(firstPart, "-") || strings.Contains(firstPart, "_") {
	return true
}
// "If it's a simple two-part package that doesn't look like a domain"
if len(parts) == 2 && !strings.Contains(firstPart, ".") {
	return true
}
// last resort
return strings.Count(importPath, "/") >= 2 && !strings.Contains(importPath, "vendor/")
```

`findCommonPrefix` (`:663`) compares bytes, not path segments, so `github.com/acme/billing` + `github.com/acme/billingv2` yields `github.com/acme/billing`, and `github.com/acme/api` + `github.com/acme/app` yields `github.com/acme/ap` — a string that is not a package path, then used as a `strings.HasPrefix` membership test (`:613`).

The type's own doc comments claim the opposite of what the code does:

```go
// :52  "...without hardcoded values, making the system adaptable..."
// :82  "...providing intelligent project package detection without hardcoded values."
```

while `DefaultFrameworkDetectorConfig` (`:108-145`) hardcodes the framework import list, `ExternalPrefixes` hardcodes six vendor prefixes, and `frameworkDetectionOrder` (`:365`) hardcodes `[]string{"gin","echo","fiber","chi","mux","fasthttp"}` a second time.

## Why it matters

[...]

Issue #200, opened 2026-07-20
Add Iris (kataras/iris) framework support


ehabterra:
## Summary
Add route/request/response detection for the [Iris](https://iris-go.com/) web framework (`github.com/kataras/iris`), so apispec can generate specs for Iris apps as it does for gin, echo, chi, fiber, gorilla/mux, and net/http.

## Why
Iris is a popular high-performance Go web framework not currently in the support matrix. Its routing (`app.Get`/`app.Post`/party groups) and context-based request/response API (`ctx.ReadJSON`, `ctx.JSON`, `ctx.Params`) are a distinct wiring style the current config-driven detectors don't cover.

## Scope (follows the framework-addition convention in CONTRIBUTING/CLAUDE.md)
1. `internal/core/detector.go` — detect the `github.com/kataras/iris` import.
2. New `internal/spec/config_iris.go` — pattern configs:
   - **Routes**: `app.Get/Post/Put/Delete/Patch(path, handler)`, party groups (`app.Party("/prefix")`), `.Handle(method, path, handler)`.
   - **Request body**: `ctx.ReadJSON(&dst)` / `ctx.ReadBody` → RequestContext + RequestBodyPatterns.
   - **Response**: `ctx.JSON(v)`, `ctx.JSON(iris.Map{...})`, status via `ctx.StatusCode(code)` → ResponsePatterns + ResponseContext writer type (`iris.Context`).
   - **Params**: `ctx.Params().Get("id")` / path template `{id}` → ParamPattern (map-key style, cf. gorilla/mux #102).
   - **Security/auth** middleware wiring.
3. Register in `cmd/apispec` (and `cmd/apispecui`).
4. `testdata/iris/` fixture (main.go + go.mod) + structural test in `generator/`.
5. README support-matrix row.

## Acceptance
- Routes, methods, request bodies, path/query params, and response schemas resolve for an Iris fixture covering router-level, party-group, and per-route wiring.
- No regression to existing frameworks (full suite + zero-drift).
- Detection is config-driven, not hardcoded (golden rule #5).

ehabterra:
**Verified absent on current `main`.** `internal/core/detector.go` recognizes only `gin-gonic/gin`, `go-chi/chi`, `labstack/echo`, `gofiber/fiber`, `gorilla/mux`, and `net/http` — no `iris`/`kataras` branch anywhere in `internal/` or `cmd/`. So Iris apps are not detected today.

PR #314, opened 2026-08-10
fix(core): single framework registry; drop the knownFrameworks=5 truncation

### LINKED ISSUE #285: The supported-framework list is duplicated in six places, one of them a bare count (knownFrameworks = 5) that silently truncates detection when a sixth is added
## What

The set of supported frameworks is written out six times, in six different shapes, with no shared source:

| # | Location | Shape | Members |
|---|---|---|---|
| 1 | `internal/core/detector.go:78-89` | `switch` on import substring | gin, chi, echo, fiber, mux |
| 2 | `internal/core/detector.go:69` | `const knownFrameworks = 5` | *a count* |
| 3 | `internal/engine/engine.go:614-626` | `switch` → config constructor | gin, chi, echo, fiber, mux |
| 4 | `internal/metadata/dependency_analyzer.go:110-145` | `FrameworkPatterns` map | gin, echo, fiber, chi, mux, http, fasthttp |
| 5 | `internal/metadata/dependency_analyzer.go:365` | `known := []string{…}` (ordering) | gin, echo, fiber, chi, mux, fasthttp |
| 6 | `cmd/apispecui/main.go:106` + `:513-521` | `supportedFrameworks` + `switch` | gin, chi, echo, fiber, mux, net/http |

Plus `ExternalPrefixes` (`dependency_analyzer.go:138`), which repeats the framework *import paths* a third time for exclusion purposes.

They already disagree: fasthttp appears in 4 and 5 but nowhere else; `net/http` appears in 4 and 6 but not 1; the config constructors in 3 and 6 are two independent switches over the same names.

## The concrete hazard

`internal/core/detector.go:69`:

```go
// ImportsOnly: parsing stops after the import block ...
// The loop also stops once every known framework is seen.
const knownFrameworks = 5
...
if len(frameworks) == knownFrameworks {
	break
}
```

The constant is an unlinked restatement of the case count in the switch 20 lines below. Adding a sixth framework — Iris (#200) is already requested — makes the early exit fire at five, silently abandoning the file walk before the sixth can be found. It compiles, all existing tests pass, and the new framework is detected only in projects that happen not to import five others. Nothing ties the constant to the switch.

## Why it's in scope for "no generalization"

CLAUDE.md's contributor path for a new framework is: `internal/core/detector.go` → new `internal/spec/config_<framework>.go` → register in `cmd/apispec` → fixture → README. That is already three edits; in practice it is six, two of which (the count in #2, the ordering list in #5) are invisible unless you go looking, and one of which fails silently.

## Suggested fix

One registry, in one place — e.g. `internal/core`:

```go
type Framework struct {
	Name           string            // "gin"
	ImportPatterns []string          // detection + ExternalPrefixes
	Config         func() *spec.APISpecConfig
	DetectionRank  int               // replaces frameworkDetectionOrder's hand list
}
func Frameworks() []Framework
```

Then:
- `detector.go` ranges it, and the early-exit bound becomes `len(Frameworks())` rather than a literal;
- `engine.defaultFrameworkConfig` and the UI's switch both look up `Config`;
- `dependency_analyzer`'s `FrameworkPatterns`, `ExternalPrefixes` and `frameworkDetectionOrder` derive from `ImportPatterns` and `DetectionRank`;
- `cmd/apispecui`'s `supportedFrameworks` becomes a projection.

Adding Iris then touches the registry, a `config_iris.go`, a fixture, and the README — which is what the contributing guide already promises.

Related: #282 (the same file's other hardcoded lists).

ehabterra:
Closes #285.

## The bug

The supported-framework set was written out six times in six different shapes, one of them a bare count bounding the detector's file walk:

```go
// internal/core/detector.go
const knownFrameworks = 5
...
if len(frameworks) == knownFrameworks {
    break
}
```

That constant was an unlinked restatement of the case count in a `switch` twenty lines below. Adding a sixth framework — Iris (#200) is already requested — makes the early exit fire at five and abandon the walk, so the new framework is detected only in projects that happen not to import five others. It compiles, every existing test passes, and the spec just comes out thinner.

The six lists already disagreed with each other: `fasthttp` appeared in two of them, `net/http` in two different ones, and the engine and the UI each carried their own independent name→config switch.

## The fix

`internal/core/frameworks.go` becomes the single source — detection patterns, dependency-analysis import patterns, external prefixes, detection rank, and whether a default config exists. Everything else projects it:

| Before | After |
|---|---|
| `detector.go` switch + `knownFrameworks = 5` | ranges the registry; the early-exit bound is `len(SourceDetectableFrameworks())` |
| `dependency_analyzer`'s `FrameworkPatterns`, `ExternalPrefixes`, `frameworkDetectionOrder` | derived from the registry |
| two independent name→config switches (engine + UI) | `spec.DefaultConfigForFramework` |
| `supportedFrameworks` literal in the UI | `core.ConfigurableFrameworkNames()` |

`DetectionRank` is explicit rather than derived from declaration order, because the two orderings genuinely differ by history (`gin, echo, fiber, chi, mux` for dependency classification vs `gin, chi, echo, fiber, mux` for the source scan and the UI picker) and silently swapping them would change which framework wins for a package importing two.

Also removes an orphaned doc comment that had drifted onto `ComposeFrameworkConfig`, and adds `spec.DefaultConfigForFramework` to the public `spec` package.

## Behaviour drift: none

Checked rather than assumed. A throwaway test compared the derived `FrameworkPatterns`, the derived `ExternalPrefixes` and `frameworkDetectionOrder()` against the literals they replace — byte-identical, ordering included. It was deleted once it passed; the permanent guards below are what ship. The full suite (metadata goldens, determinism tests, every framework fixture) passes unchanged.

## Guards

- **`TestDetectAllSurvivesANewFramework`** registers one extra framework and asserts the scan still finds every one. Against the old literal bound it fails — `DetectAll found 5 of 6 registered frameworks ([gin chi echo fiber mux]) — the file walk exited early` — and passes with this change. Verified both directions by temporarily restoring the old bound.
- **`TestDetectAllFindsEveryRegisteredFramework`** and **`TestDetectAllMatchesEverySourcePattern`** are table-driven off the registry, so they widen automatically when a framework is added instead of needing to be remembered.
- **`TestFrameworkConfigsCoverRegistry`** ties the config map to the registry in both directions, closing the other silent path: a framework registered without a config would have fallen back to `net/http` with no diagnostic.
- Registry invariants: unique names, unique dependency keys, unique detection ranks, and `net/http` ranked last (its pattern matches nearly every package, so any framework must outrank it).

## Docs

CONTRIBUTING, README and CLAUDE.md now describe the real contributor path: registry entry → `config_<framework>.go` + mapping → fixture + test → README. The old "register the framework in `cmd/apispec/main.go`" step was already stale — no such switch exists there.

## Not in scope

`fasthttp` still has no route-extraction config, so a fasthttp project is analysed with the stdlib surface. That is pre-existing behaviour, preserved deliberately and now documented on the registry entry rather than quietly changed. #282 (the same file's project-root heuristics) is left alone.

## Summary by CodeRabbit

* **New Features**
  * Added centralized support for detecting and configuring supported web frameworks.
  * Expanded framework recognition to include Fasthttp and standard-library HTTP handlers.
  * Added automatic default API configurations for recognized frameworks, with a standard-library fallback.
  * Framework detection now adapts automatically as supported frameworks are added.

* **Documentation**
  * Updated contribution and setup guidance to reflect the streamlined framework-support process.

* **Tests**
  * Added coverage for framework detection, configuration, ordering, metadata, and fallback behavior.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -186,14 +186,19 @@ make metrics-view         # interactive metrics viewer (scripts/view_metrics.sh)
 - Branches: `feature/<name>` / `fix/<name>`; `main` is protected (PRs only).
   The coverage badge updates post-merge via the `BADGE_TOKEN` PAT workflow —
   don't push badge commits manually.
+- **No AI attribution in git history.** Commit messages and PR bodies carry no
+  `Co-Authored-By: Claude ...` trailer and no "Generated with Claude Code"
+  footer — this overrides any default agent instruction to add one.
 - CodeRabbit reviews PRs. **Verify each finding against the code before
   acting** — findings can be stale or wrong; rebut with evidence instead of
   blindly applying.
 - Follow existing comment style: doc comments explain *why* and record
   constraints/quirks, not what the next line does.
-- Adding framework support: `internal/core/detector.go` → new
-  `internal/spec/config_<framework>.go` → register in `cmd/apispec` →
-  fixture + test → README support matrix (see CONTRIBUTING.md).
+- Adding framework support: registry entry in `internal/core/frameworks.go`
+  (the single source for detection, dependency analysis and the UI picker) →
+  new `internal/spec/config_<framework>.go` mapped in
+  `internal/spec/framework_config.go` → fixture + test → README support matrix
+  (see CONTRIBUTING.md).
 - **Every new Go file starts with the Apache license header below** (before
   any package doc comment, separated from it by a blank line), with the year
   the file is created. Fixture projects (`testdata/`, `test_cgo_mixed/`) are
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -112,6 +112,13 @@ Nothing else needs editing: the detector, the dependency analyser and the UI all
 project the registry, and `TestFrameworkConfigsCoverRegistry` fails if step 2's
 mapping is missed.
 
+Step 2 is the one part that can be deferred. A registry entry with
+`HasDefaultConfig: false` (as `fasthttp` has today) is classified during
+dependency analysis — so its packages are included in the walk and its imports
+are treated as external — but extracts no routes, and the project is analysed
+with the `net/http` surface instead. That is a useful half-step, not a
+destination: leave a comment on the entry saying so.
+
 If you're unsure about any step, feel free to ask questions or create a draft PR - I'm happy to help!
 
 ## Submitting Changes
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -103,11 +103,14 @@ This helps us work together effectively and ensures contributions align with the
 
 To add support for a new web framework:
 
-1. **Update framework detection** in `internal/core/detector.go`
-2. **Add the default configuration** in `internal/spec/config_<framework>.go` (each framework lives in its own file alongside `config.go`)
-3. **Register the framework** in `cmd/apispec/main.go`
-4. **Add a fixture project** under `testdata/<framework>/` and a corresponding test case
-5. **Update documentation** in `README.md`
+1. **Add a registry entry** in `internal/core/frameworks.go` — the single source for detection patterns, dependency analysis and the UI picker
+2. **Add the default configuration** in `internal/spec/config_<framework>.go` (each framework lives in its own file alongside `config.go`) and map it in `internal/spec/framework_config.go`
+3. **Add a fixture project** under `testdata/<framework>/` and a corresponding test case
+4. **Update documentation** in `README.md`
+
+Nothing else needs editing: the detector, the dependency analyser and the UI all
+project the registry, and `TestFrameworkConfigsCoverRegistry` fails if step 2's
+mapping is missed.
 
 If you're unsure about any step, feel free to ask questions or create a draft PR - I'm happy to help!
 
--- a/README.md
+++ b/README.md
@@ -1033,10 +1033,9 @@ go test ./internal/spec -v -run "Test.*Comprehensive"
 
 ### Adding a framework
 
-1. Add detection to `internal/core/detector.go`.
-2. Add the default config (route/request/response/param patterns) under `internal/spec/`.
-3. Register the framework in `cmd/apispec/main.go`.
-4. Add a fixture project under `testdata/` and a test case.
+1. Add a registry entry to `internal/core/frameworks.go` (detection patterns, import patterns, detection rank).
+2. Add the default config (route/request/response/param patterns) under `internal/spec/`, and map it in `internal/spec/framework_config.go`.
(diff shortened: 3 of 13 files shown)


ehabterra:
Thanks — verified all three against the code. Two were real and are fixed in d84d4dd; the third I applied in a narrower form.

**1. Deep-copy the nested slices in `Frameworks` / `SourceDetectableFrameworks` — valid, fixed.**

No live bug (`frameworkPatternsFromRegistry` already clones, `frameworkExternalPrefixes` copies values, the detector only reads), but the accessors were shallow-copying while `TestFrameworksReturnsACopy` asserted copy semantics — the test was making a guarantee the code did not. Both accessors now clone `SourcePatterns`, `ImportPatterns` and `ExternalPrefixes`, and the test mutates a nested field through each accessor.

**2. Normalise framework names in the engine composition — valid and reachable, fixed.**

This is the one worth flagging. It is not hypothetical: `buildAPISpecConfig` passes `req.Framework` — a field off the JSON request body — into `ComposeFrameworkConfigWithPrimary` with no validation against the registry. So a client posting `{"framework": "Gin"}` hit exactly the inconsistency you describe: the lookup resolved case-insensitively to the gin config, while the dedup loop and the `primary != "net/http"` check compared with `==`.

Confirmed by running the new test against the previous commit:

```
--- FAIL: TestComposeFrameworkConfigNormalisesPrimaryCase
    primary "GIN": frameworks = [GIN gin], want [gin] — the name was not canonicalised before dedup
    frameworks = [NET/HTTP net/http], want one entry
```

The framework came back twice, the UI's `frameworks[0]` label showed `GIN`, and gin's own patterns were merged back over themselves as a secondary view. `ComposeFrameworkConfigFrom` now canonicalises through `core.CanonicalFrameworkName` before anything compares names, writing into a copy so the caller's slice is not rewritten in place (it is exported and the detected list is shared). Unknown names still pass through unchanged — carrying an explicit house-router choice is documented behaviour that `TestComposeFrameworkConfigWithPrimary` covers.

**3. Make the config step conditional on `HasDefaultConfig` — applied to CONTRIBUTING only.**

Adding a framework in order *not* to extract routes from it isn't a path worth putting in the numbered checklist in three files; that would read as an endorsed option. But the state is real — `fasthttp` sits in it today — so CONTRIBUTING now explains it after the list: an entry with `HasDefaultConfig: false` is classified for dependency analysis (its packages join the walk, its imports count as external) but extracts no routes, and the project is analysed with the `net/http` surface. Framed as a deferred half-step rather than a destination. README and CLAUDE.md keep the four-step path, which is what someone adding a framework actually wants.

On the docstring-coverage warning: every exported identifier added here has a doc comment. The 75% figure counts unexported helpers and test functions, so I've left it.

PR #316, opened 2026-08-10
fix(metadata): classify project packages by module path, not path shape

### LINKED ISSUE #282: dependency_analyzer guesses the project root by byte-wise common prefix and hyphen heuristics, while go/packages reports Module.Path

ehabterra:
Closes #282.

## The bug is bigger than the issue says

The issue describes `findCommonPrefix` mangling `github.com/acme/api` + `github.com/acme/app` into `github.com/acme/ap`. Probing the actual code turned up something worse: for **any domain-hosted module** — i.e. essentially every real Go project — the inference produces no root at all, and every import falls through to the last-resort heuristics.

The chain, verified by direct probe:

1. `findCommonPrefix` is byte-wise, so the common prefix of the loaded packages contains a dot (`github.com/...`).
2. `detectProjectRoot` sees the dot, decides the prefix "looks like a domain", and scans instead for a package path whose first segment has no dot. For a `github.com/...` project there is none, so it returns `""`.
3. `isIntelligentProjectPackage` falls back to `fallbackProjectPackageDetection`, which accepts anything with two slashes, or any two-segment path whose first segment has no dot.

That last rule is the standard library. Running against this repo, the packages classified as project code and appended to `IncludePackages` (both flags default to true) were:

```
compress/gzip   encoding/json   go/ast     go/build   go/constant   go/doc
go/token        go/types        hash/fnv   net/http   path/filepath runtime/debug
```

`go/types` was being treated as project code. The single-word guard above (`fmt`, `os`) catches only stdlib packages without a slash.

## The fix

`go.mod` already answers this, and the engine already reads it — `moduleImportPath()` has been passed to `GenerateMetadataWithLogger` since it was added, just never to the dependency analyser. Now it is: a package is the project's iff it is the module path or under `modulePath + "/"`.

The heuristics stay for the no-module case (a bare directory with no `go.mod`), which is the one place inference is all there is. Two supporting fixes there:

- `findCommonPrefix` compares whole path segments. Byte-wise it returned strings that are not package paths, then used them as `HasPrefix` membership tests.
- With that corrected, the "looks like a domain" branch that discarded every domain-hosted root is gone, so the fallback now resolves `github.com/acme/x` instead of giving up and guessing.

The two doc comments claiming the type provided detection "without hardcoded values" are removed — they described the opposite of the code.

## Measured, not asserted

Against this repo, before vs after:

| | before | after |
|---|---|---|
| Imported packages classified as project | 16 | 4 |
| Auto-included into `IncludePackages` | 25 | 13 |
| Generated spec | 2460 lines | 2460 lines, **byte-identical** |
| Run time | 13.6s | 13.4s (noise) |

Nothing was newly included — the diff is one-directional, and the twelve that went away are exactly the stdlib packages listed above. So this removes wasted analysis without changing a single line of output. I am not claiming a performance win; the timing difference is noise.

## Deliberate test changes

Two existing assertions pinned the old behaviour and were updated with comments explaining why:

- `findCommonPrefix("github.com/a/b", "github.com/a/c")` returned `"github.com/a/"` — a trailing slash is not a path. Now `"github.com/a"`, plus a new case for the `api`/`app` pair from the issue.
- `detectProjectRoot` named `"abc"` the root of `abc/x`, `zzz.io/y`, `qqq.io/z` because it was the one path without a dot. Three paths sharing no segment have no common root; the honest answer is none, and classification falls through to the heuristics, which classify those paths identically. A new case asserts that packages which *do* share a root now resolve it, including domain-hosted ones.

## Coverage

- `dependency_analyzer_module_test.go` — module-path classification, including the sibling-module trap (`github.com/acme/billing` must not claim `github.com/acme/billingv2`, which shares every byte), the same-org-different-module case, and mock exclusion still running first. One test pins the fallback still claiming an unrelated third-party path, which is what makes the wiring load-bearing.
- `testdata/module_prefix_packages/` — the fixture the issue asks for: sibling packages `api/` and `app/` sharing the byte prefix `.../ap` but no segment. Because the fix *narrows* what counts as project code, the structural test asserts nothing legitimate was narrowed away: both packages' routes, their path params, and the request body all survive.

## Known limit, stated rather than guessed

With a module path, "ours" means "under the module". A multi-module workspace (`go.work`) would need `packages.NeedModule` to enumerate sibling modules — `go.work` is not supported anywhere in the codebase today, so this changes nothing, but it is the one case where the old `isPackageImportedByProject` OR-branch could in principle have helped. Re-adding it would re-broaden the classification to "every direct dependency", which is the bug. If workspace support is wanted it should be its own issue.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved framework and project package detection for projects with explicit module paths.
  - Correctly handles module declarations with varied spacing, comments, quoting, and line endings.
  - Prevented incorrect package matching when import paths only share partial segments.
  - Improved project-root detection for domain-hosted packages.
  - Preserved path parameters and correctly resolved POST request bodies in generated API documentation.

- **Tests**
  - Added regression coverage for module parsing, package classification, root detection, and route generation.

### DIFF
--- a/generator/testdata_module_prefix_packages_test.go
+++ b/generator/testdata_module_prefix_packages_test.go
@@ -15,6 +15,7 @@
 package generator
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/ehabterra/apispec/spec"
@@ -77,11 +78,32 @@ func TestTestdata_ModulePrefixPackages(t *testing.T) {
 		}
 	}
 
-	// The request body on POST /widgets resolves through the api package's own
-	// type — a component, not an empty schema.
-	if item, ok := out.Paths["/widgets"]; ok {
-		if op := opFor(item, "POST"); op != nil && op.RequestBody == nil {
-			t.Error("POST /widgets: request body missing")
-		}
+	// The request body on POST /widgets must resolve through the api package's
+	// own Widget type — the api package is one of the two whose classification
+	// this fix changes, so a body that degraded to an empty schema would mean
+	// the package stopped being analysed.
+	item, ok := out.Paths["/widgets"]
+	if !ok {
+		return
+	}
+	op := opFor(item, "POST")
+	if op == nil || op.RequestBody == nil {
+		t.Fatal("POST /widgets: request body missing")
+	}
+	mt, ok := op.RequestBody.Content["application/json"]
+	if !ok || mt.Schema == nil {
+		t.Fatal("POST /widgets: no application/json schema")
+	}
+	if !strings.HasSuffix(mt.Schema.Ref, "Widget") {
+		t.Fatalf("POST /widgets: body should $ref the api package's Widget, got ref=%q type=%q", mt.Schema.Ref, mt.Schema.Type)
+	}
+	// noDanglingRefs covers reachable refs generally; name the component here so
+	// the failure says which type went missing.
+	name := mt.Schema.Ref[strings.LastIndex(mt.Schema.Ref, "/")+1:]
+	if out.Components == nil || out.Components.Schemas == nil {
+		t.Fatal("POST /widgets: body $refs a component but the spec has no components")
+	}
+	if _, ok := out.Components.Schemas[name]; !ok {
+		t.Errorf("POST /widgets: body $refs %q, which is not in components.schemas", name)
 	}
 }
--- a/generator/testdata_module_prefix_packages_test.go
+++ b/generator/testdata_module_prefix_packages_test.go
@@ -1,0 +1,87 @@
+// Copyright 2026 Ehab Terra
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package generator
+
+import (
+	"testing"
+
+	"github.com/ehabterra/apispec/spec"
+)
+
+// TestTestdata_ModulePrefixPackages guards issue #282. The two route-carrying
+// packages are `.../module_prefix_packages/api` and `.../app`, whose import
+// paths share the byte prefix `.../ap` but no whole path segment beyond the
+// module. The dependency analyser used to infer the project root as the
+// byte-wise longest common prefix and then use it as a HasPrefix membership
+// test; it now takes the module path from go.mod.
+//
+// The fix narrows what counts as a project package, so the thing worth
+// asserting is that nothing legitimate was narrowed away: both packages'
+// routes, their params and their bodies must survive.
+func TestTestdata_ModulePrefixPackages(t *testing.T) {
+	out := loadTestdataWithFixtureConfig(t, "module_prefix_packages", spec.DefaultChiConfig())
+	noDanglingRefs(t, out)
+
+	want := map[string][]string{
+		"/widgets/{id}":     {"GET"},
+		"/widgets":          {"POST"},
+		"/sessions/{token}": {"GET"},
+	}
+	for path, methods := range want {
+		item, ok := out.Paths[path]
+		if !ok {
+			t.Errorf("path %q missing; have %v", path, mapPathKeys(out.Paths))
+			continue
+		}
+		for _, m := range methods {
+			if opFor(item, m) == nil {
+				t.Errorf("%s %s: expected operation, missing", m, path)
+			}
+		}
+	}
+
+	// Path params come from the handlers in both sibling packages, so a package
+	// wrongly classified as external would show up here as a missing parameter.
+	for _, tc := range []struct{ path, param string }{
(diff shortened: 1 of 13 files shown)


ehabterra:
Both findings verified and fixed in 04a5daa.

**1. `moduleImportPath` should use the module-file grammar — valid, and worse than it looks.**

The old parser matched a literal `"module "` prefix and returned the rest of the line. Probed against the forms `go.mod` allows:

| go.mod | old result |
|---|---|
| `module github.com/acme/x` | `github.com/acme/x` |
| `module\tgithub.com/acme/x` | `""` — the prefix match needs a space |
| `module github.com/acme/x // the module` | `github.com/acme/x // the module` |
| `module "github.com/acme/x"` | `"github.com/acme/x"` (quotes included) |

All three failures predate this PR, but you're right that it matters more now, and the two directions fail differently: an empty path silently reverts classification to the path-shape heuristics this PR exists to remove, while a path with a comment or quotes glued on matches *no* package, so every import looks external — the opposite error, and the one that could actually drop analysis.

Now `modfile.ModulePath`. `golang.org/x/mod` was already in the module graph via `x/tools`, so it moves from indirect to direct without adding a download. `TestModuleImportPathGrammarForms` covers tab, trailing comment, quoted path, CRLF and a following `require` block; I left the existing `TestModuleImportPath` in `engine_sweep_test.go` to own the empty/missing cases rather than duplicating them.

Re-measured after the swap: the generated spec for this repo is still byte-identical to the pre-fix baseline, and the auto-included count is still 13.

**2. Assert the request-body schema reference — valid, applied.**

Fair: `RequestBody != nil` would have passed on an empty schema. It now asserts the `$ref` ends in `Widget` and that the component exists in `components.schemas`, failing with the component name rather than leaving it to the general `noDanglingRefs` sweep. Worth having precisely because the `api` package is one of the two whose classification this PR changes — a body degrading to an empty schema is the regression this fixture is here to catch.

PR #323, opened 2026-08-11
perf(metadata): resolve lookup ordering once instead of per call

### LINKED ISSUE #322: Determinism ordering is recomputed per lookup: typeInPackage and astFileFromFn sort on every call (6.8% of allocation)
## What

Two lookup helpers re-derive a deterministic iteration order on **every call**, allocating a fresh slice and sorting it each time. Together they are ~6.8% of everything the tool allocates (0.14GB of 2.07GB on this repo generating its own spec).

### 1. `typeInPackage` — 3.7% of all allocation (0.08GB)

`mapper.go:1138`:

```go
func typeInPackage(pkg *metadata.Package, typeName string) *metadata.Type {
	...
	// A type lives in exactly one file, but iterate deterministically anyway.
	var fileNames []string
	for fileName := range pkg.Files {
		fileNames = append(fileNames, fileName)
	}
	sort.Strings(fileNames)
	for _, fileName := range fileNames {
		if typ, exists := pkg.Files[fileName].Types[typeName]; exists {
			return typ
		}
	}
	return nil
}
```

Every lookup allocates a slice of every filename in the package and sorts it, to find a type the comment itself says lives in exactly one file.

### 2. `astFileFromFn` — 3.1% of all allocation (0.06GB)

`metadata.go:1805`, inside a loop over the package's files:

```go
for _, fileName := range metadata.SortedFileNames(pkgName) {
	f := pkg.Files[fileName]
	...
	for _, typeName := range slices.Sorted(maps.Keys(f.Types)) {
```

so one call sorts the type names of every file in the package. The outer order already comes from `metadata.SortedFileNames`, which is the memoized pattern this inner sort should follow.

## Root cause

Determinism is a hard invariant here (golden rule #1: any map iteration that can reach the output must be sorted), and both sites honour it correctly. The cost is *where* the ordering is established: recomputed per lookup, rather than once per package into an index that lookups read.

The ordering these two produce is identical on every call for a given package — it depends only on the package's contents, which do not change after metadata generation.

## Suggested direction

Both want the same thing: an ordered index built once.

- `typeInPackage`: a per-package `map[typeName]*Type` (or `map[typeName]fileName`) built lazily on first lookup and cached on the `Package`. Then the lookup is a map hit and the ordering question disappears — a type lives in one file, so there is nothing to disambiguate. If two files in a package could ever declare the same type name (they cannot in valid Go), the index build is the one place to define the tie-break, deterministically, once.
- `astFileFromFn`: memoize the sorted type names per file the same way `SortedFileNames` memoizes the file list — or, better, build the same name→declaration index and drop the nested scan entirely.

Neither changes output; both should ship with the goldens and determinism tests (`TestGenerateMetadataDeterministic` / `TestGenerateDeterministic`) unchanged, which is exactly the guard that makes this safe to do.

## Note

This is the smallest of the performance issues from this profiling pass, and unlike the others it is self-contained: it touches two functions, needs no design decision, and cannot affect what is documented. Good candidate to land first.

Measured on this repo (36 routes, 16 packages) at default limits, after #225/PR #317.

ehabterra:
Closes #322. First of the five performance issues from the profiling pass, picked because it is the self-contained one: two functions, no design decision, and it cannot change what is documented.

## What was happening

`typeInPackage` allocated and sorted every file name in a package on **every lookup**, to find a type its own comment says lives in exactly one file:

```go
// A type lives in exactly one file, but iterate deterministically anyway.
var fileNames []string
for fileName := range pkg.Files { fileNames = append(fileNames, fileName) }
sort.Strings(fileNames)
```

Its caller falls back to scanning *all* packages for a bare type name, so that cost was paid per package per unresolved name. `astFileFromFn` did the same thing one level down — `slices.Sorted(maps.Keys(f.Types))` inside its loop over the package's files, per call.

Determinism is the reason both sorted (golden rule #1), and both were correct. The problem was only *where* the order got established: recomputed per lookup rather than once into an index.

## The fix

`Metadata.TypeInPackage` and `Metadata.SortedTypeNames`, both modelled on the existing `SortedFileNames` — same RWMutex discipline, same rebuild-when-it-grows guard, since these lookups can run while metadata is still being assembled. `TypeInPackage` fingerprints `(files, types)` rather than file count alone, because a type can be added to a file that already exists.

**The order is preserved, not just the answer.** The index is filled by walking `SortedFileNames` with an earlier file winning — exactly what the replaced scan did — so a name declared in two files resolves to the same declaration it did before. Go forbids that case, but the old code had a defined answer and this must not quietly pick the other one; `TestTypeInPackageMatchesSortedScan` asserts the index against the scan it replaces, that case included.

## Measured

On this repo generating its own spec:

| | before | after |
|---|---|---|
| total allocated | 2209 MB | **1995 MB** (−9.7%) |
| `typeInPackage` | 79.05 MB (3.58%) | gone from the profile |
| `slices.Sorted` (via `astFileFromFn`) | 66.52 MB (3.01%) | gone from the profile |
| wall clock | 9.35s | 9.09s (~3%) |
| generated spec | — | **byte-identical** |

## A note on how the timing was measured

Measured sequentially, this change first looked like a **15% regression** — and the profile attributed the extra time to `GetChildren`, which this PR does not touch. That inconsistency was the tell: the three runs trended upward (10.3 → 10.7 → 11.5s), i.e. machine drift, not code.

Interleaving base and change within each repetition removed it, and the ordering then reversed consistently across all four pairs:

```
base=9.670  change=9.064
base=9.721  change=9.494
base=9.104  change=9.038
base=8.918  change=8.753
```

Flagging it because the allocation numbers are the reliable ones here; the ~3% wall clock is real but close to this machine's noise floor, and I would not want it quoted as a firm figure.

## Coverage

- `TestTypeInPackageMatchesSortedScan` — index vs the sorted scan it replaces, over present, absent and duplicated names.
- `TestTypeInPackageInvalidatesOnGrowth` — a type added to an existing file, and a whole new file, must both become visible; this is what the `(files, types)` fingerprint is for.
- `TestSortedTypeNames` — order, caching, per-file scoping, growth invalidation, unknown package and unknown file.
- `TestLookupCachesAreRaceFree` — the two caches hammered concurrently; the value is under `-race`, which CI runs. Passes locally under `-race` too.
- `TestTypeInPackage_NilPackage` in the spec sweep covered the deleted function's nil guard; replaced with the equivalent against the new API rather than dropped.

Full suite passes unchanged — the goldens and determinism tests are the drift check for a change like this, and none moved.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved type lookup consistency across packages and files.
  - Ensured lookups correctly reflect newly added files and types.
  - Preserved deterministic type resolution using sorted file and type names.
  - Improved reliability during concurrent metadata access.

- **Tests**
  - Added coverage for cache invalidation, missing types and packages, sorted lookups, and concurrent access.

### DIFF
--- a/internal/metadata/lookup_cache_test.go
+++ b/internal/metadata/lookup_cache_test.go
@@ -354,3 +354,32 @@ func TestLookupCachesAreRaceFree(t *testing.T) {
 	}
 	wg.Wait()
 }
+
+// TestLookupCachesTolerateNilFiles covers a nil entry in Package.Files. Every
+// reader of that map in the spec layer skips nil, and the index build here
+// does too — the shape fingerprint did not, so computing it panicked before a
+// single lookup could run. Metadata is deserialised from YAML as well as
+// generated, so a package need not be as well-formed as the generator makes it.
+func TestLookupCachesTolerateNilFiles(t *testing.T) {
+	present := &Type{Name: 1}
+	m := &Metadata{StringPool: NewStringPool()}
+	m.Packages = map[string]*Package{
+		"app": {Files: map[string]*File{
+			"a.go": {Types: map[string]*Type{"Present": present}},
+			"b.go": nil,
+		}},
+	}
+
+	if got := m.TypeInPackage("app", "Present"); got != present {
+		t.Errorf("TypeInPackage = %p, want %p", got, present)
+	}
+	if got := m.TypeInPackage("app", "Absent"); got != nil {
+		t.Errorf("absent type = %p, want nil", got)
+	}
+	if got := m.SortedTypeNames("app", "b.go"); got != nil {
+		t.Errorf("SortedTypeNames of a nil file = %v, want nil", got)
+	}
+	if got := m.SortedTypeNames("app", "a.go"); !reflect.DeepEqual(got, []string{"Present"}) {
+		t.Errorf("SortedTypeNames = %v, want [Present]", got)
+	}
+}
--- a/internal/metadata/lookup_cache_test.go
+++ b/internal/metadata/lookup_cache_test.go
@@ -17,6 +17,7 @@ package metadata
 import (
 	"go/types"
 	"reflect"
+	"sync"
 	"testing"
 )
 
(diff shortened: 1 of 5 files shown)


ehabterra:
Verified and fixed in the pushed commit.

**`shapeOf` panics on a nil `Package.Files` entry — valid.** Reproduced before fixing:

```
--- FAIL: TestLookupCachesTolerateNilFiles
panic: runtime error: invalid memory address or nil pointer dereference
```

`len(f.Types)` dereferences `f`, so the fingerprint blows up before a single lookup runs.

What makes it clearly worth fixing rather than arguing about reachability: **the guard was already there ten lines below**, in the index build (`if f == nil { continue }`), so the two halves of the same function disagreed with each other. And the convention is established elsewhere — `context_provider.go`, `security.go`, `wrapper_params.go`, `wrapper_detect.go` and `func_field_dispatch.go` all skip nil file entries. The inconsistency was mine, not the convention's.

On reachability: the generator never stores a nil `*File` (`metadata.go:371` assigns a locally constructed one), but `Metadata` is also deserialised from YAML, where a package need not be as well-formed. Worth noting the replaced `typeInPackage` dereferenced `pkg.Files[fileName].Types` just as unguardedly, so this closes a pre-existing exposure rather than one this PR introduced — but it should not have been carried forward either way.

`TestLookupCachesTolerateNilFiles` covers `TypeInPackage` (present and absent names) and `SortedTypeNames` (of the nil file and of a real one). Full suite and `-race` still pass; the measurements in the description are unaffected, since the change only adds a skip on a branch that previously crashed.

Issue #329, opened 2026-08-11
A type can be qualified with the package that uses it, so time.Time resolves as an unknown external type


ehabterra:
## What

A type from another package can reach the mapper qualified with the package that *uses* it rather than the package that *declares* it. Found while fixing #325, on apispec's own codebase:

```
PROBE-RAW "github.com/ehabterra/apispec/cmd/apispecui-->time.Time"
```

`time.Time` is declared in `time`. The string says it lives in `cmd/apispecui`, which is merely where the value was used — the internal `TypeSep` ("-->") joins the wrong package to the type.

## Why it matters

Every downstream lookup is keyed on that name, so all of them miss:

- the well-known external-type table is keyed `time.Time`, so `time.Time` does **not** resolve to `{type: string, format: date-time}`;
- `cfg.ExternalTypes` matching (`externalType.Name == strings.ReplaceAll(typeName, TypeSep, ".")`) compares against `github.com/ehabterra/apispec/cmd/apispecui.time.Time`, which no user would ever write in config;
- `findTypesInMetadata` cannot find a declaration under that package either.

The type therefore falls through to "unresolved external". Before #325 that produced a dangling `$ref` (`github_com_ehabterra_apispec_cmd_apispecui_time_Time`, one of the two in apispec's own spec). After #325 it produces a resolvable but wrong component:

```yaml
github_com_ehabterra_apispec_cmd_apispecui_time_Time:
  type: object
  description: 'External or unresolved type: github.com/ehabterra/apispec/cmd/apispecui-->time.Time'
```

A timestamp field documented as an empty object is worse than it looks: clients generated from this spec get `object` where the API sends an RFC-3339 string. And the same mechanism will mis-qualify **any** cross-package type it touches — `time.Time` is just the one visible in a small codebase.

## Where to look

The qualification happens before the mapper: the string arrives already joined. Candidates are the argument renderer's qualification tail (CLAUDE.md calls this one of the two deliberate string-surgery islands) and wherever a field or argument type is composed with its enclosing package rather than the type's own.

Two constraints for whoever picks it up:

- Golden rule #2 applies. The right fix resolves the package from the parsed `typemodel.TypeRef`'s core rather than gluing a package prefix onto a rendered string; the correct package is already known at the point the type is recorded.
- Golden rule #10 is the likely trap: for a cross-package selector the package lives on the selector, not the ident, and reading the wrong one is exactly how a using-package prefix gets attached.

## Reproduction

```
[...]

Issue #326, opened 2026-08-11
Ref-eligibility gate misses fixed-size arrays and untyped constants, emitting components that are never defined


ehabterra:
## What

`canAddRefSchemaForType` decides whether a type may be a `$ref` target. It excludes primitives, `[]`-prefixed slices, `map[`, `_nested` names and anonymous structs. Two shapes slip through and produce components that are never defined — the same invalid-spec symptom as #259, which fixed this class for `[][]T` only.

Both reproduce on gitea at default limits:

```
$ref: '#/components/schemas/gitea_dev_routers_web_user__2int64'
$ref: '#/components/schemas/untyped-bool'
```

## Cause 1 — fixed-size arrays `[N]T`

`mapper.go:3056`:

```go
if metadata.IsPrimitiveType(key) || strings.HasPrefix(key, "[]") || strings.Contains(key, "map[") {
	return false
}
```

`[]T` is excluded; `[2]T` is not. From `gitea/routers/web/user/heatmap.go:36`:

```go
data := make([][2]int64, len(hdata))
```

The element type `[2]int64` is ref-eligible by this gate, so it becomes a component name — and `schemaComponentNameReplacer` maps `[`→`_` and `]`→`""`, rendering `[2]int64` as `_2int64`. Glued to the package path that reaches the mapper, the result is `gitea_dev_routers_web_user__2int64`: a component name for a type that is structurally an array and can never be declared.

Note the `HasPrefix(key, "[]")` test is also defeated by a package qualifier — a container type that arrives with a path prefix (`pkg.[]T`) passes the gate too. Matching on the parsed type rather than the string would settle both; the structured type model (`internal/typemodel`) already distinguishes arrays and slices from named types, and golden rule #2 says not to answer this question with string prefixes in the first place.

## Cause 2 — untyped constants

go/types renders an untyped constant's type as `untyped bool` / `untyped int` / `untyped string`. That string is not a primitive by `IsPrimitiveType`, contains no `[]` or `map[`, and so is ref-eligible. The space becomes `-` in the component name, giving `untyped-bool`.

In gitea it shows up on a literal response body:

```yaml
properties:
  ok:
    $ref: '#/components/schemas/untyped-bool'
```

`ok: true` should be `type: boolean`. An untyped constant has a well-defined default type (bool, int, rune, float64, complex128, string), so the mapping is exact, not a guess.

## Suggested fix

[...]

PR #330, opened 2026-08-11
fix(spec): keep the components a route's $refs point at

### LINKED ISSUE #325: $ref emitted for dependency types that have no declaration, so the component is never generated and the spec fails to resolve
## What

The generated spec references components that are never defined, so every consumer that resolves `$ref`s (Swagger UI, Redoc, `openapi-generator`) errors out:

```
Resolver error at requestBody.content.application/json.schema.$ref
Could not resolve reference: Invalid object key
"google_golang_org_protobuf_types_known_timestamppb_Timestamp" at position 2 in
"/components/schemas/google_golang_org_protobuf_types_known_timestamppb_Timestamp":
key not found in object
```

Reported on a project using protobuf well-known types (`timestamppb.Timestamp`, `wrapperspb.StringValue`). Reproduced on gitea, which yields three of them from the same cause:

```
148 distinct $refs, 165 components, 4 DANGLING
  - github_com_golang-jwt_jwt_v5_RegisteredClaims
  - xorm_io_xorm_convert_Conversion
  - gitea_dev_routers_web_user__2int64      (separate cause, filed separately)
  - untyped-bool                            (separate cause, filed separately)
```

`jwt.RegisteredClaims`, `xorm.Conversion`, `timestamppb.Timestamp` and `wrapperspb.StringValue` are all the same shape: **a type from a dependency module**.

## Root cause

`apispec` loads `./...` from the module root, so a dependency's type declarations are never in `metadata.Types`. A component can only be generated from a declaration. But the ref is emitted anyway — and the function that emits it says so in its own comment (`mapper.go:3285`):

```go
func addRefSchemaForType(goType string) *Schema {
	// For custom types not found in metadata, create a reference
	goType = strings.TrimPrefix(goType, "*")
	return &Schema{Ref: refComponentsSchemasPrefix + schemaComponentNameReplacer.Replace(goType)}
}
```

"For custom types **not found in metadata**" is precisely the condition under which no component will ever exist. `canAddRefSchemaForType` passes them because they are neither primitives, containers, `_nested`, nor anonymous structs — i.e. they look like ordinary named types.

The codebase already knows this failure mode and guards one instance of it. `mapGoTypeToOpenAPISchema` (`mapper.go:2655`):

```go
// Inline external types (uuid, decimal, …) are leaves with no component,
// so they are never a real cycle and a $ref to them would dangle. Exclude
// them from the cycle/recursion guards below so every occurrence inlines.
inlineExternal := isInlineExternalType(derivedGoType, cfg, meta)
```

and the same reasoning again at the struct-field site (`mapper.go:1409`). But `isInlineExternalType` only recognises types listed in the `ExternalTypes` **config registry** — `uuid.UUID`, `decimal.Decimal`, `gin.H` and friends. An external type nobody has registered gets no such treatment, and dangles.

So the rule is currently "external types are safe **if someone enumerated them**", where it needs to be "a type with no declaration cannot be a `$ref` target".

## Suggested fix

Gate on the fact rather than on the registry: before emitting a `$ref`, require that the type has a declaration in metadata (or is already a registered component). If it does not, inline whatever is known instead — `unresolvedExternalPlaceholder` already exists for this purpose and produces an honest empty-object schema rather than a broken pointer.

That single check subsumes the registry special-case and covers every dependency type without anyone having to enumerate protobuf, jwt, xorm, and the next library.

Two notes for whoever picks it up:

- Golden rule #4 applies: the decision belongs in the spec layer, and inline types must skip **every** `$ref` fast-path — there are 10+ call sites of `addRefSchemaForType`, and the cycle guard, the recursion guard, the struct-field branch and the array/map element branches each reach it by a different route. Fixing one is what left the others.
- Well-known protobuf wrappers deserve real schemas, not just `{}` — `timestamppb.Timestamp` is an RFC-3339 string in protojson, `wrapperspb.StringValue` a nullable string. Worth adding to the default `ExternalTypes` registry as a follow-up, but that is a quality improvement on top of the correctness fix, not a substitute for it: the next unregistered dependency type would dangle again.

Reproduction: any project whose handler signature or DTO field references a dependency type. `gitea` at default limits reproduces it with `jwt.RegisteredClaims` and `xorm.Conversion`.

ehabterra:
Closes #325.

## The bug

`mapGoTypeToOpenAPISchema` returns two things: the schema for a type, and the components that schema's `$ref`s point at — including the placeholder it registers for a type it cannot resolve. Five callers took the schema and threw the map away:

```go
schema, _ = mapGoTypeToOpenAPISchema(route.UsedTypes, bodyType, route.Metadata, r.cfg, nil)
```

The references survived; their targets did not. A type declared outside the analysed module has no metadata entry and so is always the unresolvable case, which is why dependency types are what surfaced — and why any consumer that resolves `$ref`s refuses the document:

```
Could not resolve reference: Invalid object key
"google_golang_org_protobuf_types_known_timestamppb_Timestamp" ... key not found in object
```

**`mergeRouteExtraction` was the second half of the same hole.** A route reached through two traversal contexts merges the second extraction into the first, and `UsedTypes` was not among the merged fields — so whichever extraction lost the merge took its recorded types, and their components, with it. Unlike the schema slots it merges by preference, this one is a union: there is no "better" answer between two records of what was referenced.

## Measured

Dangling references, before → after:

| | before | after |
|---|---|---|
| `testdata/external_type_refs` (new fixture) | 1 | **0** |
| apispec's own spec | 2 | **0** |
| gitea | 4 | **0** |

Full suite passes; goldens and determinism tests unmoved.

## How it was found

Rather than reason about which of the 10+ `addRefSchemaForType` call sites was at fault, I built the fixture first and traced the emission with a stack-printing probe. That mattered — the path was not the one I would have guessed, and it took three rounds:

1. `ExtractResponse` → discarded map (the reported case, jwt/xorm/protobuf).
2. `mapLiteralSchema` and `oneOfSchemaFor` → same discard, different callers. Found because apispec's own spec still had a dangling ref after fixing (1).
3. `mergeRouteExtraction` dropping `UsedTypes` → found because gitea still had one after fixing (2), and the probe showed the placeholder *was* being recorded but the map it landed in never reached component generation.

## A rejected approach worth recording

My first attempt made `addRefSchemaForType` mark the type in `usedTypes` itself, so every ref site would record centrally — one change covering all call sites, present and future.

That is wrong, and the test suite said so immediately: `usedTypes` membership **drives control flow** in `generateStructSchema`, which branches on whether a field type is already present. Marking a type there turned recursive types into `$ref`s and lost their properties (`TestTestdata_RecursiveTypes`: `property "products" missing`, `property "parent" missing`). The map is not a passive record, so the fix keeps to propagating the schemas the mapping actually produced.

## Known limit, filed not papered over

The placeholder is honest, not correct. `time.Time` reaching the mapper as `<caller pkg>-->time.Time` still resolves to a placeholder rather than `{string, date-time}`, because the qualification is already wrong before it arrives — filed as #329. Adding a well-known-table entry under the mis-qualified key would be the heuristic golden rule #9 forbids: it would fix this one symptom and leave every other cross-package type mis-qualified.

Similarly, #326 (fixed-size arrays, untyped constants) now produces *resolvable* placeholders rather than dangling refs, but `[2]int64` should be an array schema and `untyped bool` a boolean. This PR makes the document valid; those two make it right.

## Coverage

`testdata/external_type_refs` carries fields of external named types of several shapes (a map type, structs, pointers, an embedded one) plus a locally declared type. The structural test asserts `noDanglingRefs`, that the locally declared type still resolves to a real schema with its fields — the fix must not turn everything into a placeholder — and that the external ones carry the unresolved-external description, which is what tells a user which dependency to register in `externalTypes`.

## Summary by CodeRabbit

- **New Features**
  - Improved API schema generation for route request bodies, responses, parameters, map values, and polymorphic types.
  - External or unresolved types now receive descriptive placeholder schemas instead of dangling references.
  - Nested schemas are preserved consistently when routes are combined.

- **Bug Fixes**
  - Prevented referenced schemas from being lost during route extraction and merging.
  - Improved retention of properties defined in test fixtures and referenced types.
  - Added regression coverage for external type references and unresolved schema handling.


ehabterra:
Applied in the pushed commit.

**Assert the HTTP method, not just the path — valid.** A route that lost its method still produces a path entry, so existence alone would not notice it. The fixture registers `GET /accounts/{id}` and `POST /embedded`, and the test now asserts both, keeping the missing-path diagnostic that names the paths actually found.

While this was open I also landed the follow-up this PR's description flagged as a known limit: #329, the mis-qualification that made `time.Time` arrive as `<caller pkg>-->time.Time`. That is PR #331, and it is worth reading together with this one — this PR made such references *resolvable*, and that one makes `time.Time` resolve to `{string, date-time}` instead of an unresolved-external placeholder. Neither supersedes the other: this PR is the general guarantee that no `$ref` is emitted without a component, and #331 removes one reason a type failed to resolve in the first place.

The two are independent branches off main and touch different files (`mapper.go`/`extractor.go`/`pattern_matchers.go` here, `context_provider.go` there), so they can merge in either order.

Issue #259, opened 2026-07-29
A [][]T struct field emits a $ref to a component that is never defined, producing an invalid spec


ehabterra:
A struct field whose type is a **nested slice** (`[][]T`) is emitted as an array whose `items` is a `$ref` to a component name that apispec never defines. The result is a dangling reference, so the generated document is not a valid OpenAPI spec and strict consumers refuse to load it.

A single-level slice (`[]T`) on the same struct is emitted correctly, which makes the bug easy to miss — only the nested one is broken.

## Version

Reproduced on **v0.5.4** and **v0.4.0**, so this is long-standing rather than a regression.

## Minimal reproduction

`main.go`:

```go
package main

import (
	"encoding/json"
	"net/http"

	"github.com/go-chi/chi/v5"
)

type Table struct {
	Header []string   `json:"header"` // []string   — correct
	Rows   [][]string `json:"rows"`   // [][]string — dangling $ref
}

func getTable(w http.ResponseWriter, r *http.Request) {
	_ = json.NewEncoder(w).Encode(Table{})
}

func main() {
	r := chi.NewRouter()
	r.Get("/table", getTable)
	_ = http.ListenAndServe(":8080", r)
}
```

```
go mod init example.com/repro && go get github.com/go-chi/chi/v5
apispec -o out.yaml
```

No config file needed — defaults reproduce it.

## Actual output

```yaml
components:
  schemas:
    example_com_repro_Table:
      properties:
        header:
          type: array
          items:
            type: string                    # correct
        rows:
          type: array
          items:
            $ref: '#/components/schemas/example_com_repro__string'   # never defined
```

`example_com_repro__string` occurs **exactly once** in the whole document — the reference above. The only component actually defined is `example_com_repro_Table`.

Note the doubled underscore in `example_com_repro__string`: it looks like the anonymous inner `[]string` gets a synthesized component name (empty type name + element type), the reference is written, but the corresponding schema is never emitted.

## Expected output

The inner slice should be inlined, the same way the outer one is:

```yaml
        rows:
          type: array
          items:
            type: array
            items:
              type: string
```

(Equivalently, emit the referenced component — but inlining matches how `[]string` is already handled and needs no extra component.)

## Impact

The document fails strict `$ref` resolution. Concretely, `openapi-typescript` 7.13.0 aborts on it:

```
✘  Can't resolve $ref at #/components/schemas/example_com_repro_Table/properties/rows/items
Error: Can't resolve $ref ...
```

[...]

PR #332, opened 2026-08-11
fix(spec): resolve fixed-size arrays and untyped constants structurally

### LINKED ISSUE #326: Ref-eligibility gate misses fixed-size arrays and untyped constants, emitting components that are never defined

ehabterra:
Closes #326.

## Two shapes that cannot be components were registered as ones

**Fixed-size arrays.** A container has no package, but the argument renderer attaches one anyway, so `[2]int64` reaches the mapper as `pkg-->[2]int64`. The ref gate tests the whole key for a `[]` prefix or `map[`, and a qualifier defeats both — so the array looked like an ordinary named type. That is gitea's heatmap (`data := make([][2]int64, ...)`) arriving as a component named `..._2int64`.

**Untyped constants.** go/types renders these as `untyped bool` / `untyped int` / `untyped string`, which are not primitives *by name*, so `ok: true` in a map-literal response became a component named `untyped-bool`.

## The fix

- The gate judges the **parsed core** rather than the raw key: a core name that is itself a container, or an untyped constant, has nothing to register a component under whatever it is qualified with.
- The mapper normalises both on entry — an untyped constant becomes the type Go defines it to default to, a mis-qualified container loses the package it was never in — so the shape is visible to the structural handling that was already there, and already correct for the unqualified forms.

The untyped mapping is derived, not guessed: the Go spec fixes the default type of each constant kind, so `true` is a `bool`.

## Measured against current main (which has #330 and #331)

| | before | after |
|---|---|---|
| apispec's own spec | components `untyped-bool`, `untyped-int` | gone — `boolean`, `integer` inline |
| fixture `/heatmap` (`[][2]int64`) | `$ref: ..._2int64` | array of 2-bounded arrays of integer |
| dangling refs | 0 | 0 |

Dangling was already 0 on both sides because #330 gives every reference a target. **So this PR is not about validity — it is about the schema being right.** A boolean documented as an object is something a generated client gets wrong just as surely as a broken `$ref`, and it is quieter.

## Coverage

`testdata/array_untyped_schemas` carries:

- gitea's shape verbatim — a local `[][2]int64` variable encoded directly. This mattered: my first fixture used a **struct field** of that type, which already resolved correctly and reproduced nothing. The trigger is the value arriving through an argument.
- the three untyped kinds (`bool`, `int`, `string`) via a map literal;
- as controls that must keep working: a `[3]string` struct field, the `[][]T` shape #259 fixed, and a slice of a **named** array type (`type Point [2]int64`) which *is* declared and so legitimately keeps its component.

The test asserts the resolved shapes — element type, `minItems`/`maxItems` from `[2]`, the three default types — not merely the absence of the bad names. It fails on main with exactly the reported components.

## Noted, not fixed here

`type Point [2]int64` resolves to `{type: object}`. A named type whose underlying type is an array should carry the array shape. That is pre-existing, unchanged by this PR (verified before and after), and a different defect — filed as #333 rather than widened into this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved schema generation for fixed-size arrays, untyped constants, and `rune` values.
  * Prevented invalid or dangling schema references for inline values and package-qualified container types.
  * Preserved valid references to named array types and expected response structures.

* **Tests**
  * Added regression coverage for array, container, and untyped-constant schema resolution.


ehabterra:
All three verified and fixed in the pushed commit.

**1. `untyped rune` produced no schema — valid, and worse than described.** Checked before fixing:

```
rune         -> <nil>
int32        -> integer
byte         -> integer
```

`rune` was not in the mapper's integer case at all, so it fell through to the default and produced nothing. My normalisation inherited that: the component name was gone but the property had no type.

One correction to the suggested fix, which offered "either add rune there or normalise it to int32". I did the first, not the second: the Go spec names **rune** as that constant kind's default type, so normalising to `int32` would make the table diverge from the spec it is transcribing. `rune` is an alias for `int32`, so adding it to the integer case is exact — and it fixes the bare `rune` field too, which was broken before this PR.

**2. `isContainerCoreName` used string tests — valid, and fixable without touching typemodel.** The suggestion allowed for updating the type model if it "does not preserve qualified container structure". It doesn't need to: the qualified form parses as a named type whose core *name* is the container, and parsing **that name** recovers the kind structurally.

```
Parse("[2]int64").Kind       = KindArray
Parse("map[string]int").Kind = KindMap
Parse("[]string").Kind       = KindSlice
Parse("User").Kind           = KindNamed
```

So the judgement is the type model's rather than a prefix guess, with no parser change and no new representation to keep in step.

**3. Unit tests at the mapper layer — valid, added.** `untypedConstantDefault` against the Go spec's default types, that each default **reaches a real schema** (exactly the check that would have caught the rune gap, rather than testing the mapping table against itself), and `unqualifyContainer` over qualified containers, qualified names and bare types.

The gate's own cases went into the existing `TestCanAddRefSchemaForTypeWithMap` table rather than a parallel test — it already owns primitives, slices, maps, qualified maps and `_nested`, and the shapes this PR adds belong beside them.

Generated output is unchanged by this commit (the self-spec is identical, `untyped` count still 0); full suite and lint pass.

PR #334, opened 2026-08-11
fix(metadata): record the underlying type of named container types

### LINKED ISSUE #333: A named type whose underlying type is an array resolves to {type: object} instead of the array shape
## What

A named type whose underlying type is an array resolves to an empty object:

```go
type Point [2]int64
```

```yaml
github_com_..._Point:
  type: object
```

It should carry the array's shape — `{type: array, items: {type: integer}, minItems: 2, maxItems: 2}` — which is exactly what the *anonymous* form already produces. In the same fixture (`testdata/array_untyped_schemas`), a `[2]int64` used directly resolves correctly while `Point` does not, so the schema a field gets depends on whether the author named the type.

## Why it is separate from #326

#326 was about shapes that must **not** be components (a bare `[2]int64` has no name to register under). `Point` is declared, so a component is right — the component's *content* is wrong. Verified unchanged by PR #332 before and after, so it is pre-existing rather than introduced.

## Cause, as far as I traced it

The component is generated from the metadata `Type` for `Point` through `generateSchemaFromType`, which dispatches on the declaration's kind. The struct path is well covered; a declaration whose underlying type is an array, slice or map appears to fall through to the object default rather than being resolved as its underlying type.

Worth checking the same question for the other underlying kinds while in there, since they share the path:

```go
type Point  [2]int64          // observed: {type: object}
type IDs    []string          // ?
type Lookup map[string]int    // ?
type Count  int               // ? (a named primitive)
```

A fixture with all four, asserting each carries its underlying shape, would settle it and guard the lot.

## Impact

Lower than #325/#326 — the reference resolves and the document is valid — but a client generated from `{type: object}` for what is really `[2]int64` will get the field wrong, and named container types are common (`type Matrix [][]float64`, `type Headers map[string][]string`).

Found while fixing #326; see `testdata/array_untyped_schemas` for a ready reproduction — `Point` is already declared there as a control.

ehabterra:
Closes #333.

## The cause is in metadata, not the mapper

The issue guessed the mapper's `default:` branch. It is really one layer down: `processTypeKind` classifies a declaration by its AST node and records a **target** only for `*ast.Ident`.

```go
case *ast.Ident:
    t.Kind = metadata.StringPool.Get("alias")
    t.Target = metadata.StringPool.Get(ut.Name)   // "int"
default:
    t.Kind = metadata.StringPool.Get("other")      // no target at all
```

So the spec layer had nothing to build from and fell through to an opaque object:

| declaration | before | after |
|---|---|---|
| `type IDs []string` | `{type: object}` | array of string |
| `type Point [2]int64` | `{type: object}` | array of integer, `minItems`/`maxItems` 2 |
| `type Lookup map[string]int` | `{type: object}` | object with `additionalProperties: integer` |
| `type Count int` | `{type: integer}` | unchanged — this is why it went unnoticed |

The default branch now records the underlying type with `getTypeName`, the same renderer struct fields already use, so consumers need no second parser. The kind stays `"other"`: these are none of the three above, and calling them aliases would misdescribe defined types.

**Layering:** metadata records what the declaration says (golden rule #4); it does not decide the schema. The spec layer maps that target through the same path as any other type string, registry and config overrides included.

## The half the issue did not predict

`type Nested []Point` records its target **as written** — `"[]Point"`, unqualified. The first version of this fix therefore produced a *second* `Point` component under the bare name, alongside the qualified one, with `Nested` pointing at the duplicate.

`generateAliasSchema` now qualifies its target with the declaring package — the same `qualifyFieldType` call the struct-field path makes. It is inert for a primitive target, which is why the alias path never needed it before. The test asserts the bare component is gone, because I shipped that bug for one build and want it to stay fixed.

## Measured

- **Fixture:** four of five named types stop being empty objects.
- **apispec's own spec:** empty-object components 2 → 1.
- **gitea:** empty-object components 8 → 7, components 73 → 74 (an element type that is now referenced), dangling refs still 0.

Real but modest on existing projects, and larger the more a codebase names its container types (`type Matrix [][]float64`, `type Headers map[string][]string`).

## No golden churn — which is itself the finding

The full suite passes with the metadata goldens unchanged, because those fixtures happen to declare no named container types. That absence is precisely why this survived: nothing in the corpus exercised the branch. The new fixture covers array, slice, map, a named primitive as a control, and a named slice **of** a named type — the case that exposed the qualification half.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved schema generation for named arrays, slices, maps, pointers, and channels.
  - Preserved the structure of named container types instead of representing them as empty objects.
  - Resolved nested named types with package-qualified references, preventing duplicate or dangling schemas.
  - Continued resolving named primitive types inline.

- **Tests**
  - Added regression coverage for named underlying types and schema reference integrity.


ehabterra:
Both resolved in the pushed commit.

**1. Use the memoized `TypeRef` rather than re-parsing the string — valid, and better than I realised.** I checked whether `TypeRefOf` was actually reachable here before agreeing: it takes a **string-pool id**, and `typ.Target` is exactly that, so this call site can use the memoized parse while the struct-field caller (which only has a string) cannot.

Applied as suggested — memoized ref, `Clone()` before mutating, rendered only at the `mapGoTypeToOpenAPISchema` boundary — with one addition. Rather than reimplementing the qualification inline, the decision moved into a shared `qualifyTypeRef(ref, pkg)` that both paths call, so the string path and the ref path cannot diverge. That mattered: a divergence between those two is precisely what produced the duplicate component this PR fixes. Output is byte-identical to the string version, verified on the fixture.

**2. Route/method assertions and unit tests at the changed layers — valid, added.**

- structural: routes and methods for both fixture endpoints, so a fixture that stopped resolving handlers cannot pass the schema assertions by documenting nothing; plus no locally declared type may fall back to the unresolved-external placeholder.
- `processTypeKind`, covered directly over every declaration shape — struct, interface, ident, array, slice, map, pointer, chan, func.
- `generateAliasSchema`, covered for the qualification, a primitive target (must stay untouched), and the three container targets.

Writing the metadata test surfaced two things I have recorded in it rather than quietly accommodated:

- **`Target`'s zero value doubles as "unset"** (yaml `omitempty`) while pool index 0 is a real string, so an unset target reads back as whatever was interned first. The test burns index 0 on a sentinel to keep its assertions honest. It is not reachable today — kind `"other"` is produced in exactly one place, which now always sets `Target` — but it is the kind of thing that becomes a bug the moment a second producer appears.
- **`getTypeName` renders a func type as bare `"func"`.** Recording that is still better than recording nothing (the kind is right and the mapper has something to dispatch on), but it is why a func-typed declaration cannot produce a useful schema. Not in scope here; noted where the next person will find it.

PR #335, opened 2026-08-12
feat: distribute apispecui the same way as apispec

ehabterra:
`apispecui` was installable exactly one way — `go install` — while `apispec` had five. Someone without a Go toolchain could `brew install apispec` and then had no way to get the UI at all, even though it is a self-contained binary that embeds its own assets.

## Every method now covers both tools

| Method | Before | After |
|---|---|---|
| Homebrew | apispec only | `brew install ehabterra/tap/apispecui` |
| Pre-built binary | apispec only (6 platforms) | apispecui too, same 6 platforms + checksums |
| `go install` | worked, undocumented | documented |
| From source | `make install` / `install-local` | `make build-ui` / `install-ui` / `install-ui-local` (+ uninstalls) |
| `scripts/install.sh` | apispec only | both by default, `--tool` selects one |

`scripts/release.sh` now builds a **list** of commands instead of a hardcoded `./cmd/apispec`, so adding a future binary is one entry.

## Two prerequisites this needed first

**`apispecui` had no `--version` flag.** The Homebrew formula's `test do` block asserts on it, and parity wants it regardless. It now prints the same shape as `apispec` — `BuildTime` rather than `BuildDate`, keeping the existing deliberate behaviour of reporting the binary's mtime — and returns before binding a port or walking a directory, which is where the formula's sandboxed test runs it.

**The release workflow patched version constants in `cmd/apispec/main.go` only**, so a `go install` from a tag would have reported `dev` for the UI. It now patches both.

## The tests are the interesting part

They are data-driven over a list of formulae, so a tool added later is covered by default rather than by someone remembering:

- **every `*.rb.tmpl` on disk must be listed, and every listed tool must have a template** — read from the directory, not trusted from the list;
- **`release.yml` must publish the four Homebrew assets each template downloads**, and its tap loop must name the tool.

That second one earned its place immediately: it failed on my first attempt and told me exactly which assets and which rendering step were still missing. A formula whose asset the release never uploads installs as a 404 on a user's machine, and nothing before that point notices.

## No `apispecui.rb` is committed, deliberately

The rendered formula carries digests of published artifacts. Those cannot be invented, and a wrong one is a failed `brew install`. The first release that publishes `apispecui` artifacts creates it; until then the template is the whole source of truth, and the drift test skips with that explanation rather than passing silently.

## Verified end to end

Ran the release path as the workflow will:

- 12 binaries built (2 tools × 6 platforms) with checksums;
- both formulae rendered from the real digests, **zero placeholders left**;
- both binaries report the injected version, so `assert_match version.to_s, shell_output("#{bin}/apispecui --version")` passes.

```
apispecui version: 9.9.9
apispec version: 9.9.9
```

Docs updated: README quick-start now leads with Homebrew for both, the `apispecui` section installs rather than `go build`s, and `docs/INSTALLATION.md` covers both throughout — including removing its "Only the `apispec` CLI is distributed as a binary" statement, which this change makes false.

## One judgement call worth flagging

The install script installs **both** tools by default. Someone running an installer wants a working setup, and the UI is part of one — but it is a behaviour change for existing users of that script, so `--tool apispec` restores the old scope. Happy to flip the default if you would rather it stay CLI-only.

## Summary by CodeRabbit

* **New Features**
  * Added `apispecui` binaries to releases across supported platforms.
  * Added Homebrew installation support for `apispecui`.
  * Added `--version` and `-V` options to display build and license information.
  * Expanded the installer to support installing either or both tools.

* **Documentation**
  * Updated installation guidance with Homebrew, Go, binary, source, and script-based options for both tools.
  * Added build, installation, and uninstallation commands for `apispecui`.
  * Documented available distribution methods and clarified that `apidiag` is source-only.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -101,6 +101,14 @@ jobs:
           sed -i "s|BuildDate = \"[^\"]*\"|BuildDate = \"$BUILD_DATE\"|" cmd/apispec/main.go
           sed -i "s|GoVersion = \"[^\"]*\"|GoVersion = \"$GO_VERSION\"|" cmd/apispec/main.go
 
+          # apispecui declares the same constants (BuildTime rather than
+          # BuildDate — it reports the binary's mtime, which `make release`
+          # cannot know). Patched here for the same reason as apispec: the
+          # ldflags below cover a build, this covers a `go install` from the tag.
+          sed -i "s|Version   = \"[^\"]*\"|Version   = \"$VERSION\"|" cmd/apispecui/main.go
+          sed -i "s|Commit    = \"[^\"]*\"|Commit    = \"$COMMIT\"|" cmd/apispecui/main.go
+          sed -i "s|GoVersion = \"[^\"]*\"|GoVersion = \"$GO_VERSION\"|" cmd/apispecui/main.go
+
       - name: Build for multiple platforms
         # The Makefile declares these with `?=`, so the environment wins and no
         # command line has to be built out of untrusted values at all.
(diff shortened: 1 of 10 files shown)


ehabterra:
Both resolved in the pushed commit.

**1. `--tool` needs a value — valid, and the actual behaviour was worse than "defaults to both".** I traced it rather than patching from the description:

```console
$ ./scripts/install.sh help --tool
================================
  APISpec Installation Script
================================
[INFO] Found Go version: go1.26.0
$ echo $?
1
```

A trailing `--tool` shifts past the end of the argument list, so the `shift` at the bottom of the parse loop fails under `set -e` — the script **aborts silently** after its header. The `both|""` branch never even ran. It now rejects a missing or empty value with the expected values, and `both` is only reachable by asking for it:

```console
$ ./scripts/install.sh help --tool
[ERROR] --tool requires a value (apispec, apispecui or both)
```

A flag as the value (`--tool --help`) was already caught by the unknown-tool branch, and the three valid values still work.

**2. Test the version path — valid, added.** `TestVersionFlag` builds the binary with an injected version and runs it with `--version` and `-V` from an empty directory, which is precisely how the formula invokes it. It asserts the injected version is printed, the process exits cleanly, and it does not start serving — the last one enforced by a deadline, since a bound port keeps the process alive.

I checked the test earns its keep rather than assuming: with the early return removed, it fails instead of passing.

```
--- FAIL: TestVersionFlag/--version
```

It costs a `go build` and ~4s, which I think is justified here — this path has no other coverage, and a break in it surfaces as a failed `brew install` on someone's machine rather than as a red CI run.

Issue #297, opened 2026-08-08
Add --strict: promote quality warnings (unmapped auth middleware, dropped response types) to a non-zero exit


ehabterra:
From a downstream consumer's output-quality review.

Several conditions today are a `log.Printf` line and nothing else. In a 16-second CI log they scroll past, and their effect is invisible in the spec diff for anyone not already looking for it. The clearest example:

```
[security] 1 auth middleware not mapped to a security scheme (add securityMappings to resolve)
```

The message is correct and helpful. The **consequence** is that the operation is emitted with no `security` block — i.e. documented as public. "A route silently lost its documented auth" is exactly the class of change that should fail a build, and today it cannot.

## Ask

A `--strict` flag that turns the existing quality warnings into a non-zero exit. Candidates, all of which are already detected and reported:

- auth middleware not mapped to a security scheme (`mapper.go:156`)
- a response type dropped because the instance cap fired (#296)
- expansion truncated at `--max-nodes` / `--max-nodes-per-route` (already reported via `ExpansionStats`)
- packages skipped because they failed to load or parse (#237)

Design notes worth settling in the issue rather than in review:

- **Off by default.** It changes exit codes, so it is opt-in.
- **Granular if cheap**, e.g. `--strict=security,truncation` — a project may want to fail on lost auth but tolerate a truncated route.
- The consumer's workaround is a `make openapi-assert` target that greps the generated spec and ratchets counts. That works, but it can only see what survives into the YAML: a dropped `security` block and a dropped response type look identical to "this endpoint genuinely has none". apispec knows the difference at generation time and is the only place that can.

Related: #237 (silently dropped packages), #296 (per-operation cap reporting).

Issue #298, opened 2026-08-08
operationId and component names are fully-qualified Go symbols, leaking internal module layout into a spec that is served publicly


ehabterra:
From a downstream consumer's output-quality review.

Every `operationId` is the full import path plus receiver plus method, and every component key is that path with separators mangled to underscores:

```yaml
operationId: github.com/acme/private-api/apps/api/internal/httpapi.estimateHandler.updateLine
components:
  schemas:
    github_com_acme_private-api_apps_api_internal_estimate_LineInput: {…}
```

Two consequences, one of which has a security flavour:

- **Leakage.** For a closed-source service, the module path, the internal package layout, and unexported handler-struct names are reproduced verbatim in a document that is typically **served over HTTP** and embedded in the shipped binary. On the reviewed project this was the single finding that most constrained what could be shared publicly — including the review itself, which had to be pre-sanitised before it could be handed over.
- **Downstream ergonomics.** TypeScript types come out as 80-character identifiers, so every consuming file needs an alias.

Fully-qualified names are the right **default** — they are collision-free and reproducible, and #216 fixed the one case where they were not (absolute paths in closure operationIds). This is about making them a choice.

## Ask

A config-level naming strategy, `full` remaining the default:

```yaml
naming:
  operationId: method-path | receiver-method | full   # e.g. getEstimateLine
  schemaNames: short | full                           # e.g. LineInput
```

The constraint that makes this non-trivial, and that should be settled in the design rather than in review: **short names collide** — two packages with a `Components` type are ordinary. Disambiguation has to be deterministic and applied only on collision (golden rule #1), e.g. shortest unique suffix of the package path, extended one segment at a time, sorted so the choice cannot depend on map order. Silently letting one `Components` win would be worse than the long names.

Worth checking against `internal/spec/mapper.go` component-name generation and the `operationId` construction, both of which are already required to be deterministic.

ehabterra:
A source of good `operationId`s that already exists in the analysed code and is currently ignored: **the route's own name**. Several supported routers let a project name a route at registration, and that name is written by a human, is unique per route by construction, and leaks nothing about the module layout:

| Router | Declaration |
|---|---|
| gorilla/mux | `r.HandleFunc("/products/{pk}", h).Name("product-detail")` |
| echo | `e.GET("/products/:pk", h).Name = "product-detail"` (field assignment on the returned `*echo.Route`) |
| fiber | `app.Get("/products/:pk", h).Name("product-detail")` |
[...]

PR #339, opened 2026-08-12
feat(spec): repair and report references the document cannot satisfy

### LINKED ISSUE #327: Nothing validates that every $ref resolves before the spec is written, so dangling references ship silently
## What

Three separate defects (#325, #326) and one already fixed (#259) share a symptom: the tool writes a spec containing `$ref`s to components it never defined, and says nothing. The user finds out when Swagger UI refuses to render:

```
Could not resolve reference: Invalid object key
"google_golang_org_protobuf_types_known_timestamppb_Timestamp" ...
key not found in object
```

Each of those is worth fixing at its source. But the pattern — four causes, three of them found only because someone loaded the output into a viewer — says the missing thing is a check that the finished document is internally consistent.

On gitea at default limits the finished spec contains **148 distinct `$ref`s, 165 components, 4 of the refs dangling**. Detecting that is fifteen lines of set arithmetic over the object apispec has just built in memory.

## Why it is worth doing separately

- **It is the last line of defence for a whole bug class.** Every future path to a `$ref` — a new framework config, a new container shape, a new external library — is covered by one check rather than by remembering golden rule #4 at each of the 10+ `addRefSchemaForType` call sites.
- **The generator tests already do exactly this**, per-fixture: `noDanglingRefs` in `generator/testdata_*_test.go`. The invariant is already written down and enforced in CI for fixtures; it just is not enforced for a user's actual project, which is where the unregistered dependency types live.
- **It is cheap and total.** Walk the assembled `OpenAPISpec`, collect every `$ref` target, subtract `components.schemas`, report the difference. No re-parsing, no extra pass over the call graph.

## Suggested behaviour

At the end of generation, after components are assembled:

1. Collect every `$ref` in paths and components; subtract the defined component names.
2. If any remain, **repair rather than emit**: replace each dangling reference with the honest placeholder that already exists (`unresolvedExternalPlaceholder`) so the document stays loadable, and report what was substituted:
   ```
   warning: 4 references had no component and were inlined as unresolved:
     github.com/golang-jwt/jwt/v5.RegisteredClaims (2 sites)
     xorm.io/xorm/convert.Conversion (1 site)
     ...
   ```
3. Name the Go type, not the mangled component key — `github.com/golang-jwt/jwt/v5.RegisteredClaims` tells the user which dependency to register in `externalTypes`; `github_com_golang-jwt_jwt_v5_RegisteredClaims` does not.

A repair keeps a usable spec in the user's hands today while the underlying causes are fixed, and the warning is what turns "Swagger UI is broken" into "these four types need registering".

The exit-code question belongs to #297 (`--strict`): the warning should be one of the conditions that flag promotes to a non-zero exit, not its own flag.

## Note

This is deliberately a safety net, not a substitute for #325 and #326. A repaired `{}` is still a worse spec than a resolved type — the point is that it is *valid*, and that the user is told.

ehabterra:
Closes #327. The last item I would want in v0.5.7.

## Why now

This cycle fixed **four separate causes** of dangling `$ref`s — #325 (dependency types with no declaration), #326 (fixed-size arrays, untyped constants), #329 (mis-qualified types), #333 (named container types). Every one was found the same way: by loading the output into a viewer.

That is not a check anyone runs against their own project, and the symptom is brutal — Swagger UI refuses the whole document over a single unresolvable reference. The generator tests assert this per fixture (`noDanglingRefs`); nothing asserted it for a user's actual code, which is exactly where the unregistered dependency types live.

## What it does

Generation now ends by making the document internally consistent. A missing target is **repaired, not dropped** — the reference stays and the honest placeholder is registered under it, because removing the reference would silently change an operation's shape.

The report names the **Go type**:

```
2 type(s) had no schema and were inlined as unresolved across 4 reference(s):
github.com/golang-jwt/jwt/v5.RegisteredClaims (3 sites), xorm.io/xorm/convert.Conversion
```

`github.com/golang-jwt/jwt/v5.RegisteredClaims` tells a user which dependency to register under `externalTypes`. `github_com_golang-jwt_jwt_v5_RegisteredClaims` does not. The mangling is lossy (`/`, `.` and the type separator all become `_`), so the Go type is recovered by mangling the names already in `usedTypes` and keeping the map — which is why `generateComponentSchemas` now returns it.

Machine-readable too, via `SecurityDiagnostics.UnresolvedRefs`, `Engine.GetUnresolvedRefs` and `Generator.UnresolvedRefs`, so the UI and a CI gate need not grep stderr. **The exit-code half stays with #297 (`--strict`)** — this is the condition that flag would promote, not a second flag.

## The walker is the part that has to be right

A location it misses is a reference the repair leaves dangling — the same silent failure, one level up. So it walks the typed document rather than the schemas map: path- and operation-level parameters, request bodies, responses, response headers, and `components`' schemas, parameters, requestBodies, responses and headers.

**18 tests, one per location**, each putting a reference in exactly one place and asserting the walker finds precisely it. Plus foreign/non-schema refs ignored, a self-referential schema terminating, sorted output (golden rule #1 — the repair inserts components, so map order would reach the output), and existing components never overwritten.

## Verified against a real regression

I reintroduced one of the bugs this release fixed — disabled the untyped-constant normalisation — and re-ran:

```
[engine] 1 type(s) had no schema and were inlined as unresolved across 1 reference(s): untyped bool
dangling after repair: []
```

It catches it, names it in Go terms, and the output still loads. On a healthy tree (this repo, and the fixtures) it reports nothing and changes nothing.

## Note on the end-to-end test

The fixture-level test asserts the **invariant**, not a non-empty report: since #325 the generation pass already registers a placeholder for a type it cannot resolve, so nothing reaches the final repair. An empty report is the healthy state. I had it `t.Skip` on empty at first, which is misleading — a skipping test reads as coverage it is not providing. The repair itself is exercised directly by the unit tests.

Coverage is back at 96.0% against the floor of 96 (the new reporting path needed its own tests to get there).

## Summary by CodeRabbit

* **New Features**
  * Added reporting for unresolved schema references during document generation.
  * Generated documents now preserve unresolved references with placeholder schemas, preventing dangling references.
  * Reports include the affected component, Go type when available, and the number of reference sites.
  * Added public access to unresolved-reference details through the generator API.
  * Unresolved-reference diagnostics are now included in extraction results for easier troubleshooting.

### DIFF
--- a/generator/generator.go
+++ b/generator/generator.go
@@ -78,3 +78,11 @@ func (g *Generator) PathParamMismatches() []intspec.PathParamMismatch {
 	}
 	return g.engine.GetPathParamMismatches()
 }
+
+// UnresolvedRefs returns the $refs the last generation could not satisfy, after
+// they were repaired with a placeholder. Non-empty means the document loads but
+// some type resolved to nothing useful — the actionable fix is usually
+// registering it under externalTypes (issue #327).
+func (g *Generator) UnresolvedRefs() []intspec.UnresolvedRef {
+	return g.engine.GetUnresolvedRefs()
+}
--- a/generator/generator.go
+++ b/generator/generator.go
@@ -84,5 +84,8 @@ func (g *Generator) PathParamMismatches() []intspec.PathParamMismatch {
 // some type resolved to nothing useful — the actionable fix is usually
 // registering it under externalTypes (issue #327).
 func (g *Generator) UnresolvedRefs() []intspec.UnresolvedRef {
+	if g.engine == nil {
+		return nil
+	}
 	return g.engine.GetUnresolvedRefs()
 }
--- a/generator/testdata_external_type_refs_test.go
+++ b/generator/testdata_external_type_refs_test.go
@@ -15,6 +15,7 @@
 package generator
 
 import (
+	"path/filepath"
 	"strings"
 	"testing"
 
@@ -94,4 +95,45 @@ func schemaPropNames(s *spec.Schema) []string {
 		out = append(out, k)
 	}
 	return out
+}
+
+// TestTestdata_UnresolvedRefsAreReported covers the end of the chain for issue
+// #327: a type with no schema is repaired so the document still loads, AND the
+// fact is reported so a user knows their spec has a placeholder in it rather
+// than a type.
+//
+// external_type_refs is the right fixture: its fields are standard-library
+// types, which are outside the analysed module exactly like a dependency, so
+// they genuinely cannot resolve.
+func TestTestdata_UnresolvedRefsAreReported(t *testing.T) {
+	dir := filepath.Join("..", "testdata", "external_type_refs")
+	g := NewGenerator(spec.DefaultHTTPConfig())
+	out, err := g.GenerateFromDirectory(dir)
+	if err != nil {
+		t.Fatalf("GenerateFromDirectory: %v", err)
+	}
+
+	// The document must load regardless — that is what the repair is for.
+	noDanglingRefs(t, out)
+
+	// An empty report is the healthy state and the expected one here: since
+	// #325 the generation pass registers a placeholder for a type it cannot
+	// resolve, so nothing reaches the final repair. That is why this asserts
+	// the invariant rather than a non-empty report — the repair is a net for
+	// causes not yet known, and unit tests in internal/spec exercise it
+	// directly. What is worth checking end to end is that the accessor is
+	// wired and self-consistent.
+	for _, r := range g.UnresolvedRefs() {
+		if r.Component == "" {
+			t.Error("an unresolved ref reports no component name")
+		}
+		if r.Sites < 1 {
+			t.Errorf("%s reports %d sites, want at least 1", r.Component, r.Sites)
+		}
+		// Every component it names must now exist — the report describes what
+		// was repaired, not what is still broken.
+		if out.Components == nil || out.Components.Schemas[r.Component] == nil {
+			t.Errorf("%s was reported unresolved but no placeholder was registered for it", r.Component)
+		}
+	}
 }
--- a/generator/testdata_external_type_refs_test.go
+++ b/generator/testdata_external_type_refs_test.go
@@ -116,22 +116,34 @@ func TestTestdata_UnresolvedRefsAreReported(t *testing.T) {
 	// The document must load regardless — that is what the repair is for.
 	noDanglingRefs(t, out)
 
-	// An empty report is the healthy state and the expected one here: since
-	// #325 the generation pass registers a placeholder for a type it cannot
-	// resolve, so nothing reaches the final repair. That is why this asserts
-	// the invariant rather than a non-empty report — the repair is a net for
-	// causes not yet known, and unit tests in internal/spec exercise it
-	// directly. What is worth checking end to end is that the accessor is
-	// wired and self-consistent.
-	for _, r := range g.UnresolvedRefs() {
+	// This fixture must report NOTHING, and that is a real assertion rather
+	// than a vacuous loop: since #325 the generation pass registers a
+	// placeholder for a type it cannot resolve, so nothing should reach the
+	// final repair. A non-empty report here means type resolution regressed
+	// far enough that the safety net had to catch it — which is exactly the
+	// signal worth failing on.
+	//
+	// The repair itself is exercised directly by unit tests in internal/spec
+	// (one per document location, plus its repair and reporting behaviour).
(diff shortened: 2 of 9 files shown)


ehabterra:
Three fixed, one answered with evidence.

**1. `Generator.UnresolvedRefs` should nil-guard the engine — valid, fixed.** `PathParamMismatches` directly above it does exactly that; mine was inconsistent.

**2. The cycle guard should be a recursion stack, and shared schemas counted per mount — valid, fixed.** A `*Schema` mounted in two operations was counted once. Detection was unaffected (its references were still found), but `Sites` undercounted — and `Sites` is the number that tells a user how much of their spec is affected. Now only a schema *currently being walked* is skipped, so a self-referential type still terminates while a shared one counts per mount. Test added, as suggested.

I measured the cost that trades against, since re-walking shared subtrees is the risk: on gitea the mapping stage is **27.3s vs 29.0s** — within run-to-run noise. The walk is over the assembled document, which is tiny next to the expansion that produced it.

**3. Don't guess between colliding component names — valid, fixed.** "Prefer the shortest" was a heuristic, and naming the wrong dependency sends someone to register a type that was never the problem. Ambiguous names now report **no** Go type and the message falls back to the component name (golden rule #7).

I did not take the larger half of that suggestion — threading a component→type association through generation. The lossy mangling is applied with the *same* replacer the naming uses, so the two cannot disagree; and this is a diagnostic string, not a decision the spec depends on. Handling ambiguity honestly gets the correctness without a generation-wide refactor. Happy to revisit if you want the association tracked properly.

**4. Make the end-to-end test produce a real unresolved reference — I'd rather not, and here is why.**

I scanned **every fixture in `testdata/` that builds**, generating each and checking `UnresolvedRefs()`:

```
(no hits)
```

None produces one — because since #325 the generation pass already registers a placeholder for a type it cannot resolve, so nothing reaches the final repair. A fixture that produced one would have to encode a bug that does not currently exist, and it would stop reproducing the moment that bug was fixed elsewhere.

So the e2e test asserts the invariant (document loads, and whatever was reported was registered), and the repair itself is covered directly by unit tests — 18 of them, one per document location. I did remove the `t.Skip` you were reacting to; a skipping test reads as coverage it is not providing, which was a fair thing to flag.

---

**One thing this investigation turned up, unrelated to this PR: [#340](https://github.com/ehabterra/apispec/issues/340).**

While A/B-ing the walker change I found gitea produces **two different specs from the same command** — `478c93378a08` (572486 bytes) and `b7de309fb7d9` (572281 bytes), roughly 2 runs in 15. Both binaries produce both, so it is not from this branch. The rarer output is *missing responses*.

I chased it to be sure this PR wasn't the cause: 5 consecutive runs of each binary were identical and equal to each other, and the divergent outputs came from both. There is no time-based limit anywhere in the analysis path, so it points at a map range in the truncation path reaching the output — and `TestGenerateDeterministic` cannot see it because no fixture is large enough to truncate. Filed with the reproduction rather than folded in here.

ehabterra:
Both remaining threads are now addressed and resolved.

**`generator/testdata_external_type_refs_test.go` — "make the public-access test non-vacuous".** Fair hit, and I had only half-answered it. My reply explained why no fixture can *produce* an unresolved reference, but left the loop itself asserting nothing when the report is empty.

It now asserts the report **is** empty, which is a real claim about this fixture rather than a no-op: since #325 the generation pass registers a placeholder for anything it cannot resolve, so nothing should reach the final repair. A non-empty report means type resolution regressed far enough that the safety net had to catch it.

Verified non-vacuous by disabling the placeholder registration in `mapGoTypeToOpenAPISchema` — the test then fails and names them:

```
5 reference(s) needed repairing in a fixture whose types all resolve — type resolution regressed:
[net_url_Userinfo net_url_Values regexp_Regexp text_template_Template time_Location]
```

The per-entry checks are kept, so a future non-empty report is inspected rather than only counted.

**`internal/spec/refcheck.go` — "preserve an unambiguous component-to-type mapping".** Already fixed in 4c0c5e0, which landed before this thread was re-read: the shortest-name selection is gone. Two different types mangling to one component name now mark it ambiguous and the entry is dropped, so `GoType` comes back empty and the report falls back to the component name. That removes both halves of the concern — the selection cannot vary with map order because there is no selection, and it cannot name the wrong type because it names none.

I did not take the other half of that suggestion (recording the association during component generation). The mangling uses the *same* replacer the naming does, so the two cannot disagree; ambiguity is now handled honestly; and this is a diagnostic string rather than something the spec depends on. Worth doing if component naming is reworked under #298 — not worth a generation-wide change for a warning message.

CI green, coverage back at 96.0%.

PR #344, opened 2026-08-14
perf: index function lookups; shrink LazyNode

ehabterra:
## What

Two changes to the `spec mapped` stage, found by profiling a 163-route service
(103 packages) that took 13s to generate.

1. **`findFunctionByName` is now indexed.** It resolved a bare name by scanning
   the named package's files and then, on a miss, sorting every package key and
   scanning every file of every package — per call. It is not a cold path: the
   response-destination resolver reaches it through `assignmentsAt` →
   `callerAssignmentMap` once per candidate per path, and the lazy tree
   materialises one node per path. This is `TypeInPackage`'s problem (#322) on
   the function side, so it gets `TypeInPackage`'s answer —
   `Metadata.FunctionInPackage` and `FunctionAnywhere`, behind the same
   `pkgShape` invalidation.

2. **`LazyNode` shrinks 104 → 80 bytes**: `ArgumentType int` → `uint8`,
   `routeScope int` → `int32` (both now sharing one word with the flags), and
   the `GetTypeParamMap` memo moved to a side table on the tree since only
   generic call sites read it.

## Measured

On the 163-route service, generated spec **byte-identical** before and after
(and byte-identical on this repo's own spec):

|                | before  | after   |
|----------------|---------|---------|
| `spec mapped`  | 10.02s  | 6.32s   |
| wall clock     | 12.45s  | 8.03s   |
| peak RSS       | 1318MB  | 1148MB  |

The wall-clock win is all from (1); (2) is the RSS win and moves time not at all.

## Where the rest of the time goes

Recording this so it isn't re-derived. The stage is essentially *nothing but*
node materialization: `extractRouteChildrenScoped` is 85% cumulative and all of
it is `GetChildren` plus recursion — the actual extraction work (pattern
matching, request/response/param) is ~250ms. 12,882 distinct callee keys expand
into 7,147,505 materialized nodes, ~540 copies per key, and the top keys are
leaf utility calls (`fmt.Errorf` with a literal, an ObjectID helper,
`reflect.Value.*` in a struct walker) re-expanded once per reaching path.

Wall clock is linear in that count: at `--max-instances-per-key` 100/50/25 the
stage is 6.7s/4.3s/2.8s **for a byte-identical spec on this project**. That is
#224 — the cap is applied to the wrong unit — and this PR deliberately does not
touch the default, which #342 raised for a different service that genuinely
needs it.

Two dead ends, so nobody re-walks them:

- **It is not GC.** `GOGC=300` and `GOGC=800` are both *slower* and cost 400MB
  of RSS; `GOGC=off` is slower still (8.3s vs 6.3s).
- **It is not the field stores.** `GetChildren`'s flat CPU sits on plain
  assignments like `child.routeScope = childScope`, but that is first-touch on
  freshly committed slab pages — shrinking the struct 23% moved memory and not
  time, which is the evidence.

## Test plan

- [x] `go test ./...` and `go test -race ./internal/... ./generator/...`
- [x] `make lint` (0 issues), `go vet`, `gofmt`
- [x] Generated spec byte-identical on the 163-route service and on this repo

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved function resolution within packages and across the application.
  * Ensured lookups remain accurate and deterministic as files and package contents change.
  * Improved handling of missing function names without disrupting normal operation.
* **Performance**
  * Improved reuse of generic type information across syntax trees.
  * Reduced memory usage for type and argument tracking.
  * Added caching to speed up repeated function lookups.

### DIFF
--- a/internal/metadata/lookup_cache_test.go
+++ b/internal/metadata/lookup_cache_test.go
@@ -15,8 +15,13 @@
 package metadata
 
 import (
+	"go/ast"
+	"go/parser"
+	"go/token"
 	"go/types"
+	"os"
 	"reflect"
+	"strings"
 	"sync"
 	"testing"
 )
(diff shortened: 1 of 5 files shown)


ehabterra:
Thanks — went through all three. Summary: one was a real gap in what the code *said*, one is unreachable but the suggested form is better anyway, and none needed the fix as proposed. Pushed c9988cf.

### `FunctionAnywhere` has no invalidation (types.go 474-568, 531-566)

The asymmetry with `FunctionInPackage`'s `pkgShape` guard is real, and it's deliberate — but nothing said so or enforced it, which is the actual defect. Adding the guard isn't an option: validating a whole-program fingerprint means counting the functions of every file of every package, and the fallback is consulted often enough that doing it per lookup **costs the entire win** — on a 163-route service the mapping stage goes 6.3s → ~9s, back where it started. I also tried composing `FunctionInPackage` over every package (which would inherit the per-package guards); that's the same cost by another route, since `shapeOf` walks the files — measured 8.4s.

What makes build-once safe here is a layer boundary, not luck:

- functions and packages are only ever added inside `GenerateMetadata` (`metadata.go:371-374`, `metadata.go:1192`, `io.go:188`; `types.go:1569` rewrites an existing key with the same pointer, so it's not a shape change);
- every caller of this lookup is in `internal/spec`, which doesn't run until assembly has returned — the spec layer only ever *reads* `Files`/`Functions`.

`FunctionInPackage` needs its guard precisely because it *lacks* that property: metadata-layer code calls it mid-assembly.

So the invariant is now written at the function and enforced by `TestFunctionAnywhereIsSpecLayerOnly`, which parses the metadata package and fails on any call to `FunctionAnywhere` from inside it — verified to actually fail when one is added, and it names the two ways out (move the caller, or pay for the guard). Plus `TestFunctionAnywhereIsDeterministic` (first package in sorted order wins for a duplicate bare name), `TestFunctionInPackageMatchesFileScan` (pins the index against the scan it replaced), and `TestFunctionInPackageInvalidatesOnGrowth` — that last one covers your implicit point that a *file* tally is insufficient, since assembly adds functions to files that already exist.

### `int32` route-scope overflow (lazytree.go 503-517)

Not reachable: a scope id is one route registration, and reaching `math.MaxInt32` of them needs 2.1 billion routes while the walk that discovers them is bounded by `MaxNodesPerTree` (50,000 by default). So I haven't added a rejection path at the boundary — it would be dead code guarding an impossible state.

The loop change you suggested is worth taking on its own merits though, and is applied:

```go
-for int32(len(t.routeScopeCut)) <= scope {
+for len(t.routeScopeCut) <= int(scope) {
```

Comparing as `int` is simpler and drops a conversion from a loop condition. Same for the `routeScopeKeys` bounds check.

Spec output stays byte-identical throughout; suite, `-race` and lint clean.

ehabterra:
Applied in `ede9041` — good catch, and it was a real hole rather than a style point.

The invariant is about the caller's *phase*, not its package. Assembly is driven from `internal/metadata` today, so scanning that package caught the realistic case, but a future analysis pass called from `GenerateMetadata` could live anywhere and would have slipped through. The check now walks every non-test Go file in the repo and allows the call only under `internal/spec`, which is the enforceable form of what the doc comment actually claims. Fixture projects (`testdata/`, `test_cgo_*`) are skipped — separate modules that never import this one.

Verified rather than assumed: adding a caller in `internal/engine` passes the old package-local version and fails the new one with its position.

Also worth recording from the earlier round, since it vindicates the tests you pushed back on — CI caught `TestFunctionInPackageMatchesFileScan` failing, and it was right. `FunctionInPackage` filled its index by ranging `pkg.Files`, so a bare name declared in two files of one package resolved by map order. That is the determinism bug golden rule #1 exists to stop, and it is exactly what `TypeInPackage` avoids by walking `SortedFileNames`. Fixed in `6dfdc8d`; both lookups now walk sorted file names with the earlier file winning. The test that caught it was itself too weak — it only asserted two consecutive lookups agree, which a map-order index satisfies whenever the map iterates the same way twice, which is why it passed locally. It now names the expected winner and rebuilds 32 times (fails 3/3 against the buggy version).

Spec output stays byte-identical on both projects throughout.

PR #348, opened 2026-08-14
perf: skip subtrees no matcher can accept (#318)

### LINKED ISSUE #318: Tracker tree materializes 399 nodes per distinct callee key (6.1M nodes, 41% of all allocation)
## What

The lazy tree materializes **399 `LazyNode`s for every distinct callee key it expands**. On this repo, generating its own spec:

```
nodesBuilt        =    15,334     (distinct callee keys — what MaxNodesPerTree bounds)
nodesMaterialized = 6,113,381     (LazyNodes actually created)
ratio             =       398.7
```

That is the single largest cost in the tool. From the CPU and allocation profiles of the same run:

| | share |
|---|---|
| `LazyNode.GetChildren` | **44.9% of CPU** (cumulative) |
| `LazyTree.newNode` | **27.6% of all allocation** (0.57GB of 2.07GB) |
| `GetChildren` (flat alloc — the children slices) | 13.2% (0.27GB) |
| `runtime.madvise` under `mheap.allocSpan` | 10.7% of CPU |

`GetChildren`'s *flat* CPU time sits almost entirely on plain field stores into the freshly allocated node:

```
     400ms   child.tree = n.tree
     600ms   child.key = spec.key
     310ms   child.parent = n
     250ms   child.edge = spec.edge
    1.39s    child.routeScope = childScope
```

Those are not expensive operations — that is the cost of touching 0.57GB of cold, just-allocated memory. The work is volume, not logic.

## Root cause

Node identity is per path, by design (`lazytree.go:902`, and the header comment at the top of the file):

> node objects stay per-path so every node has exactly one true parent — per-route isolation needs no shared-node discipline

The call graph is a DAG, so a callee reached along N distinct paths gets N nodes. `planFor` already memoizes the *plan* (the structural children) per content identity, so the expensive graph work is shared — but every path still instantiates its own copy of the resulting nodes.

The comment at `lazytree.go:1091-1094` records why reuse was rejected:

> Reusing an existing instance instead would make the tree cyclic (consumers of a memoized subtree could reach themselves), so the bound is a skip

So the blow-up is bounded rather than avoided: `MaxInstancesPerKey` skips copies past the cap. On this repo that cap **refuses 9,767,185 copies** — i.e. the unbounded number would be ~16M nodes — and 6 of 36 routes still hit the 1,000,000-node per-route limit.

Related and already known: #247 (closed) established that `MaxNodesPerTree` counts keys rather than copies, and #224 covers the *correctness* consequence of the instance cap firing (response bodies going missing). This issue is the cost itself, which neither of those addresses.

## Why it matters

The materialization is what drives the other three profile entries: the 0.57GB is retained for the life of the walk (peak RSS 1.35GB), the `madvise` time is the runtime committing pages for it, and every per-node walk downstream (see the `onPath` and `chainInterner` issues) is multiplied by it.

Also worth noting: GOGC tuning does **not** help. At GOGC 100/300/600 the run time was flat (13.7s / 13.7s / 14.1s) while peak RSS rose 1.35GB → 1.80GB. The heap here is live, not churning, so there is nothing to trade.

## Directions, none of them free

1. **Shrink the node.** `LazyNode` currently carries `tree`, `key` (a string header), `parent`, `edge`, `arg`, `typeParams` (a map, for a cache), `children`, `routeScope`, `argType` and two bools. Interning `key` to an int handle and dropping the per-node `typeParams` map into a side table would cut the struct materially. This does not reduce the count, but 6.1M copies makes every saved word worth ~6MB.
2. **Share subtrees below the point where per-path identity stops mattering.** The cyclicity objection applies to a node that can reach itself; a memoized subtree that provably contains no back edge to its parent chain could be shared. That needs the reachability question answered cheaply, which is what the resolved call graph (#245/#253) could provide.
3. **Make the per-path unfolding lazier still** — materialize a node only when a consumer asks for something other than traversal. Most of the 6.1M nodes are visited once by the extraction walk and never looked at again.

Measured on this repo (36 routes, 16 packages) at default limits, after #225/PR #317. A larger service will be worse: the ratio is a property of graph density, not size.

ehabterra:
Closes #318.

## The problem

The lazy tree materializes one node per PATH, so a callee reached along many paths is rebuilt once per path. On a 163-route service, 12,882 distinct callee keys unfold into 7,147,505 nodes. Instrumenting the extraction walk to report, per node, whether it or anything below it matched a pattern:

```
nodes visited                              = 7,103,742
nodes whose ENTIRE subtree matched nothing = 6,930,424   (97.6%)
nodes that ever matched                    =   104,156    (1.5%)
```

The walk builds 7.1M nodes to consult 104K.

## Why skipping them is sound, not a heuristic

The property this rests on is already documented in the code and relied on for the per-edge matcher memos (`extractor.go`):

> The MatchNode implementations of the response/request/param matcher families are **pure functions of the call edge**... Extraction itself stays per-node: it depends on the node's ancestry.

Matching is edge-pure; only *extraction* needs ancestry. So "does this subtree contain anything a matcher would accept" is a property of **content identity, not path** — precisely what a per-path unfolding cannot exploit on its own, and why the same barren subtree is rebuilt hundreds of times.

The answer is computed once over the **plan graph** (content identities, thousands) instead of the unfolded tree (millions), and `GetChildren` skips a child that cannot reach a match. A node that *can* reach one is still built, so provenance resolution walking up from a matching node is untouched — nothing it can reach is ever pruned.

Every matcher family is represented in the predicate — route, mount, security, request, response, param. A family left out would drop subtrees it would have matched, so `edgeMatchesAnyFamily` is the place to look when adding one.

**The fixpoint runs backwards from the matches, not forwards from the roots.** That is what makes cycles a non-issue: a forward DFS with memoization caches a premature `false` for a node whose only route to a match runs through a back edge still in progress (`x -> y -> x` with `x -> match` wrongly answers false for `y`), and would need SCC condensation or repeated passes to be correct. `TestComputeReachHandlesCycleToAMatch` is exactly that graph.

Pruning is enabled by the extractor installing the predicate before it walks, since the matcher families are its own. A nil predicate is the off switch **and the default** — a consumer that wants the full unfolding simply never installs one. (The diagram server builds its own eager tree and never runs the extractor, so diagrams are unaffected; `--diagram` renders from metadata, not the tree.)

## Measured

Mapping stage, minima of three interleaved pairs:

| | before | after |
|---|---|---|
| 163-route service | 6.06s | **2.67s** |
| ↳ nodes materialized | 7,147,505 | **405,490** (−94.3%) |
| ↳ peak RSS | 1145 MB | **522 MB** (−54%) |
| this repo | 6.58s | **0.58s** |

## Output

**Byte-identical** on the 163-route service, and deterministic across three runs.

**On this repo it changes — and only by gaining.** The baseline exhausts the per-route budget on 6 of 36 routes and drops 8,706,027 call copies to the instance cap, so responses were missing; the diff adds `200`/`500` responses with proper `$ref` schemas. Confirmed this is budget and not semantics:

- the pruned output is **converged** — raising both caps 20x changes nothing;
- the same raise on the baseline **does not terminate in ten minutes**.

That is #224's symptom (bodies silently missing at scale) easing as a side effect. The cap is much less load-bearing now: 509,218 copies dropped where it dropped 3,624,817, and the per-route truncation warning disappears entirely on the 163-route service.

## Test plan

- [x] `go test -race ./...`, `make lint`, `go vet`, `gofmt` — all clean
- [x] Metadata goldens unchanged; determinism tests pass
- [x] `testdata/barren_subtree` — two routes wrapped around a dense, mutually recursive utility layer that matches nothing, asserting the paths, request body, status-coded response and array response all survive. **Its output is identical with and without pruning**, so it is a change-detector rather than a test of the optimisation's own arithmetic.
- [x] `internal/spec/prune_test.go` — the fixpoint over named graphs: path-to-a-match, cycle-to-a-match, matchless cycle, matching root, diamond join; plus `TestSpecIdentityMatchesPlanKey`, which pins that the identity a reachability answer is recorded against is the key `planFor` looks the plan up by (if those drift, the prune silently answers about a different node).

## Summary by CodeRabbit

* **Performance**
  * Improved specification processing by pruning branches that cannot produce relevant routes or schemas.
  * Preserved route methods, request-body schemas, status-coded responses, and array responses during pruning.
  * Added handling for cyclic and branching specification structures.

### DIFF
--- a/generator/testdata_barren_subtree_test.go
+++ b/generator/testdata_barren_subtree_test.go
@@ -1,0 +1,94 @@
+// Copyright 2026 Ehab Terra
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package generator
+
+import (
+	"sort"
+	"strings"
+	"testing"
+)
+
+// TestTestdata_BarrenSubtree guards the barren-subtree prune (issue #318): the
+// expansion skips subtrees no matcher family can accept, which is 97.6% of the
+// nodes on a real service.
+//
+// The fixture wraps two ordinary routes around a dense, mutually recursive
+// utility layer that matches nothing. What must survive the prune is everything
+// the routes actually declare — the paths, the request body, the status-coded
+// response and the array response — because each of those hangs off a node the
+// prune has to prove reachable. A prune that is too aggressive does not fail
+// loudly; it drops one of these quietly, which is what this asserts against.
+func TestTestdata_BarrenSubtree(t *testing.T) {
+	out := loadTestdata(t, "barren_subtree", nil)
+
+	item, ok := out.Paths["/orders"]
+	if !ok {
+		t.Fatalf("/orders missing; have %v", mapPathKeys(out.Paths))
+	}
+	if item.Post == nil {
+		t.Fatal("POST /orders missing — the route is behind the barren helper layer")
+	}
+	if item.Get == nil {
+		t.Fatal("GET /orders missing — it shares the barren helpers with POST")
+	}
+
+	// Request body: reached through the handler, which also calls into the
+	// barren layer.
+	if item.Post.RequestBody == nil {
+		t.Fatal("POST /orders: request body dropped")
+	}
+	body, ok := item.Post.RequestBody.Content["application/json"]
+	if !ok || body.Schema == nil {
+		t.Fatalf("POST /orders: want an application/json body schema, got %+v", item.Post.RequestBody.Content)
+	}
+	if !strings.Contains(body.Schema.Ref, "CreateOrderRequest") {
+		t.Errorf("POST /orders: want a $ref to CreateOrderRequest, got ref=%q type=%q", body.Schema.Ref, body.Schema.Type)
+	}
+
+	// Response: the status comes from WriteHeader and the schema from the
+	// Encode argument, both below the same handler.
+	created, ok := item.Post.Responses["201"]
+	if !ok {
+		codes := make([]string, 0, len(item.Post.Responses))
+		for code := range item.Post.Responses {
+			codes = append(codes, code)
+		}
+		sort.Strings(codes)
+		t.Fatalf("POST /orders: want a 201 response, have %v", codes)
+	}
+	resp, ok := created.Content["application/json"]
+	if !ok || resp.Schema == nil {
+		t.Fatalf("POST /orders 201: want an application/json schema, got %+v", created.Content)
+	}
+	if !strings.Contains(resp.Schema.Ref, "Order") {
+		t.Errorf("POST /orders 201: want a $ref to Order, got ref=%q type=%q", resp.Schema.Ref, resp.Schema.Type)
+	}
+
+	// The array response on the sibling route, which reaches the same helpers
+	// along different paths — the diamond the unfolding duplicates.
+	var listSchemaOK bool
+	for _, r := range item.Get.Responses {
+		c, ok := r.Content["application/json"]
+		if !ok || c.Schema == nil {
+			continue
+		}
+		if c.Schema.Type == "array" && c.Schema.Items != nil && strings.Contains(c.Schema.Items.Ref, "Order") {
+			listSchemaOK = true
+		}
+	}
+	if !listSchemaOK {
+		t.Error("GET /orders: want an array-of-Order response; the sibling route's schema was dropped")
+	}
+}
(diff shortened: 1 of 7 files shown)


ehabterra:
Both applied in `54bfb01`, one of them differently than proposed.

### `reachIdentities` not reset in `SetEdgeMatcher`

True of the code as written, but the field is **written once and never read** — and my own comment on it claimed it was "reported", which it was not; I never wired it to anything. Resetting a field nobody reads is maintaining dead code, so it is deleted instead. If the index size is wanted as a diagnostic it should arrive with a consumer, next to the `NodesBuilt` / `NodesMaterialized` it would sit beside in `ExpansionStats`.

### Test uses nil edges for both children

Good catch, and a real weakness rather than a style point: with both `childSpec.edge` and `childSpec.argEdge` nil, the test would have passed even if `specIdentity` read the parent edge for an argument child — the exact contract it exists to pin. That matters here because an argument child carries its OWN edge, not the parent edge it holds for context, and getting it wrong would make the prune test the wrong edge for a match.

It now uses three distinct non-nil edges and asserts, for both child shapes, that the identity **and** `specEdge` are the ones `GetChildren` would give the node. Verified rather than assumed — making `specIdentity` read `spec.edge` for argument children:

```
--- FAIL: TestSpecIdentityMatchesPlanKey
    argument child: specIdentity = {edge:0x...480 ...}, planFor would use {edge:0x...300 ...}
```

The nil-edge version passed that same mutation.

Spec stays byte-identical on the 163-route service; suite, `-race` and lint clean.

PR #349, opened 2026-08-14
fix(scripts): compare enum values as a set, not by position

ehabterra:
An enum is a set: membership is its whole meaning and position carries none. `compare-spec.sh` flattened it positionally, so **inserting** a value reported every later value as `CHANGED`. Sorting all-scalar lists (which the script already did) fixes a reorder, but not an insertion.

## The damage

On a real snapshot comparison where three values were added to a 32-value enum declared by two schemas:

```
CHANGED (10)  ->  CHANGED (2)
```

The eight that disappeared looked like this, and every one was an artifact:

```
~ ...Candidate.properties.reason.enum[28]: 'unenrolled' -> 'teacher_extended'
~ ...Candidate.properties.reason.enum[29]: 'user_requested' -> 'teacher_restored'
~ ...Candidate.properties.reason.enum[30]: 'within_maximum' -> 'teacher_withdrew'
~ ...Candidate.properties.reason.enum[31]: 'within_target' -> 'unenrolled'
```

Nothing changed and nothing was dropped — three values were added and the rest shifted down one. The two survivors are genuine (operationIds whose line numbers moved).

This matters beyond tidiness. For a tool whose job is catching regressions, it both invents drift **and** buries whether anything was actually lost, under noise proportional to how far up the list the insertion landed.

## The fix

Enum entries are keyed by value instead of index, so an added value is one `ADDED` entry, a removed value is one `MISSING` entry, and nothing else moves. `repr()` is used in the key so a mixed-type enum keeps `1` and `"1"` distinct.

## Verified

| case | before | after |
|---|---|---|
| 3 values inserted into a 32-value enum | 4 spurious `CHANGED`, exit 1 | 3 `ADDED`, exit 0 |
| a value genuinely dropped | reported, exit 1 | reported by name, exit 1 |
| identical mixed-type enum (`1`, `"1"`, `true`, `null`) | exit 0 | exit 0 |
| int `1` replaced by string `"1"` | exit 1 | `MISSING 1` + `ADDED '1'`, exit 1 |

Re-run against the real snapshot that prompted this: `CHANGED` 10 -> 2, `MISSING` unchanged at 13 (all genuine), `STATUS CHANGES` unchanged at 3 (all genuine).

## Deliberately not included

`required` and `tags` are the same set-like shape and have the same insert-cascade. They are named in the `SET_VALUED_KEYS` comment but left positional, because only `enum` has actually produced false drift in practice — adding them is a one-word change when it does.

Parameter arrays have the same problem for a different reason (they are objects, not scalars, so they are keyed positionally): inserting a header parameter at index 0 reports the path parameter after it as renamed. Fixing that needs keying by `(name, in)` rather than a set, so it is out of scope here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved specification comparisons for enum lists.
  * Added clear individual results when enum values are added or removed.
  * Prevented false change reports caused by shifts in enum ordering.
  * Preserved positional comparison behavior for other scalar lists.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -59,6 +59,13 @@ jobs:
     - name: Run integration tests
       run: go test -v ./cmd/apispec/...
 
+    # The snapshot comparator decides what counts as drift, so its own key
+    # semantics need guarding: keyed positionally, an inserted enum value or
+    # parameter reports everything after it as changed, which hides whether
+    # anything was really dropped.
+    - name: Comparator self-test
+      run: bash scripts/compare-spec.sh --self-test
+
     - name: Generate coverage report
       run: go tool cover -func=coverage.out
 
--- a/scripts/compare-spec-selftest.py
+++ b/scripts/compare-spec-selftest.py
@@ -1,0 +1,145 @@
+#!/usr/bin/env python3
+# Copyright 2026 Ehab Terra
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Self-test for compare-spec.sh's KEY SEMANTICS.
+
+The comparator's job is to tell drift from the same thing said differently, and
+that judgement lives entirely in how it keys values. Keyed by position, a list
+reports an INSERTION as a change to every member after it — which both invents
+drift and buries whether anything was genuinely dropped, the one thing the tool
+exists to answer.
+
+Each case writes a pair of tiny specs, runs the real comparator over them, and
+asserts the reported counts and the exit status. A regression then surfaces as a
+named failing case here, instead of as noise in a real project comparison months
+later.
+
+Run via: scripts/compare-spec.sh --self-test
+"""
+
+import json
+import os
+import subprocess
+import sys
+import tempfile
+
+
+def spec(schema=None, params=None):
+    doc = {"openapi": "3.1.1", "info": {"title": "t", "version": "1"}, "paths": {}}
+    if schema is not None:
+        doc["components"] = {"schemas": {"S": schema}}
+    if params is not None:
+        doc["paths"] = {
+            "/x": {"get": {"parameters": params, "responses": {"200": {"description": "OK"}}}}
+        }
+    return doc
+
+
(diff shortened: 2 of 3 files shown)


ehabterra:
Both resolved in `9c6d95d`, and parameters are fixed too since they have the same defect for a different reason.

### Structured `enum` values (Major)

Correct — the scalar guard meant an enum of objects or arrays fell through to positional keying and kept cascading. Enum members are now keyed by a canonical `json.dumps(..., sort_keys=True)` of the **whole** member, so object and array members are set members like any other. json rather than repr because it gives objects a canonical form while still keeping `1`, `1.0`, `"1"`, `true` and `null` apart.

### Enum keys built from raw values (Minor)

Also correct, and a genuine inconsistency: `canon_val()` normalised the value while the key had already been built from the raw one, so two spellings of the same `$ref` would key differently and report one MISSING plus one ADDED where there is no drift. Key and value are now canonicalised together in `canon_entry()`, and `canon_val` recurses so structured members canonicalise throughout. Strict mode still compares literally.

### Parameters — same symptom, different cause

Raised separately on #348's snapshot run and fixed here. Parameters are objects, so the set-keying above does not reach them; they are keyed by `(name, in)`, their identity per the spec. Inserting a header parameter ahead of a path parameter previously reported the path parameter as *renamed to the header* and its `required` as dropped:

```
~ ....parameters[0].name: 'id' -> 'Retry-After'
~ ....parameters[0].in:   'path' -> 'header'
- ....parameters[0].required = True
```

It falls back to positional keying when the identity is unusable — a `$ref`'d member has no `name`, and duplicate `(name, in)` pairs would collide — so precision never drops below what it was.

### Regression coverage

Both reviews asked for it, and it did not exist for this script at all. `scripts/compare-spec-selftest.py` runs the real comparator over pairs of tiny specs and asserts reported counts **and** exit status across eleven cases: enum insertion, removal, mixed types, int-to-string, object members inserted and dropped, `$ref` spelling, parameter insertion, removal, attribute change, and the `$ref` fallback. Run with `scripts/compare-spec.sh --self-test`, and wired into the test workflow so CI runs it.

Checked that it actually catches regressions rather than merely passing — reverting both mechanisms fails **5 of 11**:

```
FAIL enum: values inserted
FAIL enum: value dropped still fails
FAIL enum: int replaced by string is drift
FAIL enum: object members inserted
FAIL parameters: inserted ahead of an existing one
```

### Effect on the two real snapshot comparisons

| | before | after |
|---|---|---|
| project A | `CHANGED (10)` | `CHANGED (2)` |
| project B | `MISSING (1)`, `CHANGED (2)` | `MISSING (0)`, `CHANGED (0)` |

`STATUS CHANGES` unchanged in both (3 and 17) — those are genuine source changes. Project B is now entirely free of false drift; project A's two survivors are real operationIds whose line numbers moved.

One extra, because `--compare-files` (added for the self-test) makes it reachable by hand: a non-spec input now names the offending file instead of dying with an `AttributeError` inside the walk.

Issue #340, opened 2026-08-12
Spec output is not deterministic on a project large enough to truncate: the same command drops different responses between runs


ehabterra:
## What

On gitea, the same binary and the same inputs produce two different specs across runs. Determinism is golden rule #1, and `TestGenerateDeterministic` guards it — on fixtures small enough that they never truncate.

Observed while A/B-testing something unrelated:

```
$ for i in 1..N; do apispec --dir <gitea> --output out.yaml \
      --max-nodes-per-route 5000 --max-instances-per-key 10; sha256 out.yaml; done

478c93378a08   572486 bytes   (common)
b7de309fb7d9   572281 bytes   (rarer, ~2 runs in 15)
```

Both outputs came from **both** binaries I was comparing, so it is not a regression from any one change — it is a property of the run.

## What differs

The rarer output is **missing responses**. The diff is entirely additions in the common one:

```
>         "500":
>           description: Internal Server Error
>           content:
>             application/json: {}
...
>         "200":
>           description: OK
>           content:
>             application/json: {}
```

So a run can silently document fewer responses than the previous run of the same command. That is worse than a cosmetic ordering flip: the spec is smaller, and nothing says why.

## Not time-dependent

There is no deadline or time-based limit anywhere in the analysis path (`grep` for `time.After`, `context.WithTimeout`, `Deadline` in `internal/spec` and the engine turns up nothing outside phase timing). So this is not a slow-machine effect — it points at an ordering-dependent decision, which in Go means a map range whose order can reach the output.

The likely shape, given the missing content: a tie broken by map iteration order somewhere in the truncation path. gitea at these limits hits all three budgets (#328), and the instance cap refuses copies in the order it encounters them — if that order comes from a map, which copies survive is a coin flip, and with them their responses.

## Why the existing guard does not catch it

`TestGenerateDeterministic` and `TestGenerateMetadataDeterministic` run over `testdata/` fixtures. None of them are large enough to exhaust a budget, so the truncation path they would need to exercise never runs. The invariant is tested where it holds and not where it breaks.

## Suggested direction

1. Reproduce in-repo: a fixture (or an existing one with the limits forced very low, e.g. `--max-instances-per-key 1`) that truncates, run twice in one test, and compare. Truncation is the suspected trigger and needs to be *in* the determinism test rather than incidental to it.
[...]

PR #351, opened 2026-08-15
docs(changelog): record the changes since 0.5.6

ehabterra:
For the v0.5.7 release. `[Unreleased]` had **two entries where twenty-five merges had landed** — so the notes would have shipped without either of the two largest changes in the release, plus the four dangling-`$ref` fixes and the project-package misclassification.

## Contents

Grouped per Keep a Changelog: **Added** (`$ref` repair/report, Homebrew + apispecui distribution), **Changed** (instance-cap default — kept as written — plus the upgrade-diff note), **Fixed** (framework registry, project-package classification, four `$ref` causes, the UI crash — kept as written, Homebrew runtime dep), **Performance** (the prune, indexed lookups, chain-on-stack, lookup ordering + funclit walk), and **Known issues**.

## Three things worth reviewing specifically

**Specs gain content on upgrade.** Pruning frees the per-route and instance budgets for code that can actually contribute, so bodies and responses that were being truncated away reappear — six of 36 routes on this repo were hitting the per-route limit and no longer are; the ~900-route project gains 260 endpoints. Nothing is removed. 0.5.6 set the precedent for saying this out loud, and anyone diffing their spec after upgrading should read it before they see it.

**A `Known issues` section for #340.** On a project large enough to truncate, the same command can document a route's responses in one run and omit them in the next — a truncation decision broken by map-iteration order. It predates this release and is not fixed by it, but it is *more visible* now, because more content is documented and therefore sits nearer a budget boundary. I measured it at default limits on the ~900-route project: two distinct outputs across four runs. Shipping without mentioning it leaves a user to discover a known defect by diffing two runs of their own. Happy to drop the section if you would rather not lead with it.

**The stdlib bug is described by what it did.** "Classify project packages by module path" understates it: on any domain-hosted module — essentially every real Go project — root inference produced no root, and the fallback classified `net/http`, `go/types`, `encoding/json` and nine others as project code.

## Notes

- Measurements are the PRs' own, quoted by project shape (163-route service, ~900-route project) rather than by name, per the convention already in the file.
- Left as `## [Unreleased]`: renaming it to `## [0.5.7] - <date>` and adding the `[0.5.7]` link reference at the bottom are the two mechanical steps at tag time.
- Purely internal merges are deliberately absent (stray-fixture-binary guard, the revert/reland pair, the snapshot-comparator fix) — they change nothing a user of the tool can observe.

## Summary by CodeRabbit

* **Performance**
  * Improved processing performance and documented new performance measurements.
  * Increased default limits and serialized concurrent processing for more consistent results.

* **Improvements**
  * Improved completeness, `$ref` repair, framework detection, package classification, component retention, and type normalization.
  * Enhanced handling of named containers.
  * Specification output is now additive.

* **Distribution**
  * Added Homebrew distribution support.

* **Known Issues**
  * Documented potential nondeterministic results for truncated runs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+Speed and completeness. The walk that builds the spec no longer visits code that
+cannot contribute to it, which on the ~900-route project from 0.5.6 takes the
+documented surface from **640 paths to 900** while halving peak memory. Four
+separate causes of dangling `$ref`s are fixed, and a final pass now guarantees
+the document resolves.
+
+### Added
+
+- **`$ref`s that the document cannot satisfy are repaired and reported.** A
+  single unresolvable reference makes Swagger UI refuse the whole document, and
+  the four causes fixed this cycle were all found the same way — by loading the
+  output into a viewer, which is not a check anyone runs against their own
+  project. Generation now ends by making the document internally consistent: a
+  missing target is **repaired, not dropped**, since removing the reference
+  would silently change an operation's shape. The report names the **Go** type,
+  not the mangled component name, because that is what tells you which
+  dependency to register under `externalTypes`. (#327)
+- **Homebrew tap, and install instructions that install.** `brew install` now
+  works for both binaries — `apispecui` is built, released and distributed the
+  same way as `apispec` rather than being source-only. (#309, #335)
+
 ### Changed
 
 - **`--max-instances-per-key` now defaults to 100 (was 25).** The reason is how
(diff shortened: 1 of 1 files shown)


ehabterra:
Checked, and the contradiction is real — but the changelog is the side that is right, so I have fixed the docs rather than the entry.

Both formulae already declare a **runtime** dependency:

```ruby
# packaging/homebrew/apispec.rb:17, apispec-homebrew-core.rb:34
depends_on "go"        # not `=> :build`
```

That is what #312 fixed, and it is correct: apispec analyses a project by loading its packages through `go/packages`, which shells out to `go list`. Verified against a release build with the toolchain removed rather than taken from the PR title:

```console
$ env -i PATH=/usr/bin:/bin apispec --dir . --output out.yaml
failed to load filtered packages: err: go command required, not found
```

`packaging/homebrew/README.md` claiming "it needs no Go toolchain" was repeating exactly the build-vs-runtime confusion #312 fixed in the formulae, left uncorrected in the prose.

Following it up turned out to matter more than the one line: the same claim was in the installation guide in its most authoritative place —

> **Prerequisites**
> **None** for the pre-built binaries below — they are self-contained.

— and the from-source section ended with "Use method 1 if you have no Go toolchain", which is the one piece of advice guaranteed not to work. A user without Go who followed the guide got a tool that could not analyse anything, having been steered there on purpose.

Fixed in **#352** (installation guide, packaging README, both formula templates, main README), kept separate so this PR stays a changelog change. This one is unchanged and the entry stands as written.

Issue #380, opened 2026-08-19
Variable-origin tracing gives up on a builder method that is not in the package's first file: the negative method-lookup cache is filled after scanning ONE file


ehabterra:
## What

`traceVariableOriginHelper` (internal/metadata/analysis.go) resolves an assignment like

```go
cmd := builder.New().WithArgs("clone")
```

by following the producing call into the callee package and, when the callee is a *method*, reading the method's return var so the origin lands on the receiver. That method lookup is memoized by `metadata.methodLookupCache`, keyed `pkg + "." + funcName` — **package** scope. But it is performed **inside the per-file loop**, and the miss is cached before the other files are looked at:

```go
for _, calleeFileName := range metadata.SortedFileNames(assign.CalleePkg) {   // file loop
    ...
    if !exists {
        for _, typeName := range metadata.SortedTypeNames(...) { /* scan THIS file only */ }
        if calleeMethod == nil && metadata.methodLookupCache != nil {
            metadata.methodLookupCache[methodKey] = nil          // <- package-wide "not found"
        }
    }
    if calleeMethod != nil { /* follow the return var */ }
}
```

The next iteration reads the cache, gets `exists == true, calleeMethod == nil`, and skips its own scan. So **the method is only ever found if it is declared in the package's first file**; otherwise the loop falls through and the trace stops at the local assignment.

## Why it matters

The wrong answer is silently different, not obviously absent: the variable resolves to itself instead of to its producer, so the tracker's `assignIndex` lookup finds no producer, the argument node expands into nothing, and whatever the producer's subtree would have contributed (request/response bodies, params reached through the builder) is missing from the spec.

## Reproduction

`internal/metadata/trace_origin_determinism_test.go` (added with #340) is exactly this shape: package `builder` has `aux.go` (no methods) sorting before `command.go` (which declares `func (c *Command) WithArgs(...) *Command`). Tracing `cmd` answers `{cmd, traceorigin/svc, Build}` — the local assignment — instead of following `WithArgs` to its receiver. The test pins that answer deliberately, as a change detector: it should be updated when this is fixed.

## History

[...]

PR #381, opened 2026-08-19
fix: make resolution deterministic — sort every first-match file/type scan (#340)

### LINKED ISSUE #340: Spec output is not deterministic on a project large enough to truncate: the same command drops different responses between runs

ehabterra:
Fixes #340.

## The bug

The same command on gitea produced different specs run to run. Re-measured on `main` before touching anything: **3 distinct outputs in 4 runs** at `--max-nodes-per-route 5000 --max-instances-per-key 10`, and **two alternating outputs at the defaults**. The difference is a route silently documenting fewer responses, and header parameters (`X-Forwarded-Proto`, `User-Agent`, …) appearing and vanishing.

## Root cause

Not the truncation tie-break the issue theorised. Found by dumping the tree's full materialization sequence (`parent -> child`, 1.1M lines) for two runs and diffing to the **first** divergent node — everything before it, including every counter, was identical.

Variable-origin resolution scans a package's `Files` **map** and answers from the first file that declares what it is looking for. Which file that is was decided by Go's per-run map randomisation, and the answer is memoized for the rest of the run. On gitea, tracing `cmd` in `modules/git.Clone` either found gitcmd's builder method and followed its receiver, or hit the method lookup's negative cache first and stopped at the local assignment. Those two answers put `HandleGitCmdHTTPRedirection` in or out of the reachable set — 9 identities' reachability flipped with it — which changed which subtrees the barren-subtree prune (#348) kept, and therefore which copies the instance cap dropped.

So the chain is: one map-order coin flip during loading → different producer resolution → different reach index → different truncation victims → a smaller spec, with nothing to say why.

## The fix

1. **Sorted every remaining first-match scan** through the memoized `Metadata.SortedFileNames` / `SortedTypeNames` — the helpers that exist for exactly this (`"Sorting IS the determinism guarantee"`): both loops in `traceVariableOriginHelper`, `resolveIdentReturnType`, the function-signature resolver, `BuildAssignmentRelationships`, and the spec layer's function/type/const/method lookups (`body_source`, `field_resolver`, `context_provider`, `extractor`, `tracker`, `wrapper_detect`). `collectImports` sorts its result for the same reason.

2. **Made the assignment-relationship order a total order.** Both trackers sorted the links by producing callee ID alone, which is not unique — one edge can be linked to several assignments — and `sort.Slice` is unstable, so the ties fell back to the source map's order. **9,284 of 21,735 links on gitea sit in such a tie**, and the consumers are last-write-wins indexes (`assignIndex`) and per-producer child lists. Extracted as `sortedAssignmentRelationships` (`internal/spec/assignment_order.go`), now shared by the eager and lazy trees instead of duplicated.

## Verification

| | before | after |
|---|---|---|
| gitea, issue's flags | 3 distinct outputs / 4 runs | **6/6 identical** |
| gitea, default limits | 2 alternating outputs | **4/4 identical** |

Runtime is unchanged (107s / 102s before vs 93s / 104s after — noise; the sorted lists are cached). Full suite passes unchanged, metadata goldens included: no output drift on any fixture.

## Guards

- `TestTraceVariableOriginDeterministic` (metadata) — a fixture whose builder method lives in one file of a multi-file package; traces the same variable over repeated **fresh** loads and pins the answer. Fails 3/3 without the fix.
- `TestGenerateDeterministicUnderTruncation` (generator) — generates fixtures twice with both budgets forced low, and asserts truncation actually happened so the test cannot quietly stop covering that path. This is the gap the issue identified: every existing determinism fixture is too small to spend a budget.

## Follow-up

#380 — the determinism fix made a precision gap reproducible: the method lookup caches its miss package-wide after scanning one file, so a builder method outside the package's first file is never followed. Deliberately not changed here (this PR changes no resolution results); the new test pins the current answer as a change detector.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency when generating API specifications, ensuring repeated runs produce stable results.
  * Reduced variability in metadata, type resolution, relationship processing, imports, and declaration discovery.
  * Improved handling of truncated generation scenarios for more predictable outputs.

* **Tests**
  * Added regression coverage to verify deterministic specification and metadata generation across repeated runs.
  * Added validation for stable variable tracing and expansion-budget truncation behavior.

### DIFF
--- a/generator/determinism_test.go
+++ b/generator/determinism_test.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/ehabterra/apispec/internal/engine"
 	"gopkg.in/yaml.v3"
 )
 
@@ -42,6 +43,57 @@ func TestGenerateDeterministic(t *testing.T) {
 			}
 		})
 	}
+}
+
+// TestGenerateDeterministicUnderTruncation runs the same fixtures with the
+// budgets forced low enough to truncate, because that is where determinism
+// broke in the field (issue #340): the same command on a large project dropped
+// different responses and header parameters between runs, while every fixture
+// small enough for TestGenerateDeterministic never spent a budget at all.
+//
+// Truncation itself is asserted, so the test cannot quietly stop exercising the
+// path it exists for.
+func TestGenerateDeterministicUnderTruncation(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping determinism fixtures in -short mode")
+	}
+	for _, name := range []string{"complex_chi_router", "group_closure_instances", "mixed_gin_mux"} {
+		t.Run(name, func(t *testing.T) {
+			base, truncated := marshalTruncatedSpec(t, name)
+			if !truncated {
+				t.Fatalf("%s did not truncate at the forced budgets, so this fixture no longer covers the truncation path", name)
+			}
+			for run := 1; run < 3; run++ {
+				got, _ := marshalTruncatedSpec(t, name)
+				if string(got) != string(base) {
+					t.Fatalf("spec for %s differs between runs at a truncating budget:\n%s",
+						name, firstDiffLine(string(base), string(got)))
+				}
+			}
+		})
+	}
+}
+
+// marshalTruncatedSpec generates the fixture with both expansion budgets forced
+// to their harshest useful values, and reports whether either one truncated.
+func marshalTruncatedSpec(t *testing.T, fixture string) ([]byte, bool) {
+	t.Helper()
+	cfg := engine.DefaultEngineConfig()
+	cfg.InputDir = filepath.Join("..", "testdata", fixture)
+	cfg.MaxInstancesPerKey = 1
+	cfg.MaxNodesPerRoute = 200
+
+	eng := engine.NewEngine(cfg)
+	out, err := eng.GenerateOpenAPI()
+	if err != nil {
+		t.Fatalf("GenerateOpenAPI(%s): %v", fixture, err)
+	}
+	data, err := yaml.Marshal(out)
+	if err != nil {
+		t.Fatal(err)
+	}
+	stats := eng.GetExpansionStats()
+	return data, stats.InstanceTruncations > 0 || stats.RouteTruncations > 0
 }
 
 func marshalSpec(t *testing.T, dir string) []byte {
--- a/internal/metadata/analysis.go
+++ b/internal/metadata/analysis.go
@@ -625,7 +625,14 @@ func traceVariableOriginHelper(
 									calleeMethod, exists = metadata.methodLookupCache[methodKey]
 								}
 								if !exists {
-									for _, t := range calleeFile.Types {
+									// Sorted: two types in one file can declare
+									// the same method name, and the first match
+									// wins — and is cached for the whole run.
+									for _, typeName := range metadata.SortedTypeNames(assign.CalleePkg, calleeFileName) {
+										t := calleeFile.Types[typeName]
+										if t == nil {
+											continue
+										}
 										for _, method := range t.Methods {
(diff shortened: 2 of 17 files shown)


ehabterra:
Pushed two follow-ups after the first CI run:

**coverage-ratchet (95.9% vs floor 96)** — the per-call-site sorted scans each carried their own nil-file guard: 12 copies of the same three lines, and 12 branches no fixture can reach. Replaced with `Metadata.SortedFiles` / `SortedTypes` iterators that walk the cached sorted names and skip nil entries, so every call site is a plain loop over real files and the guard exists once. That removed the uncovered branches rather than papering over them with tests for defensive code, and the diff got smaller (the package-existence lookups the loops needed are gone too). New unit tests cover what remains: the lookups' nil-file skip, the iterators' unknown-package / nil-file / stop-early paths, and the assignment-order tie-break. Now 96.0% (exact 96.02, vs main's 96.00).

**CodeRabbit's finding on `findMethodByName`** — valid in the sense that `p.Files[fname].Types` panics on a nil file entry, though the exposure predates this PR (the loop it replaced ranged the same map values). Moot now: that function iterates `SortedFiles`/`SortedTypes`, which never yield nil.

Re-verified after the refactor: gitea 3/3 identical, same hash as before it, 93-94s (unchanged). Full suite and metadata goldens pass.

Issue #382, opened 2026-08-19
A `switch r.Method` inside a closure handler is never split: the route is documented as POST only, with every branch's responses merged onto it


ehabterra:
## What

`splitMethodDispatchRoutes` expands a verb-less registration whose handler branches on `r.Method` into one operation per method. It does that for a **named** handler function (`testdata/method_switch` proves it), but not when the handler is a **closure written at the registration site** — the shape `http.HandleFunc("/x", func(w, r) { switch r.Method { ... } })`, which is at least as common.

The route then falls to `ExtractRoute`'s POST default, and since no split happens, every branch's responses pile onto that one operation.

## Reproduction

```go
// (a) closure handler, inline branch bodies
http.HandleFunc("/closure-inline", func(w http.ResponseWriter, r *http.Request) {
    switch r.Method {
    case http.MethodGet:
        _ = json.NewEncoder(w).Encode(User{})
    case http.MethodPost:
        w.WriteHeader(201)
    }
})

// (b) named factory returning the same closure
http.HandleFunc("/named-methods", namedMethods(h))

// (c) closure whose branches call methods
http.HandleFunc("/closure-methods", func(w http.ResponseWriter, r *http.Request) {
    switch r.Method {
    case http.MethodGet:  h.Get(w, r)
    case http.MethodPost: h.Create(w, r)
    }
})
```

Generated:

```
/closure-inline      POST   [200, 201]      <- one operation, both branches' responses
/closure-methods     POST   [200, 201]      <- same
/named-methods       GET    [default]       <- split works; responses are a separate gap
/named-methods       POST   [default]
```

(a) and (c) show it is the **closure**, not the method indirection: the inline case has its branch bodies right there and still is not split. `testdata/method_switch`, whose handlers are named functions, splits correctly into GET/POST, GET/DELETE and GET/HEAD.

## Why

`methodDispatchFor` resolves the handler with `findFunctionByName(meta, route.Package, bare)` and reads `fn.MethodDispatch`. A closure is not in the package's `Functions` table under a name, so the lookup returns nil and the route passes through unsplit. The dispatch itself is recorded per `metadata.Function`; a func literal needs the same record reachable from the route's handler identity.

## Effect on the spec

[...]

Issue #354, opened 2026-08-16
Non-JSON renderers (XML, plain text, YAML, HTML) are all documented as `application/json`


ehabterra:
## Context

Every response and request body reaches the spec as `application/json`, whatever the handler actually
writes. The renderer name is matched (`(?i)(JSON|String|XML|YAML|ProtoBuf|Data|File|Redirect)` in
`netHTTPResponsePatterns`/`config_gin.go`) but the content type is not derived from it: the mapper falls
back to `defaultResponseContentType = "application/json"` (`internal/spec/config.go:27-28`) unless the
pattern carries an explicit `DefaultContentType` — which today only `http.Error` does
(`text/plain; charset=utf-8`).

The mechanism to fix this already exists (`ResponsePattern.DefaultContentType`, `extractor.go:2439`);
it is simply not wired per renderer.

### Reproduction (gin)

```go
func getItemXML(c *gin.Context)    { c.XML(200, Item{}) }
func createItemXML(c *gin.Context) { var it Item; c.BindXML(&it); c.XML(201, it) }
func plain(c *gin.Context)         { c.String(200, "ok") }
func yamlOut(c *gin.Context)       { c.YAML(200, Item{}) }
```

Generated (abridged):

```yaml
  /items:
    post:
      requestBody:
        content:
          application/json:          # BindXML -> should be application/xml
            schema: {$ref: '#/components/schemas/probe_gin_Item'}
  /items/{id}:
    get:
      responses:
        "200":
          content:
            application/json:        # c.XML -> should be application/xml
  /plain:
    get:
      responses:
        "200":
          content:
            application/json:        # c.String -> should be text/plain
              schema: {type: string}
  /yaml:
    get:
      responses:
        "200":
          content:
            application/json:        # c.YAML -> should be application/yaml
```

### Reproduction (net/http — same result without a framework)

```go
mux.HandleFunc("GET /items/{id}", func(w http.ResponseWriter, r *http.Request) {
    w.Header().Set("Content-Type", "application/xml")
    xml.NewEncoder(w).Encode(Item{})
})
```

→ `content: {application/json: ...}`. The explicit `Content-Type` header the handler sets is not read
either.

### Secondary gap: the `xml:` tag layer

`spec.XML` is declared (`internal/spec/openapi.go:186`) and never populated anywhere in the codebase.
So even once the content type is right, `xml:"id,attr"`, `xml:"ns url"`, and `XMLName xml.Name` produce
[...]

PR #383, opened 2026-08-19
fix: document a body written without WriteHeader as 200, not default (#369)

### LINKED ISSUE #369: Success responses land under `default` instead of `200`: a body written without `WriteHeader` is treated as an undetermined status
## Context

`net/http` states the status of a handler that never calls `WriteHeader`: the first `Write` sends **200**.
That is runtime semantics, not an inference. apispec instead emits the response under `default` with
`description: Status code could not be determined` (`internal/spec/mapper.go:847`), so the most common Go
handler shape produces a spec with **no documented success response**.

This is not an edge case — it is apispec's own `testdata/chi` fixture. Generated from `main`, its
responses are:

| status key | count | source |
|---|---|---|
| `default` ("Status code could not be determined") | 8 | handlers that `json.NewEncoder(w).Encode(v)` with no `WriteHeader` |
| `"400"` | 2 | the two `http.Error` calls |

So the *error* paths are documented precisely and every *success* path is `default`. Same in
`testdata/mux` (4 × `default`). gin and echo fixtures have zero `default`s, because `c.JSON(200, v)`
carries the status in the call — i.e. this hits the whole net/http family: `net/http`, chi, gorilla/mux,
and any project using plain `ResponseWriter` handlers.

### Minimal reproduction

```go
func getItem(w http.ResponseWriter, r *http.Request) {
    _ = json.NewEncoder(w).Encode(Item{ID: r.PathValue("id")})   // implicit 200
}
mux.HandleFunc("GET /items/{id}", getItem)
```

```yaml
  /items/{id}:
    get:
      responses:
        default:
          description: Status code could not be determined
          content:
            application/json:
              schema: {$ref: '#/components/schemas/probe_Item'}
```

Adding `w.WriteHeader(http.StatusOK)` to the same handler produces the correct `"200"`, which confirms the
gap is exactly the implicit case.

### Why it matters more than a cosmetic key

In OpenAPI, `default` means "any status code not otherwise declared" — conventionally the *error*
fallback. Client generators lean on that: `openapi-generator` and `oapi-codegen` map `default` to the
error branch, so a generated client for the spec above has **no typed success return** for the endpoint,
while the 400 error type is fully typed. Documentation viewers render the endpoint as having no success
response.

## Proposed

1. When the response body's write destination is the response writer (the `RequireResponseDestination`
   machinery already establishes this) and no `WriteHeader`/status-setting call is reachable on that path,
   emit **200** — with `201` staying whatever the code says, i.e. this only fills the *absent* case.
   This satisfies golden rule #7 rather than bending it: 200 is what the server sends, so `default` is the
   less honest answer here, not the safer one.
2. Keep `default` for its real meaning — a response whose status genuinely could not be resolved (a status
   variable that does not trace, a helper reached through an unresolved indirection). The distinction is
   worth surfacing in the description text too: today both cases read alike.
3. A handler with a branch that sets a status and a branch that does not keeps both responses (the
   explicit one and the implicit 200) — no merging.
4. `http.ServeFile`/`http.Redirect`-style stdlib helpers that write their own status should map through
   the same table (`302`/`304`), rather than falling to `default`.

## Acceptance

- `testdata/implicit_status`: handlers that (a) only `Encode`, (b) `WriteHeader` then `Encode`,
  (c) branch between the two — asserting `200`, the explicit status, and both, respectively.
- `testdata/chi` and `testdata/mux` regenerate with their success responses under `200`; the structural
  tests gain assertions so this cannot silently regress.
- No drift in gin/echo/fiber fixtures (their renderers already carry a status).

Found by comparing handlers that write through a response helper carrying an explicit status
(`ResponseJSON(w, http.StatusOK, v)`) against plain `ResponseWriter` handlers that do not. The helper form
resolves; the plain form is what falls through to `default`, and the contrast is what makes it visible.

ehabterra:
Fixes #369.

## The bug

`net/http` states it outright: *"If WriteHeader is not called explicitly, the first call to Write will trigger an implicit `WriteHeader(http.StatusOK)`"*. apispec documented that body under `default` — "Status code could not be determined" — so the most common Go handler shape produced a spec with **no success response at all**.

That is not cosmetic. In OpenAPI `default` is the catch-all that client generators map to the **error** branch, so `openapi-generator` and `oapi-codegen` gave those endpoints a typed 400 and no typed success return.

It was apispec's own fixtures: `testdata/chi` had 8 responses under `default` against 2 real statuses (its two `http.Error` calls); `testdata/mux` had 4.

## The fix

`ResponseContext.ImplicitStatus` — config, not a hardcoded 200 — says what the framework sends when a handler writes a body without stating one:

| framework | implicit status | why |
|---|---|---|
| net/http, chi, mux | 200 | the first `Write` implies it |
| fiber | 200 | `c.JSON(v)` sends the context's status; `c.Status()` pairs with the body through the existing `^Status$` pattern |
| gin, echo | none | their renderers always carry the status (`c.JSON(200, v)`), so there is no absent case to fill |

Applied **at pairing, not at extraction**, so the existing precedence is untouched: `w.WriteHeader(201)` before the encode still claims the body, and only a body that no status write claimed takes the implicit one. A handler that states a status on one branch and not on the other keeps both responses.

### What still lands under `default`, deliberately

- **A pattern that has a status argument which failed to resolve.** The handler states a status; we just cannot read it, and answering 200 there would be a guess (golden rule #7). A status write that resolves to nothing now emits a marker fragment — no status, no body, never stored — purely so pairing knows this chain sets a status and refuses the implicit one for the body that follows.
- A body whose write destination never resolved at all.

### Config merging

Carried through merging under #212's rule that being secondary costs a framework nothing: the implicit status is a property of HTTP, so a chi handler still implies 200 when gin sorts first. **Only that field travels** — the writer *type* regexes must not, since they gate which encodes count as responses and one framework's writer types would drop the other's. `TestPrimaryFrameworkOrderInvariance` caught this on the first run.

## Effect

42 fixtures move, every one in the same direction — `default` → `200`, or `[400 default]` → `[200 400]`. A few worth calling out:

```
chi          8 × default          -> 200          (plus its 2 × 400 untouched)
mux/servemux 4 × default          -> 200
fiber        default              -> 200          (c.JSON with no status)
response_writer_provenance        -> 200          the leak-* handlers' real
                                                   response is w.Write([]byte("ok"));
                                                   the leaked Secret stays undocumented
```

`enum_validation` gains a 200 next to its 201 on `POST /users`, which is a pre-existing attribution bug becoming visible rather than a new one: a `switch r.Method` inside a **closure** handler is never split, so all branches' responses pile onto one operation. That was already true before this change (the GET branch's body sat under `default` and was dropped only because it happened to duplicate the 201's body). Filed as **#382** with a minimal reproduction.

## Tests

- `testdata/implicit_status` — the four shapes: implied 200, stated 201, a branch that states 400 and one that does not, a raw `w.Write` of marshalled bytes, and a handler that writes an unreadable status (which must stay `default`).
- `TestTestdata_ImplicitStatus` asserts the exact status set per operation.
- `TestTestdata_NetHTTPSuccessStatusesAreExplicit` asserts no success response under `default` in chi/mux/servemux, and fails if a fixture stops covering the path.
- `internal/spec` unit tests pin which frameworks declare an implicit status, and that it survives being secondary while the writer regexes do not.

Full suite, metadata goldens and lint pass; coverage 96.0% (floor 96).

## Out of scope

Issue point 4 (stdlib helpers mapping through a status table): `http.Redirect` → 302 and `http.NotFound` → 404 already resolve. The one gap left is `http.ServeFile`, which produces no response at all — documenting it needs a content type this change cannot honestly supply, which is #354's territory. Noted on the issue.

## Summary by CodeRabbit

* **New Features**
  * Added configuration for implicit HTTP response status codes when handlers write response bodies without explicitly setting a status.
  * Added an editable implicit response status field to the response context editor.
* **Bug Fixes**
  * Generated API specifications now correctly document implicit `200 OK` responses across supported frameworks.
  * Explicit, unresolved, and default response statuses are now represented more accurately.
* **Tests**
  * Added regression coverage for implicit success statuses, status precedence, and framework-specific response behavior.

### DIFF
--- a/cmd/apispecui/assets/js/config.js
+++ b/cmd/apispecui/assets/js/config.js
@@ -645,14 +645,24 @@ function RequestContextEditor({ rc, onChange }) {
   `;
 }
 
-// ResponseContextEditor edits the writer-type lists only. Body transforms
-// (envelope unwrapping) are nested per-transform structures; those are edited in
-// YAML mode rather than duplicated here, and they round-trip untouched.
+// ResponseContextEditor edits the writer-type lists and the implicit status.
+// Body transforms (envelope unwrapping) are nested per-transform structures;
+// those are edited in YAML mode rather than duplicated here, and they round-trip
+// untouched.
 function ResponseContextEditor({ rc, onChange }) {
   const set = (k, v) => onChange({ ...(rc || {}), [k]: v });
   return html`
     ${area("Writer type regexes (one per line)", lines(rc?.writerTypeRegexes), (e) => set("writerTypeRegexes", toLines(e.target.value)), "^net/http\\.ResponseWriter$")}
     ${area("Writer-compatible type regexes (one per line)", lines(rc?.writerCompatibleTypeRegexes), (e) => set("writerCompatibleTypeRegexes", toLines(e.target.value)), "^io\\.Writer$")}
+    <div class="field">
+      <label class="row" style="gap:6px">
+        <span>Implicit status</span>
+        <${Info} text="Status the framework sends when a handler writes a body without stating one — 200 for net/http, whose first Write implies it. Leave empty for frameworks whose renderers always carry the status." />
+      </label>
+      <input class="input" type="number" value=${rc?.implicitStatus ?? ""}
+        placeholder="200"
+        onInput=${(e) => set("implicitStatus", e.target.value === "" ? 0 : parseInt(e.target.value, 10) || 0)} />
+    </div>
   `;
 }
 
--- a/cmd/apispecui/assets/js/config.js
+++ b/cmd/apispecui/assets/js/config.js
@@ -659,7 +659,7 @@ function ResponseContextEditor({ rc, onChange }) {
         <span>Implicit status</span>
         <${Info} text="Status the framework sends when a handler writes a body without stating one — 200 for net/http, whose first Write implies it. Leave empty for frameworks whose renderers always carry the status." />
       </label>
-      <input class="input" type="number" value=${rc?.implicitStatus ?? ""}
+      <input class="input" type="number" min="100" max="599" step="1" value=${rc?.implicitStatus ?? ""}
         placeholder="200"
         onInput=${(e) => set("implicitStatus", e.target.value === "" ? 0 : parseInt(e.target.value, 10) || 0)} />
     </div>
--- a/generator/testdata_implicit_status_test.go
+++ b/generator/testdata_implicit_status_test.go
@@ -1,0 +1,156 @@
+// Copyright 2026 Ehab Terra
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package generator
+
+import (
+	"sort"
+	"testing"
+
+	"github.com/ehabterra/apispec/spec"
+)
+
+// TestTestdata_ImplicitStatus locks in issue #369: a net/http handler that
+// writes a body without calling WriteHeader sends 200, so that is what the spec
+// must say. Documenting it under "default" gave the endpoint no success
+// response at all — client generators map "default" to the error branch.
+//
+// The fixture covers the three shapes the fix has to keep apart: the status is
+// implied, the status is stated, and one branch states it while the other does
+// not.
+func TestTestdata_ImplicitStatus(t *testing.T) {
+	out := loadTestdataWithFixtureConfig(t, "implicit_status", spec.DefaultHTTPConfig())
+	noDanglingRefs(t, out)
+
+	want := map[string]map[string][]string{
+		"/items/{id}":      {"GET": {"200"}, "DELETE": {"204"}},
+		"/items":           {"GET": {"200", "400"}, "POST": {"201"}},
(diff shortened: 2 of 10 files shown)


ehabterra:
Went through the three findings.

**Applied — extractor-level unit tests (Major).** Fair: the fixture and structural tests cover the output, but `pairAndFillResponses`' state machine is where the implicit status is actually applied and it had no direct test. Added four at that layer, driven through a stub response matcher so no tracker tree is needed: an unclaimed body takes the implicit status, a stated status claims the body instead, a body after an unreadable status write stays undetermined, and a body carrying no implicit status stays undetermined. Checked they are change detectors, not decoration — removing the `statusUnresolved` guard fails the third with `got [200]`.

**Applied in part — UI input range.** Bounded the input to `min="100" max="599"` in the markup. Not clamping in the handler, though: the generic `int` field editor (config.js, `fieldEditor`) is `parseInt(...) || 0` with no range check, and the sibling status field `defaultStatus` is edited through exactly that. Clamping this one field only would make the editor inconsistent for no gain — an out-of-range `implicitStatus` is no more reachable than an out-of-range `defaultStatus` is today.

**Not applied — "an explicitly disabled zero is indistinguishable from an omitted one" in the merge.** Verified against the code and it cannot happen:

1. A user-supplied config never reaches `MergeFrameworkConfigs`. `engine.go` takes it as-is — `if e.config.APISpecConfig != nil { … } else if e.config.ConfigFile != "" { LoadAPISpecConfig(…) } else { ComposeFrameworkConfigFrom(…) }` — and only that last branch merges. The merge combines *built-in* presets, whose values are set in this repo (net/http, chi, mux and fiber declare 200; gin and echo declare none).
2. `implicitStatus` is `omitempty`, like every other int on the config surface (`defaultStatus`, `statusArgIndex`, …), so `0` is already indistinguishable from omitted by construction. Making this one field a pointer to recover the distinction would be inconsistent with the rest, for a case the code path cannot reach.

Full suite, goldens and lint pass on the updated branch.

PR #384, opened 2026-08-19
docs: add an at-a-glance feature list and a forks section

ehabterra:
Two README additions, 28 lines, nothing removed.

## At a glance

The first structured block in the README was the table of contents, so anything summarising the project — a search result, an AI answer, someone skimming the page — had to work from the one-paragraph intro or from the dense prose bullets 300 lines down. This adds a 14-line capability list between the intro and the TL;DR, each line leading with the term people actually type: *no annotations*, *OpenAPI 3.1*, the six routers, *multipart*, *JWT*, *generics*, *validator tags*, *deterministic*, *CI*.

Every claim restates something the README already documents in full further down — no new promises. One is newly accurate rather than newly written: "regenerating an unchanged project yields a byte-identical file" holds on real projects only since #381 closed the last known source of run-to-run variation this morning; before that it was true of fixtures and not of gitea.

`## Why APISpec` is untouched — it answers *why*, the new list answers *what*.

## Forks & derivatives

A short section before `## License` acknowledging [antst/go-apispec](https://github.com/antst/go-apispec), which builds on this project under Apache-2.0, plus a TOC entry.

## Summary by CodeRabbit

* **Documentation**
  * Added an “At a glance” overview of APISpec’s supported capabilities.
  * Documented the `antst/go-apispec` fork and provided guidance for downstream projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,23 @@
 
 **APISpec** analyzes your Go source and generates an OpenAPI 3.1 spec (YAML or JSON). It detects routes for popular frameworks (Gin, Echo, Chi, Fiber, Gorilla Mux, `net/http`), follows the call graph to the real handlers, and infers request/response types from actual code — struct tags, literals, generics, and more.
 
+**At a glance**
+
+- **No annotations, no comments, no code changes** — the spec is derived from your AST and call graph, so it cannot drift out of sync with the implementation.
+- **OpenAPI 3.1** output as YAML or JSON.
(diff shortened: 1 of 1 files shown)


ehabterra:
Both applied.

**"No annotations, no comments … cannot drift"** — fair, and it contradicted the list itself: four bullets down, "Doc comments on handlers become the operation `summary` and `description`", so comments *are* read when present. "Cannot drift" was also an absolute the tool can't promise — an undetected route leaves the spec incomplete, which is a different failure from drift but still a gap. Now: *"**No annotations required** — nothing to add to your code: routes, parameters and bodies come from the AST and the call graph, not from a parallel set of comments you have to keep in sync. (Doc comments you already write are used for `summary`/`description` when they are there.)"* — keeps the search term, drops the overclaim.

**Fork licence** — added, one clause: the fork is Apache-2.0 too.

Issue #386, opened 2026-08-20
gin's per-route middleware takes the handler's place: r.GET(path, mw, handler) documents the middleware


ehabterra:
## What

gin takes its handler chain **variadically** — `r.GET(path, mw1, mw2, handler)`, where the handler is the LAST argument and everything before it is middleware. apispec's gin route pattern names a fixed handler position, so on a route with per-route middleware the configured position holds the **middleware**, and the whole operation is built from it.

## Reproduction

`testdata/auth_gin_perroute` already has the shape (`r.GET("/users/{id}", jwtAuth(), getUser)`). Generated from `main`:

```yaml
  /users/{id}:
    get:
      operationId: auth-gin-perroute.jwtAuth   # should be ...getUser
      security:
        - bearerAuth: []                       # this part works
```

`TestTestdata_AuthPresets` now pins that value as a change detector with a comment pointing here — flip it to `getUser` when this is fixed.

## Why it is separate from #364

#364 is the *wrapped-call* shape — `mux.Handle(p, mw(http.HandlerFunc(h)))`, where the handler is an argument **of** the middleware call — and its fix peels that call. Here the middleware and the handler are **sibling arguments** of the registration; nothing is wrapped, so there is nothing to peel. The two produce the same symptom (the operation documents the middleware) through different wiring, and the gin case needs the route pattern to know that the handler is the last argument rather than a fixed index.

## Effect

Same as #364's: the operationId and summary name the middleware, its own reads become the operation's parameters, and the handler's request body and responses are missing. Auth detection is unaffected — the security requirement resolves either way, which is why `testdata/auth_gin_perroute` passed while documenting the wrong operation.

## Suggested direction

Express the variadic chain in the route pattern rather than special-casing gin in code (golden rule #5): a `HandlerArgFromEnd`-style option meaning "the handler is the last argument, from `HandlerArgIndex` onwards", set on gin's registration patterns. echo is the mirror image — `e.GET(path, h, mw...)` puts middleware AFTER the handler — so a single "last argument wins" rule cannot be shared; the option has to be per-config.

Fixture note: `testdata/auth_gin_perroute`'s handlers have empty bodies, so a fix should give them a real body and status the way `testdata/auth_nethttp_wrap` gained one in the #364 PR — otherwise the misattribution stays invisible in the generated spec.

ehabterra:
[...]

Issue #367, opened 2026-08-16
Generic handler adapters: the response type parameter is emitted as a component literally named after the type parameter (`Res`)


ehabterra:
## Context

A generic adapter that turns a typed function into an `http.Handler` is a common modern Go shape — it is
how huma-style frameworks work, and it is the wrapper most projects hand-roll once and reuse everywhere:

```go
func HandleJSON[Req any, Res any](fn func(context.Context, Req) (Res, error)) http.HandlerFunc
```

apispec resolves the **request** side of this correctly (the type argument reaches the body schema) and
then fails on the **response** side, emitting a component named after the *type parameter* itself.

## Reproduction (net/http, no third-party dependency)

```go
func HandleJSON[Req any, Res any](fn func(context.Context, Req) (Res, error)) http.HandlerFunc {
    return func(w http.ResponseWriter, r *http.Request) {
        var in Req
        if err := json.NewDecoder(r.Body).Decode(&in); err != nil {
            http.Error(w, err.Error(), http.StatusBadRequest)
            return
        }
        out, err := fn(r.Context(), in)
        if err != nil {
            http.Error(w, err.Error(), http.StatusInternalServerError)
            return
        }
        w.WriteHeader(http.StatusOK)
        _ = json.NewEncoder(w).Encode(out)
    }
}

func HandleJSONResponse[Res any](fn func(context.Context) (Res, error)) http.HandlerFunc { ... }

func createUser(_ context.Context, in CreateUserRequest) (User, error) { ... }
func listUsers(_ context.Context) (ListResponse, error)                { ... }

mux.Handle("POST /users", HandleJSON(createUser))
mux.Handle("GET /users", HandleJSONResponse(listUsers))
```

Generated:

```yaml
  /users:
    post:
      operationId: probe_generic.HandleJSON[probe_generic.CreateUserRequest, probe_generic.User]
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/probe_generic_CreateUserRequest'   # correct
      responses:
        "200":
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/probe_generic_Res'               # should be ...User
    get:
      operationId: probe_generic.HandleJSONResponse[probe_generic.ListResponse]
      responses:
        default:
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/probe_generic_Res'               # should be ...ListResponse

components:
  schemas:
    probe_generic_CreateUserRequest:
      type: object
[...]

PR #387, opened 2026-08-20
fix: a middleware wrapped at the registration site no longer replaces the handler (#364)

### LINKED ISSUE #364: A middleware that wraps the handler at the registration site replaces it: the operation documents the middleware, not the handler
## Context

`handlerArgValue` (`internal/spec/pattern_matchers.go:1492`) unwraps exactly one shape at the handler
argument position — a **type conversion**, `http.HandlerFunc(getItem)`. When the argument is a **call** —
`withLogging(http.HandlerFunc(getItem))`, i.e. the standard `func(http.Handler) http.Handler` middleware
applied inline at registration — nothing unwraps it, and the whole operation is built from the
*middleware* instead of from the handler.

The README lists "handler wrappers" among the wiring styles auth detection follows, and it does follow
them for `security`. The **handler identity** does not survive the same wiring.

## Reproduction (net/http, same handlers wrapped and direct)

```go
func withLogging(next http.Handler) http.Handler {
    return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) { next.ServeHTTP(w, r) })
}

// getItem is the real handler: it produces a 200 with an Item.
func getItem(w http.ResponseWriter, r *http.Request) {
    w.WriteHeader(http.StatusOK)
    _ = json.NewEncoder(w).Encode(Item{ID: r.PathValue("id")})
}

// createItem decodes a CreateItem body and returns 201 with an Item.
func createItem(w http.ResponseWriter, r *http.Request) {
    var in CreateItem
    _ = json.NewDecoder(r.Body).Decode(&in)
    w.WriteHeader(http.StatusCreated)
    _ = json.NewEncoder(w).Encode(Item{Name: in.Name})
}

mux.Handle("GET /wrapped/items/{id}", withLogging(http.HandlerFunc(getItem)))
mux.Handle("POST /wrapped/items", withLogging(http.HandlerFunc(createItem)))
mux.HandleFunc("GET /direct/items/{id}", getItem)     // control
mux.HandleFunc("POST /direct/items", createItem)      // control
```

Generated (abridged — the controls are correct, the wrapped routes are not):

```yaml
  /direct/items:
    post:
      summary: createItem decodes a CreateItem body and returns 201 with an Item.
      operationId: probe_wrap.createItem
      requestBody:
        content:
          application/json:
            schema: {$ref: '#/components/schemas/probe_wrap_CreateItem'}
      responses:
        "201":
          content:
            application/json:
              schema: {$ref: '#/components/schemas/probe_wrap_Item'}

  /wrapped/items:
    post:
      operationId: probe_wrap.withLogging      # the middleware
      responses:
        default:
          description: Default response (no response found)
          content:
            application/json:
              schema: {type: object}
  /wrapped/items/{id}:
    get:
      operationId: probe_wrap.withLogging      # same operationId again
      parameters:
        - name: id
          in: path
          required: true
          schema: {type: string}
          x-warning: This parameter is present in the path but not found in the code.
      responses:
        default:
          description: Default response (no response found)
```

Everything the handler states is lost: request body, response schema, status code, summary. On top of
that:

- **Duplicate `operationId`.** Every route wrapped by the same middleware gets that middleware's name, so
  a router with one logging wrapper emits N identical `operationId`s — invalid OpenAPI, and most
  generators fail or silently rename.
- **The middleware's own reads leak in as request parameters**, because the middleware body is what gets
  walked.
- The path parameter picks up `x-warning: ... not found in the code` — the handler that reads it is never
  visited.

## Same defect on a shipped fixture

`testdata/auth_nethttp_wrap` already has this exact wiring (`mux.Handle(p, jwtAuth(http.HandlerFunc(getUser)))`).
Generated from `main`:

```yaml
  /users/{id}:
    get:
      summary: jwtAuth is a custom middleware whose returned closure validates a JWT via golang-jwt.
      operationId: auth-nethttp-wrap.jwtAuth        # should be ...getUser
      parameters:
        - name: Authorization      # the middleware's header read, not the handler's
          in: header
        - name: id
          in: path
          x-warning: This parameter is present in the path but not found in the code.
      security:
        - bearerAuth: []           # this part works
```

It stays invisible because that fixture's handlers (`getUser`, `health`) have empty bodies, and
`TestTestdata_AuthPresets` (`generator/testdata_auth_test.go:75`) asserts only the security requirement —
never the operation's identity or content.

## Proposed

1. At a handler argument position, when the argument is a **call** whose signature is a middleware shape —
   it takes and returns the framework's handler type (`func(http.Handler) http.Handler`,
   `func(http.HandlerFunc) http.HandlerFunc`, and the framework equivalents) — resolve the handler from
   the wrapped argument instead, recursively, since chains nest
   (`auth(logging(http.HandlerFunc(h)))`). The wrapper stays on the route's middleware list, so auth
   detection is untouched.
2. Config-driven and framework-agnostic (golden rule #5) — expressed as "this argument position carries a
   handler that may be wrapped", not as a net/http special case. Shapes to cover: net/http
   `mux.Handle(p, mw(h))`, gorilla `r.Handle(p, mw(h))`, chi `r.Method("GET", p, mw(h))` and
   `chain.Then(h)`-style helpers (`justinas/alice`), gin `r.GET(p, gin.WrapH(mw(h)))`, echo
   `e.GET(p, echo.WrapHandler(mw(h)))`, fiber `adaptor.HTTPHandler(mw(h))`.
3. Honest over wrong (golden rule #7): when the wrapped argument does not resolve to a concrete handler,
   keep today's behaviour but warn — a wrong operation is worse than a reported gap. Fold the warning into
   `--strict` (#297).

## Acceptance

- `testdata/handler_wrapper_attribution`: one handler registered both wrapped and direct, asserting the
  two operations carry the **same** `operationId` suffix, request body and responses; plus two different
  handlers behind the same wrapper asserting **distinct** `operationId`s.
- The same shape in at least two frameworks (net/http + chi or gorilla), per golden rule #5.
- `testdata/auth_nethttp_wrap`'s handlers gain a real body/response so the existing auth fixture would
  catch a regression of this.

Found while auditing the middleware wiring styles the supported routers document — `r.Use`, the
`r.With(mw).HandleFunc(...)` chain form, and per-route wrappers. The defect is plain net/http and affects
every supported framework.

ehabterra:
Fixes #364.

## The bug

`mux.Handle(p, withLogging(http.HandlerFunc(getItem)))` documented the **middleware**. Not under-documented — documented as something else:

| | before | after |
|---|---|---|
| operationId | `…withLogging` | `…getItem` |
| summary | the wrapper's doc comment | the handler's |
| parameters | the wrapper's header reads; `{id}` flagged *"present in the path but not found in the code"* | the handler's, `{id}` resolved |
| request body | missing | the handler's |
| responses | `default` | the handler's `200`/`201` |

## Three independent causes

1. **Identity.** `handlerArgValue` peeled exactly one shape — a type conversion (#221). It now also peels a wrapper **call**, repeatedly, so nested chains (`withAuth(withLogging(http.HandlerFunc(h)))`) resolve to the innermost handler.

2. **Expansion.** A conversion names a *type*, and a type has no body, so `http.HandlerFunc(getItem)` was a dead end anywhere it was not the route's own argument — which is exactly where a wrapper puts it. The tracker now expands what a conversion converts (`LazyNode.convertedValueKeys`). Self-limiting: a conversion of a non-function yields a key with no edges, so `[]byte(payload)` expands to nothing rather than to something wrong.

3. **Provenance.** #269's gate lets a route reach its own code through the handler **argument**, one hop from the registration. A wrapped handler sits two hops down, so the request body was refused. Call and conversion argument hops are now transparent; **variable and selector hops are not** — and since the lateral producer relation attaches only to those two kinds (`argProducerIDs`), #269's own case stays blocked by construction.

## What counts as a wrapper

The shape at the handler position, not the framework:

- a call whose callee **takes and returns the same type** — middleware in any framework, including a `Chain.Then(http.Handler) http.Handler` helper;
- **or** any call handed a **conversion of a function**, which carries the adapters whose own signature is not middleware-shaped (`gin.WrapH(mw(h))`, `echo.WrapHandler(mw(h))`).

Deliberately **not** peeled:

- a handler **factory** (`h.Create()`) — no argument is a handler, which is what #221's factory support needs;
- a call with **several** handler-shaped arguments — which one serves the route is genuinely ambiguous (golden rule #7);
- an **adapter** that takes a function without being middleware-shaped. `HandleRequest(handleCreateUser)` in `testdata/generic` turns a `func(Req) (Resp, error)` into an `http.HandlerFunc`; peeling it would rename those operations, which is #367's question and not this PR's. I checked: an earlier, broader rule did rename them, and this narrowing is what keeps that fixture byte-identical.

## Tests

- `testdata/handler_wrapper_attribution` (net/http) and `testdata/handler_wrapper_attribution_chi` register the **same handlers wrapped and directly**, so the test compares the two rather than a snapshot: whatever the direct route documents — operationId, summary, request body, response set, an unflagged path parameter — the wrapped one must document too. Two frameworks per golden rule #5. Covers the plain-ident wrapper (`func(http.HandlerFunc) http.HandlerFunc`) and the nested chain.
- `testdata/auth_nethttp_wrap`'s handlers gained a real body and status. With empty bodies an operation built from the middleware looked identical to a correct one, which is exactly why that fixture never caught this.
- `TestTestdata_AuthPresets` now asserts **which** handler each guarded operation is attributed to, not just that the security requirement resolved.
- Unit tests at the layer that changed: every peel and every refusal (`handlerArgValue`, `wrappedHandler`, `middlewareShaped`, `calleeFunctionOf`, `namesAFunction`, `convertedValueKeys`), including method values and edge-resolved callees.

## Found on the way

**#386** — gin takes its handler chain variadically (`r.GET(path, mw, h)`), so the configured handler position holds the middleware and the operation is built from it. Same symptom, different wiring: sibling arguments, nothing wrapped, so this PR's peel does not apply. The new auth assertion surfaced it; the expected value is pinned as a change detector with a comment pointing at #386.

Full suite, goldens and lint pass; coverage 96.0% (floor 96). The only fixture output that moves is `auth_nethttp_wrap` (operationId `jwtAuth` → `getUser`, plus the responses its new handler bodies produce) and `wrapper_router`'s `/api/items/count`, whose handler content now resolves through the conversion.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved API specification generation for handlers wrapped in middleware, adapters, and type conversions.
  - Preserved accurate handler attribution and distinct operation IDs for wrapped and multiple handlers.
  - Maintained request-body, response, summary, and path-parameter details when handlers are wrapped.
  - Improved support for method values and nested wrapper patterns while safely ignoring ambiguous cases.

- **Tests**
  - Added comprehensive regression coverage across supported routing frameworks and wrapper scenarios.

### DIFF
--- a/generator/testdata_auth_test.go
+++ b/generator/testdata_auth_test.go
@@ -15,6 +15,7 @@
 package generator
 
 import (
+	"strings"
 	"testing"
 
 	intspec "github.com/ehabterra/apispec/internal/spec"
@@ -79,42 +80,56 @@ func TestTestdata_AuthPresets(t *testing.T) {
 		cfg       func() *spec.APISpecConfig
 		protected struct{ method, path string }
 		open      struct{ method, path string }
+		// handler is the function the protected operation must be attributed to.
+		handler string
 	}{
 		{
 			name:      "auth_chi_with",
 			cfg:       spec.DefaultChiConfig,
 			protected: struct{ method, path string }{"GET", "/users/{id}"},
 			open:      struct{ method, path string }{"GET", "/health"},
+			handler:   "getUser",
 		},
 		{
 			name:      "auth_echo_group",
 			cfg:       spec.DefaultEchoConfig,
 			protected: struct{ method, path string }{"GET", "/api/me"},
 			open:      struct{ method, path string }{"GET", "/health"},
+			handler:   "me",
 		},
 		{
 			name:      "auth_fiber_group",
 			cfg:       spec.DefaultFiberConfig,
 			protected: struct{ method, path string }{"GET", "/api/me"},
 			open:      struct{ method, path string }{"GET", "/health"},
+			handler:   "me",
 		},
 		{
 			name:      "auth_gin_perroute",
 			cfg:       spec.DefaultGinConfig,
 			protected: struct{ method, path string }{"GET", "/users/{id}"},
 			open:      struct{ method, path string }{"GET", "/health"},
+			// KNOWN WRONG, pinned as a change detector: gin takes its handler
+			// chain variadically (`r.GET(path, mw, h)`), so the configured
+			// handler position holds the MIDDLEWARE and the operation is
+			// attributed to it. A sibling-argument chain, not the wrapped-call
+			// shape #364 fixes — tracked as #386; flip this to "getUser"
+			// when it lands.
+			handler: "jwtAuth",
 		},
 		{
 			name:      "auth_mux_subrouter",
 			cfg:       spec.DefaultMuxConfig,
 			protected: struct{ method, path string }{"POST", "/api/me"},
 			open:      struct{ method, path string }{"POST", "/health"},
+			handler:   "getUser",
 		},
 		{
 			name:      "auth_nethttp_wrap",
 			cfg:       spec.DefaultHTTPConfig,
 			protected: struct{ method, path string }{"GET", "/users/{id}"},
 			open:      struct{ method, path string }{"GET", "/health"},
+			handler:   "getUser",
 		},
 	}
 
@@ -149,6 +164,21 @@ func TestTestdata_AuthPresets(t *testing.T) {
 			if oOp.Security != nil {
 				t.Errorf("%s %s: expected no security (sibling of guarded route), got %v",
 					tc.open.method, tc.open.path, *oOp.Security)
+			}
+
+			// The guarded operation must still be the HANDLER's, not the
+			// middleware's. Asserted here because auth_nethttp_wrap wires the
+			// guard as `mux.Handle(p, jwtAuth(http.HandlerFunc(getUser)))` — the
+			// exact shape that used to document jwtAuth as the operation, with
+			// the handler's response missing. Security passing while the
+			// operation is wrong is precisely what this fixture missed (#364).
+			if len(pOp.Responses) == 0 {
+				t.Errorf("%s %s: the handler's response is missing — the operation may have been attributed to the middleware (operationId %q)",
+					tc.protected.method, tc.protected.path, pOp.OperationID)
+			}
+			if !strings.HasSuffix(pOp.OperationID, "."+tc.handler) {
+				t.Errorf("%s %s: operationId %q should be attributed to %s",
+					tc.protected.method, tc.protected.path, pOp.OperationID, tc.handler)
 			}
 
 			// The bearerAuth scheme must be catalogued in components.
--- a/generator/testdata_handler_wrapper_test.go
+++ b/generator/testdata_handler_wrapper_test.go
@@ -83,22 +83,46 @@ func TestTestdata_HandlerWrapperAttribution(t *testing.T) {
 					t.Errorf("%s %s: summary %q, want the handler's %q",
 						pair.method, pair.wrapped, wrapped.Summary, direct.Summary)
 				}
-				if (wrapped.RequestBody == nil) != (direct.RequestBody == nil) {
-					t.Errorf("%s %s: request body presence differs from the direct registration (wrapped=%v direct=%v)",
-						pair.method, pair.wrapped, wrapped.RequestBody != nil, direct.RequestBody != nil)
+				// The body must be the same SCHEMA, not merely present: a
+				// wrapper that resolved to a generic object would pass a
+				// presence check while documenting nothing useful.
+				if got, want := bodySchemaRef(wrapped), bodySchemaRef(direct); got != want {
+					t.Errorf("%s %s: request body schema %q, want the handler's %q",
+						pair.method, pair.wrapped, got, want)
 				}
 				if got, want := sortedStatusKeys(wrapped), sortedStatusKeys(direct); !sameStrings(got, want) {
 					t.Errorf("%s %s: responses %v, want the handler's %v",
 						pair.method, pair.wrapped, got, want)
 				}
+				for _, status := range sortedStatusKeys(direct) {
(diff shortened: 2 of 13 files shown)


ehabterra:
All four applied — each was valid on inspection.

**`baseTypeName` string comparison → structured refs.** Correct call, and it was my own trimming reinvented. `middlewareShaped` now compares `typemodel.TypeRef` values from `Metadata.TypeRefOf`, matching on `Name`, with `Pkg` compared only when both refs carry one — a parameter records its type as *written* (`Handler`), the result as *resolved* (`http.Handler`), so the names are what must match. Pointer unwrapping is `Kind == KindPointer` rather than a `TrimPrefix("*")`. `baseTypeName` is deleted.

**Method fallback in `calleeFunctionOf`.** Real gap: the edge path already consulted the method table, the no-edge path did not, so `chain.Then(getItem)` — alice's shape, and one the issue lists — resolved to nothing, its middleware shape was unknown, and a plain-ident handler behind it was not peeled. Added, with a unit case; I confirmed it fails without the fallback (`got ""`).

**Schemas, not presence, in the wrapped-vs-direct comparison.** Also right — a wrapper that resolved to a generic `object` would have passed the old check. It now compares the request-body schema, each response's schema, and the path parameter's own schema and `required` flag, and requires the parameter to *exist* rather than only checking the ones that happen to be there.

**Direct tests for the provenance allow-list** (nitpick). Worth having, since that gate is what protects #269: a call or conversion hop beyond the handler argument is accepted, an ident or selector hop at the same depth is not — the distinction the change rests on, now asserted rather than argued in a comment.

Full suite, goldens and lint pass; coverage still 96.0%.

PR #423, opened 2026-08-31
docs(readme): trim development history, reorganize, and fix gaps

ehabterra:
The README had grown to 1181 lines, and most of the growth was the *story* of
how features were built rather than what they do — benchmark tables from single
measurement sessions, "the default moved from 25 to 100 because…", issue numbers
for closed work, "used to document nothing" framing, and ~20 `See testdata/x/`
pointers that only mean something to a contributor. Four overlapping feature
lists ("At a glance", "Why APISpec", "Framework Support", "Go Language Support")
said the same things in four places.

## Moved, not deleted

The detail is real — it just isn't front-page material:

- **`docs/PIPELINE.md`** — the ten pipeline stages, previously ~60 lines of
  Role/Purpose/Importance prose in the README, with the mermaid diagram.
- **`docs/PERFORMANCE.md`** — lazy vs eager engines, the reasoning behind every
  limit, the `--max-instances-per-key` measurements, and profiling.

Both are linked from the README and from `docs/README.md`.

## Reorganized

Reading order now answers a newcomer's questions in the order they ask them:
Demo → Quick Start → Framework Support → What APISpec Understands → The Tools →
Configuration → How It Works → Performance → Development.

The old "Go Language Support" wall of paragraph-length bullets is grouped under
**Routing & handlers / Requests / Responses / Types & schemas / Auth**, with the
caveats that were buried mid-sentence collected into one **Not supported (yet)**
list. The 60-row flag table and the worked examples are collapsed into
`<details>`, so they stay reachable without setting the page's apparent length.

## Content fixes found while checking the prose against the code

- `--verbose`/`-vb` was missing from the flag table, though two sections tell
  you to run with it. The six profile output-path flags and `--metrics-path`
  were missing too — `metrics.json` was named with no flag documented to
  produce it.
- A positional directory argument overrides `--dir` (`main.go:357`) —
  undocumented.
- Two limitations lost in the condensing, restored: `r.Method` dispatch inside a
  **receiver-method** handler doesn't split (`findFunctionByName` searches
  `file.Functions`, never `Type.Methods`), and a cross-package generic type
  argument resolves but drops its package from the component name.
- The four `/api/insight/*` endpoints were absent from the `apispecui` table
  while the Insight tab was described in prose. Added with the query parameters
  read from the handlers.
- The eager tracker's deprecation is now a callout carrying the numbers behind
  it (194 of ~280 routes, silently; 1.6× slower; four fixture shapes wrong)
  rather than a clause.

## Verification

Checked against the source, not assumed:

- Every default in the flag table matches `internal/engine/engine.go` and
  `internal/spec/lazytree.go`.
- Go version matches `go.mod`; coverage badge matches
  `scripts/coverage-floor.txt`.
- All 27 relative links resolve; no anchor points at a removed section
  (`docs/README.md#contributing` still works).
- Both testable "not supported" claims still hold: no `binding:` tag handling
  anywhere, no `gt`/`gte`/`lt`/`lte` mapping.
- "Six routers" stands despite a seventh registry entry — `fasthttp` is
  classified during dependency analysis but has no route-extraction config.

Docs only; no code touched.

## Summary by CodeRabbit

* **Documentation**
  * Reorganized and condensed the main README with clearer feature-focused sections and updated navigation.
  * Expanded guidance for framework detection, routing, authentication, configuration, CLI usage, UI endpoints, responses, and schema comments.
  * Added pipeline documentation describing analysis from source code through OpenAPI output.
  * Added performance documentation covering analysis limits, profiling, warnings, and tracker engine behavior.
  * Updated the documentation index with links to the new guides.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,24 +17,16 @@
 
 **APISpec** analyzes your Go source and generates an OpenAPI 3.1 spec (YAML or JSON). It detects routes for popular frameworks (Gin, Echo, Chi, Fiber, Gorilla Mux, `net/http`), follows the call graph to the real handlers, and infers request/response types from actual code — struct tags, literals, generics, and more.
 
-**At a glance**
+**Point APISpec at your module. Get an OpenAPI spec** — plus, optionally, an interactive call-graph diagram and a browser-based config UI.
 
-- **No annotations required** — nothing to add to your code: routes, parameters and bodies come from the AST and the call graph, not from a parallel set of comments you have to keep in sync. (Doc comments you already write are used for `summary`/`description` when they are there.)
-- **OpenAPI 3.1** output as YAML or JSON.
-- **Six routers out of the box** — Gin, Echo, Chi, Fiber, Gorilla Mux, and `net/http` (including Go 1.22 `ServeMux` method-aware patterns, `{id}` wildcards and `r.PathValue`), plus mixed multi-framework projects in one binary.
-- **Your own router, detected automatically** — house wrapper types and house contexts (`func (r *Router) Get(...)`, `ctx.JSON/Bind/Query`) need no configuration.
-- **Request bodies** from `json.Decode`/`Unmarshal`, framework binders, custom wrapper helpers, and form / multipart file uploads (`multipart/form-data` vs `x-www-form-urlencoded`).
-- **Responses and status codes** per branch — conditional statuses, envelopes, generic wrappers, map-literal payloads, and interface-typed bodies resolved to the concrete type actually written.
-- **Parameters** in path, query, header, cookie and form position.
-- **Auth and security detection** — bearer/JWT, basic and apiKey schemes, with middleware followed through router-wide `Use`, group closures, per-route chains and handler wrappers.
-- **Type-aware schemas** — generics (including inferred instantiations), aliases, enums from constants, fixed-size arrays, pointers, embedded and inline structs, and external package types.
-- **Validation constraints** from `go-playground/validator` tags — `required`, formats, patterns, and length/value/item bounds routed by field type.
-- **Doc comments** on handlers become the operation `summary` and `description`.
-- **Deterministic output** — regenerating an unchanged project yields a byte-identical file, so the spec can be committed and diffed in CI without false failures.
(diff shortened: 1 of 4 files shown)


ehabterra:
Thanks — verified each finding against the code. Four applied, one rebutted, one filed.

**Applied**

- **`docs/PERFORMANCE.md:135` — trace flag.** Correct, and it was wrong in the README before the move too: `go tool trace profiles/trace.out` reads a file the command never writes. Added `--trace-profile`.
- **`README.md:273` — MD038.** Correct that the span is malformed, but the suggested diff doesn't fix it: `` `` `` followed by content ending in a backtick puts three consecutive backticks at the close, which is a run of 3 and can't close a run of 2. CommonMark requires the space there and strips one from each side on render, and markdownlint's MD038 exempts a span whose content begins or ends with a backtick. Used ``` `` _ struct{} `validate:"gtefield=Min"` `` ```; `markdownlint-cli2` now reports no MD038 on that line.
- **`README.md:541` — endpoint inventory.** The table is deliberately partial, so I retitled it "Key endpoints" and named what it leaves out (config load/save, generation progress/cancel, browsing, health) rather than mirroring all 30 routes.
- **`README.md:975` — binary count.** Correct; the phrase predates `apidiag`. Now names all three.

**Rebutted — `README.md:143`, primary framework**

The row is right as written, and the suggested rewording would weaken a claim that a test pins. `TestPrimaryFrameworkOrderInvariance` (`generator/primary_framework_order_test.go`) generates the multi-framework fixture twice from copies renamed to reverse the detection order and asserts the two specs are equal **in full** — `info` included, not just paths. It passes on this branch. Being secondary costs a framework nothing, which is exactly what #211 and #212 fixed.

The tension you spotted in the prose was real, though: "its defaults and info apply" implied an observable difference that doesn't exist. I've restored the evidence the condensing dropped, so the row now cites the test.

**Filed — `README.md:555`, symlink escape**

Confirmed, and worse than described: `sourcePathAllowed` validates `abs` (`main.go:1820`, lexical `Abs` + `Rel`), but the next line calls `insight.SourceSnippet(pos, …)`, which re-parses the raw `pos` and opens *that* — so the validated path and the read path aren't even the same string. A symlink under the module root passes the check and is served.

This PR touches no Go source, so fixing it here would be out of scope. Filed with a reproduction as **#424**.

Issue #433, opened 2026-09-02
A path wrapped in a type conversion renders as its TYPE (`string(prefix)` → /string/…), so a phantom path is documented


ehabterra:
A path expression wrapped in a **type conversion** is not resolved, and renders
as the conversion's *type* — so the document gains a literal-looking path that
exists nowhere:

```go
type RoutePrefix string

func mountNamed(prefix RoutePrefix, r *chi.Mux, server http.Handler) {
	r.Mount(string(prefix), server)     // documented under /string/…
}

mountNamed(RoutePrefix("/named"), root, Server())
```

`testdata/mount_via_helper` has covered this shape since #275, and its spec
shows the symptom today:

```console
$ apispec --dir testdata/mount_via_helper
  /api/things:        # literal prefix inside a helper
  /direct/things:     # mounted at the call site
  /param/things:      # prefix through a plain parameter — resolves since #431
  /string/things:     # <- mountNamed: the conversion, rendered as its type
```

`/named/things` is the right answer and is absent; `/string/things` is a phantom.
It is not flagged either: nothing marks it as synthesized, so
`noUnresolvedPlaceholders` passes over it and #428's report never fires — it
looks exactly like a real path.

## Why it survives the resolvers

`resolvePathArg`/`resolvePathOperand` answer for a literal, a constant, a
caller's argument, a struct field and (since #431) a local variable. A
conversion is none of those kinds, so it falls through to `GetArgumentInfo`,
which renders an argument it cannot evaluate as its type — the same rendering
that produced `/string` for an unreadable variable before #431.

## Proposed

A conversion to a string kind does not change the value, so resolve the operand
inside it and answer with that:

1. Recognise the shape in the path resolvers — a call whose callee names a type
   rather than a function (`string(x)`, `RoutePrefix(x)`, a named string type).
   Metadata already distinguishes conversions when building the call graph
   (`isTypeConversion`), so the fact is available rather than guessable.
2. Resolve the inner expression through the existing ladder, so
   `string(prefix)` resolves exactly as `prefix` does — including a caller's
   argument (`RoutePrefix("/named")` → `/named`), itself a conversion.
3. Where the inner expression does not resolve, emit a placeholder as usual.
   **Never render the type**: that is what makes the wrong answer invisible.

Worth checking in the same pass whether any other path kind still falls through
to `GetArgumentInfo` — the type rendering is only ever wrong for a path.

## Acceptance

[...]

Issue #428, opened 2026-09-01
An unresolvable registration path is emitted as a placeholder operation (`/{Method} {Path}`, `/{pattern}`) instead of being reported


ehabterra:
When a registration's path cannot be resolved statically, APISpec emits an
operation at a **placeholder path** built from the unresolved expression instead
of reporting the registration as unresolvable. The result is a path no request
can match, sitting in the spec next to the real ones.

Found while auditing the README's "Not supported (yet)" list: both shapes there
are described as *not documented*, and they are in fact *documented wrongly*.

## Reproduction A — routes registered from a table

```go
var routes = []route{
	{"GET", "/users", listUsers},
	{"GET", "/users/{id}", getUser},
}

for _, rt := range routes {
	mux.HandleFunc(rt.Method+" "+rt.Path, rt.Handler)
}
```

```yaml
paths:
  /{Method} {Path}:
    post:
      operationId: p4.route.Handler
      parameters:
        - $ref: '#/components/parameters/MethodParam'
        - $ref: '#/components/parameters/PathParam'
        - name: Method
          in: path
          required: true
          x-warning: This parameter is present in the path but not found in the code.
        - name: Path
          …
      responses:
        default:
          description: Default response (no response found)
```

Note the path also keeps the space from the Go 1.22 `"GET /users"` concatenation,
so the key is not even a syntactically sensible path template.

## Reproduction B — a house router chained through a returned object

```go
func (r *Router) Combo(pattern string) *Combo { return &Combo{r: r, pattern: pattern} }
func (c *Combo) Get(h http.HandlerFunc) *Combo  { c.r.mux.HandleFunc("GET "+c.pattern, h); return c }
func (c *Combo) Post(h http.HandlerFunc) *Combo { c.r.mux.HandleFunc("POST "+c.pattern, h); return c }

rt.Combo("/users").Get(list).Post(create)
```

`--verbose` correctly declines the derivation:

```text
Router wrappers: p5.*Combo [Get Post] route via net/http.HandleFunc (incomplete, not applied)
```

…and then the framework call *inside* the wrapper is matched on its own, so the
spec gets:

```yaml
paths:
  /{pattern}:
    get:  { operationId: 'p5.net/http.HandlerFunc', … }
    post: { operationId: 'p5.net/http.HandlerFunc', … }
```

Two operations, an operationId that is a stdlib type name, and no handler body —
after the wrapper machinery had already concluded it could not resolve this.

## Why it matters

[...]

PR #434, opened 2026-09-02
fix(spec): read a registration path out of a variable

### LINKED ISSUE #431: A path held in a variable is not traced, so a statically-known path is reported as unresolvable
A registration whose path is held in a **variable** does not resolve, even when
the variable's value is a literal written two lines above:

```go
lit := "/literal"
mux.HandleFunc("GET "+lit, h)     // documented at /{lit}, not /literal
```

`resolvePathOperand` (`internal/spec/pattern_matchers.go`) resolves a literal, a
constant (including one from another package), a value the caller passed in, and
a field read off a struct literal — but a plain local variable falls through to
`placeholderFor`, so the path becomes the placeholder `{lit}` named after the
variable.

## Why it matters more now

Before #428 the route was documented at `/{lit}` — a wrong path, but visible.
Since #428 a path that is *nothing but* a placeholder is not documented at all,
so this shape now costs the route entirely (it is reported on stderr and by
`Generator.UnresolvedPaths()`, so it is not silent). That is the right call for
a path genuinely built at runtime; it is the wrong outcome here, because this
path **is** statically knowable.

The concatenated form is unaffected and still documented approximately:

```go
base := "/api/v1"
mux.HandleFunc("GET "+base+"/users", h)   // /{base}/users — kept, prefix flagged
```

so the loss is limited to a path held *entirely* in a variable.

## Reproduction

```go
func buildPath() string { return "/users/" + os.Getenv("VERSION") }

func main() {
	mux := http.NewServeMux()

	p := buildPath()
	mux.HandleFunc("GET "+p, h)      // genuinely unresolvable — reporting is correct

	lit := "/literal"
	mux.HandleFunc("GET "+lit, h)    // resolvable in principle; reported today
}
```

```console
[routes] main.go:26: path is not statically resolvable (/{p}), so handler pvar.h is not documented
[routes] main.go:30: path is not statically resolvable (/{lit}), so handler pvar.h is not documented
Generated (0 paths — every registration builds its path at runtime)
```

The second line is the bug: `lit` has exactly one assignment, and it is a
string literal.

## Proposed

Trace the variable before falling back to a placeholder. The machinery exists in
the spec layer and is already used for request bodies and response
destinations — `latestAssignment` / `assignmentsAt` (`internal/spec/extractor.go`)
give the most recent assignment visible at the call site, with the edge-first
scope that handles a closure-local variable.

Rules that keep it honest (golden rule #7):

1. Resolve only when **every** assignment visible at the site resolves to the
   same value. Two branches assigning different paths is real ambiguity — keep
   the placeholder and let #428 report it, rather than picking one.
2. Resolve transitively through aliases (`a := "/x"; b := a`), bounded like the
   other assignment walks.
3. A variable assigned from a call (`p := buildPath()`) stays unresolved — that
   is the case the placeholder exists for.

## Acceptance

- A `testdata/` fixture with a literal-in-a-variable path (documented at its
  real path), an alias chain, a two-branch ambiguous assignment (reported, not
  guessed), and a call-assigned variable (reported).
- `testdata/route_table`'s partly-resolved route keeps its current output — the
  concat path is not affected by this change.
- No drift on the existing fixtures.

ehabterra:
Fixes #431.

`p := "/users"` two lines above the registration is a statically-known path, and
it was treated as unreadable. Two symptoms, depending on where the variable sat:

| shape | before | after |
|---|---|---|
| `whole := "/users"; mux.HandleFunc("GET "+whole, h)` | reported, left out (#428) | `/users` |
| `bare := "/bare"; mux.HandleFunc(bare, h)` | **`/string`** — the identifier's TYPE, documented as an endpoint | `/bare` |
| `first := "/aliased"; second := first` | reported | `/aliased` |
| `prefix := "/api"; …"GET "+prefix+"/items"` | `/{prefix}/items` | `/api/items` |
| `root := "/v2"; joined := root + "/joined"` | reported | `/v2/joined` |
| `amb := "/first"; if … { amb = "/second" }` | reported | **still reported** |
| `built := buildPath()` | reported | **still reported** |

A bare identifier now goes through the same ladder as one operand of a
concatenation — constant, caller's argument, struct field — extended with the
variable's own assignments. Three rules keep it honest (golden rule #7): every
assignment visible at the call site must agree, an alias chain is followed
(bounded, which also terminates a cycle), and a value that is not itself
readable resolves to nothing.

## The lookup needed to be its own

`assignmentsAt` is the canonical "what does this variable hold here" lookup, and
its edge fast path answers with the assignment **in effect** at the call — right
for a request body or a status, wrong for a path. Measured on
`amb := "/one"; if … { amb = "/two" }`: it resolved to `/two` with no sign that
`/one` existed, and the route was documented at one branch. `pathVarAssignments`
prefers the enclosing function's scope, which records them all;
`assignmentsAt`'s doc comment claimed the two scopes were interchangeable and is
corrected.

## A fixture's change detector fired, as designed

Reusing the operand ladder also resolved a prefix passed to a mount helper as a
parameter — `mountParam("/param", r, sub)` reaching `r.Mount(prefix, server)`.
`testdata/mount_via_helper` has carried a subtest asserting today's *wrong*
behavior since #275, with instructions to update it when fixed; it now asserts
`/param/things` properly.

## Drift

- **103 of the 105 fixtures byte-identical.** `mount_via_helper` gains
  `/param/things`; `variable_path` is new.
- **gitea: 894 paths before and after** — and six phantom `/…/string` keys
  became flagged `{fullPattern}` placeholders:

  ```diff
  - /login/oauth/string:
  + /login/oauth/{fullPattern}:
  - /{username}/arch/string:
  + /{username}/arch/{fullPattern}:
  ```

  `fullPattern := r.getPattern(pattern)` is assigned from a call, so declining
  it is the correct answer — and the placeholder carries the synthesized-param
  warning, where `/string` looked like a real endpoint. No route gained or lost,
  no new reports.
- Full suite green: goldens, determinism, both engines' parity tests, all
  framework fixtures.

## Filed from this work

**#433** — a path wrapped in a type conversion (`r.Mount(string(prefix), sub)`)
still renders as the conversion's type, so `mount_via_helper` documents
`/string/things` and not `/named/things`. It is the last shape that falls
through to rendering an argument as a path; the fixture's subtest pins today's
behavior and names the issue, and `docs/DEBUGGING.md` gains a row for the
symptom.

## Summary by CodeRabbit

- **Bug Fixes**
  - Registration paths stored in local variables are now resolved when assignments agree.
  - Alias chains, prefixes, concatenations, and mount-helper parameters are supported.
  - Conflicting assignments and runtime-derived values remain marked as unresolved.
  - Improved handling of partially unresolved paths and type-converted values.

- **Documentation**
  - Updated troubleshooting and usage guidance for variable-based paths, unresolved placeholders, and type-named path segments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **A registration path held in a variable is now read, instead of being
+  reported as unknowable.** `p := "/users"` two lines above the registration is
+  a statically-known path, and it was treated as unreadable: left out of the
+  document (since the change below), or — when the whole path was one variable —
+  rendered as the variable's *type*, so `/string` appeared as if it were an
+  endpoint. A path argument that is a bare identifier now goes through the same
+  resolution ladder as one operand of a concatenation, plus the variable's own
+  assignments: a literal, an alias chain, a variable prefix with a literal tail,
+  and a value assembled from parts that all resolve. Only when every assignment
+  visible at the call site agrees — two branches assigning different paths stays
+  reported, because picking one would document an endpoint the server may not
+  serve, and so does a path assigned from a call. This also resolves a prefix
+  passed to a mount helper as a parameter (`mountAt(prefix, r, sub)`), which had
+  its own change-detector test. Measured on a real project: six phantom
+  `/…/string` paths became flagged placeholders, with no route gained or lost.
+  (#431)
 - **A registration whose path is built at runtime is reported instead of
   documented at a placeholder path.** A route table
   (`mux.HandleFunc(rt.Method+" "+rt.Path, rt.Handler)`) was emitted as an
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ APISpec aims for practical coverage of real-world Go services.
 
 - Import and type aliases, resolved to underlying primitives.
 - Enum resolution from constants, `enum` tags, or `oneof` validator tags — attributed per declared type, never borrowed between two types sharing an underlying type.
-- Assignment & alias tracking: `:=`, `=`, multi-assign, tuple returns, alias chains, latest-wins shadowing.
+- Assignment & alias tracking: `:=`, `=`, multi-assign, tuple returns, alias chains, latest-wins shadowing. Registration **paths** are read through the same tracking (`p := "/users"`, an alias chain, a variable prefix with a literal tail) — but only when every assignment agrees: two branches assigning different paths is reported as unresolvable rather than documented at one of them.
 - Composite literals, maps, slices, fixed-size and variable-length arrays (`[16]byte`, `[5]int`, `[...]int`), pointers and automatic dereferencing, selectors and nested field access.
 - Struct fields, embedded fields, and tag-based metadata (`json`, `xml`, `form`, `validate`, …).
 - Inline (anonymous) struct types as request/response bodies and as nested fields — captured structurally from `go/types`, so the inline schema shows real properties and resolves named field types to `$ref`s.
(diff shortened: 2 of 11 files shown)


ehabterra:
Verified both shapes — the finding is real, and it is a **precision** limit rather than a correctness one. Probed on this branch:

```yaml
paths:
  /{prefix}/a:   # prefix := "/outer"           should be /outer/a
  /{prefix}/b:   # inner block: prefix := "/inner"   should be /inner/b
  /{p}/c:        # p = "/second" written after it    should be /first/c
  /only/d:       # control — resolves
```

So the failure mode is an approximate path carrying a flagged synthesized parameter, never a wrong one (and a path with no literal part at all is reported and left out by #428). Fewer paths resolve than could, which is worth fixing but not worth guessing at.

Only half of it can be implemented from what is recorded today, which is why it is not in this PR:

- **Reachability** — implementable. `Assignment.Position` plus the call site position would exclude a write below the registration, but not with a plain source-order filter: in `for { p = "/x"; register(p) }` a write further down the body reaches the next iteration. `Function.Blocks` and `internal/spec/control_flow.go`'s `blockIndex.dominates` exist for exactly that question.
- **Lexical binding** — blocked. `Assignment.Scope` is `getScope(name)`, i.e. exported/unexported, not a scope, and the assignment token is not recorded, so `:=` (a new binding) and `=` (a write) are indistinguishable. Block ranges are not a substitute: two assignments in different blocks are as often one binding written twice (`if`/`else`) as two bindings. Approximating it with block ranges would trade this false negative for a false positive — golden rule #9.

Filed as **#436** with both halves and the reproduction. The three shapes are now in `testdata/variable_path` with change-detector assertions (6d91375), so the test flips when either half lands, and the limit is recorded on `pathVarAssignments` where the rule lives.

Issue #436, opened 2026-09-02
A path variable's assignments are matched by NAME only: a shadowed binding or a later write makes a knowable path unresolvable


ehabterra:
Raised by CodeRabbit on PR #434 and **verified**: `pathVarAssignments` compares
every function-scope assignment that shares the identifier's name, without
regard to which binding it belongs to or whether it can reach the registration.
Since the agreement rule then rejects on any disagreement, a path that is
perfectly knowable is reported as unresolvable.

## Reproduction

```go
func main() {
	mux := http.NewServeMux()

	// (1) shadowing: the inner block re-declares the name
	prefix := "/outer"
	mux.HandleFunc("GET "+prefix+"/a", h)
	{
		prefix := "/inner"
		mux.HandleFunc("GET "+prefix+"/b", h)
	}

	// (2) a straight-line write AFTER the registration
	p := "/first"
	mux.HandleFunc("GET "+p+"/c", h)
	p = "/second"
	_ = p

	// (3) control
	only := "/only"
	mux.HandleFunc("GET "+only+"/d", h)
}
```

```yaml
paths:
  /{prefix}/a:   # should be /outer/a
  /{prefix}/b:   # should be /inner/b
  /{p}/c:        # should be /first/c
  /only/d:       # correct
```

`testdata/variable_path` carries these three shapes with a change-detector
assertion, so the test flips when this is fixed.

## Severity: precision, not correctness

The failure mode is a **placeholder**, so nothing wrong is documented: these
paths keep their operation with a flagged synthesized parameter, and one with no
literal part at all is reported and left out (#428). The cost is that fewer
paths resolve than could — and `prefix` reused across a long `main()` is common
enough to matter.

## Two halves, and only one is implementable today

**Reachability (implementable).** `Assignment.Position` is recorded, and the call
site's position is on the node, so writes after the registration can be
excluded. It needs loop care rather than a plain source-order filter: in
`for { p = "/x"; register(p) }` a write that appears later in the body reaches
the next iteration, so excluding it would resolve to a value that is not always
live. The facts for that exist too — `Function.Blocks` records loop bodies, and
`internal/spec/control_flow.go`'s `blockIndex.dominates` was built for exactly
this class of question. Rule sketch: a write counts if it dominates the call
site, or if it shares a loop body with it.

**Lexical binding (blocked, metadata gap).** Nothing recorded distinguishes one
binding of `prefix` from another:

[...]

PR #437, opened 2026-09-02
test: pin the assignment-matching limit behind #436

ehabterra:
Follow-up to the CodeRabbit finding on #434, which landed before this could go
into it. No behavior change — a fixture, its assertions, and a comment.

## The finding, verified

`pathVarAssignments` collects the function-scope assignments that share the
identifier's **name**, so two of them that have nothing to do with each other
read as disagreement and the agreement rule rejects:

```yaml
paths:
  /{shadowed}/a:   # shadowed := "/outer"              should be /outer/a
  /{shadowed}/b:   # inner block: shadowed := "/inner"  should be /inner/b
  /{later}/c:      # later = "/second" written below it should be /first/c
  /only/d:         # control — resolves
```

**Precision, not correctness.** The failure mode is an approximate path carrying
a flagged synthesized parameter, never a wrong one, and a path with no literal
part at all is reported and left out (#428). What it costs is resolutions that
were available.

## Why it is pinned rather than fixed

Half of it cannot be implemented from what metadata records:

- **Reachability** (a write below the registration) *is* implementable —
  `Assignment.Position` plus the call-site position — but not as a source-order
  filter: in `for { p = "/x"; register(p) }` a write further down the body
  reaches the next iteration. `Function.Blocks` and `blockIndex.dominates`
  (`internal/spec/control_flow.go`) exist for exactly that question.
- **Lexical binding** is blocked. `Assignment.Scope` is `getScope(name)` —
  exported/unexported, not a scope — and the assignment token is not recorded,
  so `:=` (a new binding) and `=` (a write to one) are indistinguishable. Block
  ranges are no substitute: two assignments in different blocks are as often one
  binding written twice (`if`/`else`) as two bindings. Approximating it there
  would trade this false negative for a false positive, which golden rule #9
  forbids.

Both halves are written up in **#436** with the reproduction.

## What is here

- the three shapes added to `testdata/variable_path`;
- change-detector assertions in `TestTestdata_VariablePath`: the knowable paths
  must NOT resolve yet, and the approximate ones must stay documented — so the
  test fails, and gets updated, as each half of #436 lands;
- the limit recorded on `pathVarAssignments`, where the rule lives, including
  why a write below the call cannot simply be filtered out.

## Summary by CodeRabbit

- **Documentation**
  - Clarified how generated paths are represented when variable bindings are shadowed or assignments occur later in a scope.
  - Documented that uncertain paths remain approximate placeholders rather than being presented as potentially incorrect concrete paths.

- **Tests**
  - Expanded coverage for variable-path resolution, including shadowed bindings, later writes, and preventing documentation under variable type paths.

### DIFF
--- a/generator/testdata_variable_path_test.go
+++ b/generator/testdata_variable_path_test.go
@@ -46,16 +46,33 @@ func TestTestdata_VariablePath(t *testing.T) {
 		}
 	}
 
-	// Nothing may be documented under a variable's name or type: those are the
-	// two wrong answers this replaces.
+	// Nothing may be documented under a variable's TYPE: that is the wrong
+	// answer this replaces.
 	for path := range out.Paths {
-		if strings.Contains(path, "{") {
-			t.Errorf("path %q keeps a placeholder — every variable here resolves", path)
-		}
 		if path == "/string" || strings.HasPrefix(path, "/string/") {
 			t.Errorf("path %q is the variable's TYPE, not its value", path)
 		}
 	}
+
+	// CHANGE DETECTOR, not an endorsement (issue #436): assignments are matched
+	// by NAME, so a shadowed binding and a write after the registration each
+	// look like two disagreeing values and keep a placeholder. Both paths ARE
+	// knowable. Asserted at today's conservative behaviour — approximate rather
+	// than wrong — so this fails, and gets updated, when #436 lands.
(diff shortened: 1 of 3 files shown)


ehabterra:
@coderabbitai review

Issue #438, opened 2026-09-02
An auth middleware's AuthScheme prefix and custom bearer header are still assumed, not read


ehabterra:
Left over from #370, which shaped an apiKey scheme from the middleware's
configured `KeyLookup`. Two neighbouring facts these middlewares also carry are
still assumed rather than read:

## 1. The value's prefix (`AuthScheme`)

Echo's `KeyAuthConfig.AuthScheme` and fiber's `keyauth.Config.AuthScheme`
default to `"Bearer"` and are routinely set to something else:

```go
e.Use(middleware.KeyAuthWithConfig(middleware.KeyAuthConfig{
	KeyLookup:  "header:X-API-Key",
	AuthScheme: "Token",          // the header value is `Token <key>`
}))
```

The scheme APISpec emits says the key is the whole header value, so a generated
client sends `X-API-Key: <key>` where the server expects `Token <key>`.

OpenAPI has nowhere to put a prefix on an `apiKey` scheme, so this cannot be
expressed structurally. What it can do is *say* so — the scheme's
`description` is the honest place ("value is prefixed with `Token `"), which is
what a reader and a hand-written client need. An empty `AuthScheme` (echo
accepts it, meaning "no prefix") is then worth stating too, since it differs
from the default.

## 2. A bearer token in a custom header

`http`/`bearer` in OpenAPI implies the `Authorization` header. A middleware
configured to read a bearer token from elsewhere is therefore not an `http`
scheme at all — it is an `apiKey` scheme in that header, with the prefix noted
as above. Today such a middleware maps to `bearerAuth` and the custom location
is lost.

This is a modelling decision more than a resolution one: it changes which
scheme *type* a mapping produces, based on a field read at the call site. Worth
doing deliberately, with the mapping declaring the field the way #370's
`lookupField` does, rather than by special-casing a library.

## Acceptance

- A configured `AuthScheme` reaches the scheme's `description` (and an
  explicitly empty one is stated as no prefix), for echo and fiber, via a
  mapping-declared field rather than per-library code.
- A bearer middleware reading a non-`Authorization` header is documented as an
  apiKey scheme in that header, not as `http`/`bearer`.
- Fixtures for both, and no drift where nothing is configured — the default
  `Bearer` prefix on `Authorization` is what `http`/`bearer` already means, so
  it must add nothing.

Issue #412, opened 2026-08-27
A credential consumed by a security scheme is also documented as a plain header parameter


ehabterra:
When a security scheme is derived from a middleware's header read, that **same read** is also emitted as an ordinary header parameter, so the credential is documented twice — once as the scheme that governs it, once as a parameter the client is told to supply itself.

## Reproduction

`testdata/auth_gin_perroute`:

```yaml
  /users/{id}:
    get:
      operationId: auth-gin-perroute.getUser
      parameters:
        - name: Authorization      # <-- the credential, again
          in: header
          schema:
            type: string
      security:
        - bearerAuth: []           # <-- already governs the Authorization header
```

Both come from the one `c.GetHeader("Authorization")` inside `jwtAuth`.

## Not framework-specific

Two fixtures, two frameworks, three operations:

| fixture | operations |
|---|---|
| `auth_gin_perroute` | `GET /users/{id}`, `POST /users` |
| `auth_nethttp_wrap` | `GET /users/{id}` |

Both wire the guard differently — gin as a sibling argument, net/http as a wrapping call — so this is not a property of either wiring style.

## Why it is wrong

OpenAPI treats the two as different contracts. A `http`/`bearer` scheme already declares that the operation is authenticated via the `Authorization` header; re-declaring it under `parameters` tells a generator to emit a second, manually-supplied `authorization` argument alongside the one the security scheme drives. Generated clients end up with a duplicate argument and generated docs list the credential twice.

## Root cause

Attribution is not the cause — it reproduces both before and after #386/#411, and the read belongs to the middleware either way. The cause is that the parameter matchers walk the route's whole subtree, including sibling and wrapping middleware, and nothing reconciles a matched header read against the security scheme that consumed the same header.

## What a fix has to be careful about

The naive version — drop any parameter named `Authorization` — is exactly the type-name denylist golden rule #9 forbids, and it would be wrong twice over:

1. A handler that genuinely reads `Authorization` for its own purposes, on an operation with **no** security scheme, should keep the parameter. This really happens: on a real project surveyed for this issue, the only `Authorization` parameter in the whole spec was of this shape, and it is correct there.
2. Other schemes name other headers — an `apiKey` scheme's header is configurable (#370), so the reconciliation has to key on *the header the scheme actually consumes*, not on a fixed name.

So the rule should be: suppress a header parameter only when a security scheme **on that same operation** consumes that header. That keeps the honest general case (#7) and needs no name list.

[...]

PR #439, opened 2026-09-02
fix(spec): document an apiKey scheme where the credential travels

### LINKED ISSUE #370: An API-key middleware's configured lookup is ignored: every apiKey scheme is documented as `in: header, name: Authorization`
## Context

`schemeAPIKey` is a constant — `SecurityScheme{Type: "apiKey", In: "header", Name: "Authorization"}`
(`internal/spec/config_security.go:124`) — and every API-key middleware mapping points at it: echo
`middleware.KeyAuth`/`KeyAuthWithConfig` (`config_security.go:148`), fiber `keyauth.New`
(`config_security.go:181`). The middleware's own configuration decides where the key actually travels
(`KeyLookup: "query:api_key"`, `"cookie:token"`, `"header:X-API-Key"`), and that configuration is never
read, so the spec confidently documents the wrong place.

A client generated from this spec sends `Authorization: <key>` to an API that only reads `?api_key=` —
the request fails, and the spec looks authoritative while being wrong.

### Reproduction (echo)

```go
e := echo.New()
g := e.Group("/api")
g.Use(middleware.KeyAuthWithConfig(middleware.KeyAuthConfig{
	KeyLookup: "query:api_key",                     // <- the key is a query parameter
	Validator: func(key string, c echo.Context) (bool, error) { return key == "x", nil },
}))
g.GET("/items", listItems)
```

Generated:

```yaml
      security:
        - apiKeyAuth: []
components:
  securitySchemes:
    apiKeyAuth:
      type: apiKey
      name: Authorization      # wrong
      in: header               # wrong
```

Detection of *which* routes are protected is right; only the scheme's shape is wrong. Fiber's
`keyauth.Config{KeyLookup: …}` has the same shape and the same outcome.

Note the current constant is correct for the *default* configuration (echo's default `KeyLookup` is
`"header:Authorization"`), which is why this has stayed invisible: it is only wrong when a project
configures the lookup — which is the common case for query- or cookie-carried keys.

## Proposed

Config-driven, framework-agnostic (golden rule #5):

1. Let a `SecurityMapping` name a **config argument** and the field within it that carries the lookup
   (e.g. `ConfigArgIndex: 0`, `LookupField: "KeyLookup"`), so the mapping stays data rather than
   per-library code. The composite-literal field resolution this needs already exists in the spec layer
   (`field_resolver.go`, `map_literal.go`).
2. Parse the resolved literal into `in`/`name` using the library's own grammar (`"<source>:<name>"`,
   sources `header`/`query`/`cookie`/`form`), and emit one scheme per distinct shape rather than
   collapsing everything onto a single `apiKeyAuth` key — two groups with different lookups are two
   schemes.
3. Same treatment for a custom bearer/basic header name where the middleware exposes one, and for
   `AuthScheme`-style prefixes.
4. Honest-over-wrong (golden rule #7): when the config value is not statically resolvable (built at
   runtime, read from env), fall back to the library's documented default **and** raise the existing
   unmapped-middleware style quality warning, so `--strict` (#297) can surface it — never silently
   assert a location that was not observed.

## Acceptance

- Fixtures for echo `KeyAuthWithConfig` with `query:` and `cookie:` lookups, fiber `keyauth.New` with a
  custom header name, and one project keeping the default (asserting no drift).
- Structural test on `components.securitySchemes[*].in`/`name`, and on two differently-configured groups
  producing two schemes.
- An unresolvable lookup produces the default plus a warning, and the warning is covered by a test.

ehabterra:
Fixes #370. Also unblocks #412, which had to wait for the scheme to know which
header it consumes.

`schemeAPIKey` was a constant — `{type: apiKey, in: header, name: Authorization}`
— and every API-key middleware mapping pointed at it. That is echo's and
fiber's *default* lookup, so a project that configures one got a spec that reads
as authoritative while telling clients to send the key where the server never
looks:

```go
g.Use(middleware.KeyAuthWithConfig(middleware.KeyAuthConfig{
	KeyLookup: "query:api_key",
	Validator: validate,
}))
```

```diff
   securitySchemes:
     apiKeyAuth:
       type: apiKey
-      name: Authorization
-      in: header
+      name: api_key
+      in: query
```

## Config-driven, not per-library

A `SecurityMapping` declares where the fact lives — `lookupField` (+
`lookupArgIndex`) — the value is read off the composite literal at the call
site, and the grammar echo and fiber share (`"<source>:<name>"`) is parsed into
the scheme's `in`/`name`. The presets declare it for echo's
`KeyAuth`/`KeyAuthWithConfig` and fiber's `keyauth.New`; a house middleware
carrying the same kind of configuration can say so in config, which is the
point of keeping it data (golden rule #5).

## Naming needs the whole project

| what the project does | what is emitted |
|---|---|
| one configured shape | the declared name, corrected — `apiKeyAuth {query, api_key}` |
| several shapes | one scheme each: `apiKeyAuthQueryApiKey`, `apiKeyAuthCookieToken` — **none** keeps the plain name, since handing it to whichever route was walked first would read as the project's single answer and depend on walk order (golden rule #1) |
| a group at the library default | keeps `apiKeyAuth {header, Authorization}` |

That last row is a bug I introduced and then measured: without counting an
*unshaped* reference as one of the shapes, a project with one configured group
and one default group redefined the shared name, and the default group's
credential was documented in the wrong place. It has its own test.

## Honest where it cannot know

"Configured but unreadable" is told apart from "not configured", because only
the first is a fallback that may be wrong:

```console
[security] …middleware.KeyAuthWithConfig: KeyLookup is not statically readable — the scheme keeps the library default, which may be wrong
[security] …middleware.KeyAuthWithConfig: KeyLookup is "form:api_key", which OpenAPI has no apiKey location for — the scheme keeps the library default, which may be wrong
```

A form field is in that second bucket deliberately: OpenAPI has no apiKey
location for one, so inventing `query` would be exactly the confident wrong
answer the issue is about.

## Drift

- **All 105 pre-existing fixtures byte-identical**, the six existing auth
  fixtures included — nothing configured a lookup, so nothing moved.
- **gitea: byte-identical**, 894 paths, and its one `[security]` line is the
  pre-existing unmapped-middleware warning, unchanged.
- Full suite green: goldens, determinism, both engines' parity tests.

## Notes

- `MiddlewareRef` carries the constructor call as a single pointer rather than
  its argument slice: the type is exported and returned to library consumers, so
  a slice field would silently make it non-comparable.
- New fixtures: `testdata/auth_echo_keylookup` (query, cookie and default groups
  side by side) and `testdata/auth_fiber_keylookup` (custom header).
- `lookupField`/`lookupArgIndex` are validated (a lookup needs a scheme to
  shape; an index needs a field) and documented in README + CONFIGURATION.
- Step 3 of the issue — a custom `AuthScheme` prefix, and a bearer token read
  from a non-`Authorization` header — is **not** here: it changes which scheme
  *type* a mapping produces and OpenAPI cannot express a prefix structurally.
  Filed as **#438**.

## Summary by CodeRabbit

- **Bug Fixes**
  - API-key security schemes in generated OpenAPI documentation now reflect middleware configuration, including header, query, and cookie locations.
  - Differently configured authentication routes now receive distinct security schemes.
  - Default behavior remains unchanged when lookup settings are omitted or cannot be resolved.

- **Documentation**
  - Added guidance and examples for configuring API-key lookup locations.
  - Documented runtime-lookup fallback behavior and supported framework defaults.


ehabterra:
Both findings verified against the code before fixing, and both were real. Pushed in 1c1785e.

**1. User-defined scheme precedence** — confirmed. With `securitySchemes.apiKeyAuth` written by hand and two lookup shapes in the code, the routes ended up pointing at derived names while the user’s definition sat in the document unreferenced:

```
emitted schemes = {apiKeyAuth: {header X-User-Chosen}, apiKeyAuthCookieToken: …, apiKeyAuthQueryApiKey: …}
route /a references "apiKeyAuthQueryApiKey" while the user defined apiKeyAuth as {header X-User-Chosen}
```

Fixed by taking the option you listed second: a scheme present in `cfg.SecuritySchemes` is left out of specialisation entirely, so the requirement keeps the declared name and the definition is the user’s. Silence would have hidden a real disagreement, though, so where what they wrote differs from what the middleware reads it is logged once per scheme:

```
[security] securitySchemes.apiKeyAuth is defined as {in: header, name: X-User-Chosen}, but the middleware reads {in: query, name: api_key} — keeping your definition
```

**2. Key collisions** — confirmed. `api_key` and `api-key` both rendered `apiKeyAuthQueryApiKey`, one definition overwrote the other, and both routes referenced it:

```
defs = {apiKeyAuthQueryApiKey: {query api-key}}          # one entry for two shapes
route a -> apiKeyAuthQueryApiKey, route b -> apiKeyAuthQueryApiKey
```

Fixed with the deterministic-suffix option: keys are assigned over the shapes in sorted order, and a collision takes the next free numeric suffix (`apiKeyAuthQueryApiKey`, `apiKeyAuthQueryApiKey2`), so assignment does not depend on map order — the definitions carry the real `name`, so nothing is lost by the suffix being opaque.

Three regression tests, one per property: the user-defined scheme is untouched (your requested case: two shapes, hand-written `apiKeyAuth`), punctuation-only differences get distinct keys pointing at their own definitions, and 20 repeated runs assign the same keys. All nine auth fixtures generate byte-identically to the pre-review commit, so this is a behaviour fix on the new paths only.

PR #441, opened 2026-09-03
fix(spec): stop documenting a credential twice

### LINKED ISSUE #412: A credential consumed by a security scheme is also documented as a plain header parameter

ehabterra:
Fixes #412. Depends on #370 (merged): the rule needs to know which header a
scheme consumes, which was hardcoded to `Authorization` until then.

## The duplication

One `c.GetHeader("Authorization")` inside a middleware produced both halves:

```yaml
  /users/{id}:
    get:
      parameters:
        - name: Authorization      # the credential, again
          in: header
      security:
        - bearerAuth: []           # already governs the Authorization header
```

OpenAPI treats those as different contracts, so a generated client grows a
second, manually-supplied argument beside the one the scheme drives.

## The rule, and what it deliberately keeps

A header parameter is dropped when a security scheme **on that same operation**
consumes that header — `http`/`oauth2`/`openIdConnect` consume `Authorization`
(the type implies it), an `apiKey` consumes the header it names.

`testdata/auth_credential_param` is four routes, and three of them keep their
parameter:

| route | scheme | header params |
|---|---|---|
| `/basic/items` | `http`/basic | ~~`Authorization`~~ — the scheme's job |
| `/header/items` | apiKey in `X-API-Key` | `X-Request-Id` — the unrelated header survives |
| `/query/items` | apiKey in **query** | `Authorization` — a different credential |
| `/open` | none | `Authorization` — the handler's own business |

Those three are the point. A name denylist would have satisfied the issue's
headline and broken all of them — and the issue notes that on a real project
surveyed for it, the *only* `Authorization` parameter in the whole spec was of
the `/open` shape and correct there (golden rule #9).

Also handled, with unit tests:

- **case-insensitive** header matching — the middleware and the handler are
  written by different hands;
- **document-level** security governs an operation that inherits it;
- an **explicitly public** operation (`security: []`) overrides that and keeps
  its header;
- a requirement naming a scheme with **no definition** consumes nothing —
  suppressing on a guess is how a real parameter disappears;
- a **verb split** does not lose its siblings' parameters: a dispatch split
  copies the RouteInfo, so the operations share one parameter slice and
  filtering in place would edit both. The slice is rebuilt instead.

## Drift

- the two fixtures the issue names lose **exactly** their duplicated
  `Authorization` parameter block, nothing else;
- **all 105 other fixtures byte-identical**;
- **gitea byte-identical** at 894 paths, with zero parameters removed — its auth
  middleware maps to no scheme, so no operation there is governed and the pass
  is inert. Worth stating plainly: gitea confirms no collateral damage but
  offers no positive evidence, which is why the fixture carries all four
  answers;
- coverage 96.0% (floor 96).

## Note

The pass runs after apiKey specialisation, since that is where the header an
apiKey scheme consumes is finally settled (#370). Testing it needs the ENGINE
rather than a single framework config: the header reads in the fixture are
matched by the `net/http` patterns that detection layers under the detected
framework, and echo's own config carries none — with echo's config alone the
spec has no header parameters at all, so the test would pass while proving
nothing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented authentication credentials from appearing twice in generated API documentation.
  * Header parameters consumed by HTTP, OAuth2, OpenID Connect, or header-based API key security schemes are now omitted.
  * Preserved unrelated headers and credentials supplied through query or cookie parameters.
  * Header matching is case-insensitive, with public and unsecured operations retaining their existing behavior.

* **Documentation**
  * Updated the README and changelog to describe the credential parameter behavior.


ehabterra:
Both findings verified before fixing, and both were real. Pushed in 94aa576.

**1. Alternatives were unioned** — confirmed:

```
reqs = [{bearerAuth: []}, {queryKey: []}]  ->  consumed = {authorization: true}
```

A client authenticating with the query key alone never sends an Authorization credential, so dropping that parameter removes a header the handler can still require — the same failure this pass exists to prevent, in the other direction. Now intersected as you suggested: the headers of one requirement **object** still accumulate (its schemes are required together), across objects only the headers every alternative consumes survive, and an alternative that consumes nothing empties the whole answer. Three cases added to `TestSchemeConsumedHeaders` (intersection, all-alternatives-consume, AND-inside-an-OR) plus a route-level case.

**2. The filtering catalog disagreed with the emitted one** — confirmed. Where a user had written the name discovery also produced:

```
filtering = {Type:apiKey In:header Name:X-API-Key}     # what the filter judged by
emitted   = {Type:apiKey In:query  Name:api_key}       # what the document publishes
```

so a header parameter could be dropped because of a definition the spec never contains. Rather than passing the user map down to `mergedSchemes`, there is now a single `effectiveSchemes(cfg, discovered)` carrying reconcile’s own precedence — preset, specialised by discovery, overridden by anything the user wrote — and `TestEffectiveSchemesPrecedence` asserts the filter and the document agree per scheme, so the two call sites cannot drift apart again.

Behaviour double-checked after both fixes: the two fixtures from the issue and all four routes of `auth_credential_param` produce exactly what they did before, so these are corrections on the new paths only.

Issue #356, opened 2026-08-16
Route matchers other than the HTTP method are ignored: query/header matchers vanish, host matchers corrupt the path


ehabterra:
## Context

Route registration is matched for path, handler and method only. Every other matcher a router supports —
required query pair, required header, host, scheme — is either dropped silently or, when the host is part
of the pattern string, folded into the path and emitted as an invalid path template.

### Reproduction (gorilla/mux)

```go
r.HandleFunc("/items", list).Methods("GET").Queries("page", "{page:[0-9]+}")
r.HandleFunc("/items/{id:[0-9]+}", get).Methods("GET").
    Headers("X-Tenant", "").Schemes("https").Host("api.example.com")
```

Generated:

```yaml
  /items:
    get:
      operationId: probe_mux.list        # no `page` query parameter
  /items/{id}:
    get:
      parameters:
        - name: id
          in: path
          schema: {type: string, pattern: '[0-9]+'}
                                          # no `X-Tenant` header parameter
                                          # no servers entry for https://api.example.com
```

`.Queries()` states a **required** query parameter (with a regex constraint, which the path branch already
knows how to turn into `pattern`). `.Headers()` states a **required** header parameter. `.Host()` +
`.Schemes()` state a server URL. All three are dropped.

### Reproduction (net/http, Go 1.22 patterns) — worse than dropped

```go
mux.HandleFunc("GET api.example.com/items", handler)
```

Generated:

```yaml
  /api.example.com/items:      # host folded into the path — an endpoint that does not exist
```

The host component of a Go 1.22 `ServeMux` pattern is parsed as if it were the first path segment. Any
project using host-based routing on stdlib `net/http` gets wrong paths, not just missing metadata.

`config_mux.go` matches `^Methods$` and nothing else; `Servers` exists only as a hand-written config field
(`internal/spec/config.go:834`) and is never inferred from code.

## Proposed

Extend the pattern configs (golden rule #5 — this must not be a mux-only branch) so a registration chain
can contribute more than a path/method/handler:

1. **Query matchers → `in: query` parameters, `required: true`.**
   mux `.Queries(k, v)` (pairs; `{name:regex}` values → `pattern`); echo/gin/chi have no direct
   equivalent — the config must simply express nothing for them.
2. **Header matchers → `in: header` parameters, `required: true`.**
   mux `.Headers(k, v)` / `.HeadersRegexp(k, re)`.
[...]

PR #474, opened 2026-09-09
fix(spec): a ServeMux pattern's host is not part of the path

ehabterra:
Addresses **part 4** of #356 — the half that issue marks as *"a correctness bug [that] can land on its own"*. Parts 1–3 (query/header matchers, `servers` from host+scheme) stay open there.

## The defect

A Go 1.22 ServeMux pattern is `[METHOD ][HOST]/[PATH]`. Only the method was split off:

```go
mux.HandleFunc("GET api.example.com/items", listItems)
```
```yaml
  /api.example.com/items:      # an endpoint that does not exist
```

Worse than a missing route — nothing in the document says it's wrong, so a consumer builds a request against a path that can never match. `splitMethodFromPath`'s own doc comment already named the grammar; the host half was never implemented.

## Declared, not detected

Go treats **any** text before the first `/` as the host, so `mux.Handle("items/x", h)` really does register host `items`. That rule cannot be applied here: a path reaching the matcher has already been through resolution and may have lost its leading slash (a concatenated prefix, a mount join), so `v1/users` would lose a real segment. Deciding by the *shape* of the text — "dotted, or localhost, with an optional port" — is the same guess wearing a regex, and its failure mode is silent.

So the host set is configured:

```yaml
hosts:
  - api.example.com
  - localhost:8080
```

Hosts already named in `servers` count too, because `servers: [{url: https://api.example.com}]` has already said what the project's hosts are. Comparison is case-insensitive (RFC 4343); a different port is a different host.

**A project that declares nothing keeps today's output exactly**, `/api.example.com/items` included. The invalid path is fixed when the project says which hosts are its own — a decision rather than an accident, pinned by a test asserting the unchanged path.

## What this does not fix, pinned rather than hidden

Two hosts serving the same path now collapse into **one** operation — the path is the whole key and the host has nowhere to go yet:

```go
mux.HandleFunc("GET api.example.com/status", health)
mux.HandleFunc("GET admin.example.com/status", listItems)   // → one /status
```

Previously both were documented, as two paths that were both invalid. The fixture covers this with a comment saying the assertion changes when `servers` (part 3 of #356) lands. The host is **returned** by `splitHostFromPath` rather than discarded, so that work has the fact to hand.

## Verification

- `testdata/http_host_pattern` + structural tests, verified to fail on `main`:
  `path "/items" missing; have [/api.example.com/items /api.example.com/items/new /localhost:8080/debug …]`
- `TestSplitHostFromPath` leads with the negative cases: an undeclared but host-shaped prefix, a different port, `v1/users`, `{tenant}/items`
- The fixture runs three ways: hosts declared, host named only by a server URL, and nothing declared
- Fixture A/B: **118 of 118 identical** — with nothing declared, no existing output changes at all
- gitea: **930 paths, nothing lost or gained** — it registers no host patterns, which is why this went unnoticed

Suite, `make lint`, coverage ratchet (96.0%) pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected route generation for host-based HTTP patterns so hostnames no longer appear as part of URL paths.
  * Consolidated routes served by multiple hosts into a single operation when their paths and methods match.

* **Tests**
  * Added coverage for host and path handling, including ports, localhost, hostless paths, and unresolved patterns.

The new config field is wired into the apispecui editor (`TestConfigEditorCoversEveryFrameworkSection` enforces it).

### DIFF
--- a/cmd/apispecui/assets/js/config.js
+++ b/cmd/apispecui/assets/js/config.js
@@ -349,6 +349,18 @@ export function ConfigMode() {
             <button class="btn secondary sm" onClick=${() => addTo("servers", { url: "", description: "" })}>+ Add server</button>
           <//>
 
+          <${Section} title="Virtual hosts" help="Host names your routes are registered under, for routers whose registration pattern can carry one — a Go 1.22 ServeMux pattern is '[METHOD ][HOST]/[PATH]', so mux.HandleFunc(\"GET api.example.com/items\", h) serves /items on that host. Listing a host here documents the route at /items instead of /api.example.com/items. Hosts are listed rather than detected because a pattern that lost its leading slash is indistinguishable from one carrying a host. Hosts already named in Servers count too." hint=${`${(c.hosts || []).length}`}>
+            ${(c.hosts || []).map(
+              (h, i) => html`
+                <div class="row">
+                  ${txt("", h, (e) => setKey("hosts", (c.hosts || []).map((v, j) => (j === i ? e.target.value : v))), "api.example.com")}
+                  <span class="spacer"></span>${RowDelete(() => setKey("hosts", (c.hosts || []).filter((_, j) => j !== i)))}
+                </div>
+              `,
+            )}
+            <button class="btn secondary sm" onClick=${() => addTo("hosts", "")}>+ Add host</button>
+          <//>
+
           <${Section} title="Tags" help="Named groups that organise operations into sections in the docs UI. Routes are usually tagged automatically from their mount/group prefix; add tags here to give them an order and a description. Example: name 'Users', description 'Account and profile endpoints'." hint=${`${(c.tags || []).length}`}>
             ${(c.tags || []).map(
               (t, i) => html`
--- a/generator/testdata_http_host_pattern_test.go
+++ b/generator/testdata_http_host_pattern_test.go
@@ -1,0 +1,67 @@
+// Copyright 2026 Ehab Terra
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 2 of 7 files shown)


ehabterra:
Reworked to your call: **only hosts the config names, no guess** (commit `be3bbe0`).

## What changed

The first version decided by shape — a dotted name or `localhost`, optionally with a port. Narrower than Go's own rule, but still a guess about text, and the failure mode is silent and destructive: a path that reached the matcher having lost its leading slash and happened to start with something dotted would lose a real segment, with nothing in the output saying so.

Now `APISpecConfig` grows `hosts`, and only a host named there is split off:

```yaml
hosts:
  - api.example.com
  - localhost:8080
```

Hosts already named in `servers` count too — declaring `servers: [{url: https://api.example.com}]` has said what the project's hosts are, and requiring the same name twice would be a trap rather than a safeguard. Comparison is case-insensitive (RFC 4343); a different port is a different host.

## The consequence, stated plainly

A project that declares nothing keeps today's output exactly, `/api.example.com/items` included. The invalid path is no longer fixed by default — it is fixed when the project says which hosts are its own. That is now a decision rather than an accident, and there is a test asserting the unchanged path, alongside the ones asserting the split when a host is declared and when only a server URL names it.

## Measurements

- Fixture A/B: **118 of 118 identical** — with no host declared anywhere, this changes no existing output at all (the earlier version changed one fixture).
- `TestSplitHostFromPath` now leads with the negative cases: an undeclared but host-shaped prefix, a different port, `v1/users`, `{tenant}/items`.
- New config field wired into the apispecui editor (`TestConfigEditorCoversEveryFrameworkSection` enforces it).

Suite, lint and the ratchet (96.0%) pass.

PR #496, opened 2026-09-16
perf(spec): memoize the response-call matcher on the callee name handle

ehabterra:
## The find

Profiling v0.5.9-to-be on gitea, after #491 cleared the 42%-of-allocation offender:

```
      flat  flat%   cum   cum%
     4.27s  1.76%  13.96s  5.77%  regexp.(*Regexp).doOnePass
     0.07s 0.029%  17.20s  7.11%  regexp.(*Regexp).MatchString
```

and every bit of it has one source:

```
$ go tool pprof -peek 'regexp\.\(\*Regexp\)\.MatchString'
                                            17.09s 99.36% |   spec.ResponseCallMatcher.func2 (inline)
                                             0.08s  0.47% |   spec.registrationMatcher.func3 (inline)
                                             0.03s  0.17% |   spec.cachedMatch (inline)
```

`LazyTree.isResponseCall` calls this **per child spec** while materialising the tree — 15.7M nodes on gitea — to decide which instance budget applies. Those nodes carry a few thousand distinct callee names, so `"Render"` was re-matched hundreds of thousands of times. `registrationMatcher` next to it goes through `cachedMatch`; this one went straight to `regexp`.

This is the regression #224's response budget introduced. It does not exist in v0.5.8, which had no response-call matcher.

## The change

Memoize on `edge.Callee.Name` — the string-pool handle, which is the closure's **whole** input: the compiled patterns are fixed when it is built, and the name is all it reads.

A plain map, not `cachedMatch`: the walk is single-threaded and the closure is built per tree, so there is nothing to share it with. (`cachedMatch` takes a lock because it is a package-level cache keyed by pattern text; this is neither.)

## Measurements

gitea (374 packages, 931 paths), same machine, two runs of each binary:

| | main | this | delta |
|---|---|---|---|
| wall | 291.4s | **159.6s** | **-45%** |
| instructions retired | 1.302T | **0.620T** | **-52%** |
| spec mapped | 4m22s | **2m14s** | **-49%** |

The gain is much larger than the 7.1% the profile attributes to `MatchString`, because each execution allocates and frees a `onePassMachine`: removing 15.7M of those removes the GC work too, which was charged to `scanObject`/`madvise` rather than to the frame that caused it. (GC pacing is also why the two runs of each binary vary — 187.5s/159.6s here, and instructions 0.752T/0.620T. Both are far outside that band.)

**In context**: v0.5.8 ran this project in **149.9s**. `main` today takes **291.4s**. This brings it to **159.6s** while keeping the 32 extra paths the release gained — essentially the whole regression, from one missing cache.

**Output is byte-identical** to main (`cmp` on the 931-path spec).

## The test

`TestResponseCallMatcherIsDecidedByTheCalleeNameAlone` pins the premise the memo rests on.

The hazard is specific: if an edit ever makes the decision read the receiver or the package as well, the memo answers from whichever edge asked **first**, and every later edge sharing that name silently inherits it. Asking one matcher cannot detect that — the memo is exactly what hides it. My first version of this test did that and passed against a deliberately broken matcher.

So it asks **two fresh matchers** the same pair of edges in opposite orders. Confirmed to fail on the shape it guards:

```
--- FAIL: TestResponseCallMatcherIsDecidedByTheCalleeNameAlone
    order changed the answer (gin true then false, echo false then true): the
    decision reads more than the callee name, which the memo cannot key on
```

## Verification

- `go test ./...` — all packages pass
- `make lint` — 0 issues
- `scripts/check-coverage.sh` — 96.0%, floor 96
- gitea output byte-identical to main

Independent of #491; both are in the same walk.

https://claude.ai/code/session_01GGyBc4ucpCvWbpyP8rpqa5

## Summary by CodeRabbit

* **Performance**
  * Improved repeated response-call matching by reusing previously determined results for the same callee name.

* **Bug Fixes**
  * Ensured matching decisions remain consistent regardless of receiver, package, or call order.
  * Preserved expected behavior for unmatched names, empty names, and nil edges.


ehabterra:
@coderabbitai review

PR #508, opened 2026-09-18
test: pin the assignment-matching limit CodeRabbit found

ehabterra:
Verified both shapes it named. Assignments are collected by NAME, so a shadowed binding (`prefix := "/outer"` and an inner block's `prefix := "/inner"`) and a write below the registration (`p = "/second"` after it) each read as two disagreeing values, and a knowable path keeps its placeholder:

    /{shadowed}/a   should be /outer/a
    /{shadowed}/b   should be /inner/b
    /{later}/c      should be /first/c

Precision, not correctness: the failure mode is an approximate path with a flagged synthesized parameter, never a wrong one.

Only half of it is implementable today. Reachability can use recorded facts — Assignment.Position plus Function.Blocks/blockIndex.dominates, since a plain source-order filter would break `for { p = "/x"; register(p) }`. The lexical half cannot: Assignment.Scope is exported/unexported, not a scope, and the assignment token is not recorded, so `:=` and `=` are indistinguishable. Filed as #436 with both halves rather than approximated with block ranges, which would trade this false negative for a false positive.

The three shapes join testdata/variable_path with change-detector assertions, and the limit is recorded on pathVarAssignments where the rule lives.

### DIFF
--- a/generator/testdata_variable_path_test.go
+++ b/generator/testdata_variable_path_test.go
@@ -46,16 +46,33 @@ func TestTestdata_VariablePath(t *testing.T) {
 		}
 	}
 
-	// Nothing may be documented under a variable's name or type: those are the
-	// two wrong answers this replaces.
+	// Nothing may be documented under a variable's TYPE: that is the wrong
+	// answer this replaces.
 	for path := range out.Paths {
-		if strings.Contains(path, "{") {
-			t.Errorf("path %q keeps a placeholder — every variable here resolves", path)
-		}
 		if path == "/string" || strings.HasPrefix(path, "/string/") {
 			t.Errorf("path %q is the variable's TYPE, not its value", path)
 		}
 	}
+
+	// CHANGE DETECTOR, not an endorsement (issue #436): assignments are matched
+	// by NAME, so a shadowed binding and a write after the registration each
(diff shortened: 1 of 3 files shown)


ehabterra:
Closing as superseded — `main` has moved past this and carries strictly more.

This PR recorded the limit and added the three shapes as change-detectors. Since then the **reachability half actually landed** on main, so merging it now would regress the test rather than add to it:

| | this PR | current main |
|---|---|---|
| `testdata/variable_path` shapes | added | present |
| `/first/c`, `/outer/a` | asserted as NOT resolving | **resolve**, asserted as such |
| `/inner/b` | asserted as not resolving | still a change-detector, with the metadata reason |
| the rule itself | comment only | `assignmentsReaching` + `blockIndex.dominates`, permissive on missing facts so a loop's back-edge write still blocks |

That is exactly the split #436 proposed — reachability implementable from recorded facts, the lexical binding blocked on a metadata fact — and main implements the first half rather than only documenting it.

Merging was attempted and conflicts in `generator/testdata_variable_path_test.go`; resolving it would mean taking main's side wholesale. The branch is also far behind (215 files), so there is nothing here to salvage.

#436 stays open for the lexical-binding half, which is still blocked on `Assignment` recording the token (`:=` vs `=`) or the binding each write targets.

Issue #471, opened 2026-09-09
An encode that reaches the response through a buffer is invisible to the write-destination gate


ehabterra:
Split out of #354 with measurements, because it decides how much of the encoder half can be trusted.

## What

`RequireResponseDestination` vets an encode by resolving its destination and dropping the encode when that destination provably is not the response writer (`extractor.go`, `destination()` / `ShouldDrop`). It sees one hop. A handler that encodes into a **buffer** and then writes the buffer to the response therefore looks identical to one that encodes into a buffer and throws it away:

```go
// modules/sitemap/sitemap.go — the buffer IS the response
func (s *Sitemap) WriteTo(w io.Writer) (int64, error) {
	buf := bytes.NewBufferString(xml.Header)
	encoder := xml.NewEncoder(buf)
	encoder.Encode(s)      // destination: buf
	...
	return buf.WriteTo(w)  // …which is then written to w
}

// modules/actions/jobparser/model.go — the buffer is internal
func (w *SingleWorkflow) Marshal() ([]byte, error) {
	var buf bytes.Buffer
	enc := yaml.NewEncoder(&buf)
	enc.Encode(w)          // destination: buf, and it never goes near a writer
	return buf.Bytes(), nil
}
```

Both are `Encode` into a `*bytes.Buffer`. The first is a response body and the second is not, and nothing in the gate can tell them apart.

## Why it matters now

It sets the accuracy ceiling for the non-JSON encoder patterns added in #354. Measured on a large real project (930 paths), running those patterns with the **permissive** default — keep an encode whose destination cannot be resolved — produced:

- **18 correct** `application/xml` responses (the sitemap handlers above), and
- **8 false** `application/yaml` responses on catch-all routes, every one of them a workflow serializer or migration dumper writing to a buffer or a file.

Every yaml encode reached from a route on that project wrote away from the wire. Not one wrote a response.

So #354 ships those patterns with `DropUnresolvedDestination`, which trades the 18 right answers for zero wrong ones. The sitemap XML is the concrete thing that costs, and it is exactly what this issue would recover.

## The gap

The gate needs the same treatment `unwrapWriteSink` gives `json.Marshal` (issue #195): follow the buffer forward to a sink. Two shapes cover most of it:

1. `buf.WriteTo(w)` / `io.Copy(w, buf)` / `w.Write(buf.Bytes())` in the same function — the buffer reaches a writer.
[...]

PR #510, opened 2026-09-18
feat(spec): an encode into a buffer that is flushed to the writer is a response

ehabterra:
Implements the mechanism #471 asks for. Does **not** yet recover the gitea endpoints it was filed for, and the fixture says why.

## The gap

The write-destination gate resolves a destination **backwards** — assignments, parameters, struct construction — which answers *"where did this value come from"*. A buffer's answer is always "a buffer", so these are indistinguishable and both were dropped:

```go
xml.NewEncoder(buf).Encode(v); buf.WriteTo(w)   // the buffer IS the response
yaml.NewEncoder(&buf).Encode(v); return buf     // the buffer is internal
```

The question they differ on is **forward**: does anything take this buffer's bytes to a writer?

## The change

`ResponseContext.BufferSinks` names the calls that do — config-driven like the `BodyTransforms` beside them (golden rule #5), and serializer-level, so every net/http-family framework shares them:

| call | buffer | writer |
|---|---|---|
| `buf.WriteTo(w)` | receiver | arg 0 |
| `io.Copy(w, buf)` / `io.CopyBuffer(w, buf, scratch)` | arg 1 | arg 0 |

Scoped to the function the encode is written in — a buffer handed elsewhere is that function's business. **Order is not checked**: a flush above the encode is the same buffer reaching the same writer, and requiring dominance would drop the `if err != nil { return }` shape every one of these handlers is written in.

The flush target is judged by the rule the gate already applies one level out: provenance to the response writer, **or** a type that could still be it. That second half matters — a helper's writer is an `io.Writer` parameter, and an encode written straight into one is already kept, so refusing it after a buffer hop would be the gate disagreeing with itself.

It stays honest because a resolved **non**-writer still fails: `buf.WriteTo(otherBuf)` is a `*bytes.Buffer`, `buf.WriteTo(f)` an `*os.File`. That is exactly what #471 measured — every yaml encode reached from a route on that project wrote to a buffer or a file, and none to the wire.

## Fixture

`testdata/encode_buffer_sink`, six shapes, **confirmed to fail on four without this change**:

| shape | documented |
|---|---|
| `buf.WriteTo(w)` | ✅ `application/xml` |
| `io.Copy(w, buf)` | ✅ |
| encoder assigned to a variable first | ✅ |
| encode + flush inside a helper **function** | ✅ |
| encode + flush inside a **method** on the payload type | ❌ — see below |
| buffer never written anywhere | ✅ correctly **not** documented |

The last row is the half that makes the gate worth having: documenting it would claim a body the endpoint never sends.

## What this does not recover, and why

**gitea's 18 sitemap responses.** Their encode lives in a method on the payload type:

```go
func (s *Sitemap) WriteTo(w io.Writer) (int64, error) {
	buf := bytes.NewBufferString(xml.Header)
	encoder := xml.NewEncoder(buf)
	encoder.Encode(s)
	return buf.WriteTo(w)
}
```

Instrumenting the gate shows the encode **never reaches it** — there is no drop decision for that function at all, so the walk is not matching the encode in the first place. A plain helper *function* with an identical body does resolve (`/via-func`), so the gap is the **method hop**, not the extra frame.

gitea and a 549-path service are therefore **byte-identical** under this change. I would rather ship the mechanism with that stated than imply a payoff it does not deliver; `/via-helper` is asserted as currently-failing so it flips loudly when the method hop is fixed, and #471 stays open for it.

## Verification

- `go test ./...` — 31/31 packages
- `make lint` — 0 issues
- `scripts/check-coverage.sh` — 96.0%, floor 96
- gitea and a 549-path service byte-identical

`bufferSinks` joins `bodyTransforms` in the apispecui parity exemption, for the same stated reason: a nested struct list, edited in YAML mode.

https://claude.ai/code/session_01GGyBc4ucpCvWbpyP8rpqa5

## Summary by CodeRabbit

- **New Features**
  - Improved response analysis for data encoded into buffers and later written to HTTP responses.
  - Added support for common buffer-to-writer patterns, including `WriteTo` and copy operations.

- **Bug Fixes**
  - Prevented valid XML responses from being incorrectly discarded when content passes through an intermediate buffer.

- **Tests**
  - Added coverage for buffer-based response handling across multiple encoding scenarios.


ehabterra:
@coderabbitai review

PR #522, opened 2026-09-20
fix(spec): derive a request body only from a decode of the request

### LINKED ISSUE #513: An outbound response decode inside a method is auto-detected as a request-body wrapper, and replaces the handler's real request body
From a downstream consumer's output-quality review (chi, ~720 routes), reduced to a minimal module and reproduced on `main` with the default config. It was the only *false* type in their spec — everything else wrong was merely loose — and nothing on stderr mentions it.

## What

A method that decodes an **outbound** HTTP response into an `any` parameter is auto-detected as a request-body wrapper, and the type its caller passes is documented as the handler's request body.

```go
type providerReply struct {
	Contacts []string `json:"contacts"`
}

type client struct{}

func (c *client) fetch(out any) error {
	resp, err := http.Get("https://provider.example")
	if err != nil {
		return err
	}
	defer resp.Body.Close()
	return json.NewDecoder(resp.Body).Decode(out) // the PROVIDER's response
}

var c = &client{}

func h(w http.ResponseWriter, r *http.Request) { // reads no body at all
	var out providerReply
	_ = c.fetch(&out)
	w.WriteHeader(204)
}
// r.Put("/x", h)
```

### Actual

```yaml
  /x:
    put:
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/example_com_repro_providerReply'
        required: true
```

### Expected

No `requestBody`. Bytes read from an `http.Response` are never a request body.

## Root cause

The source gate is fine — instrumented, `RequireRequestSource` evaluates the inner `Decode` and correctly answers **false** (`src=resp.Body`, not a `net/http.Request` accessor). The body comes from somewhere else: `deriveRequestWrapper` (`internal/spec/wrapper_detect.go`).

Wrapper detection decides on `m.MatchNode(node)` alone. Body-source verification deliberately lives in `ExtractRequest`, not `MatchNode` (it is per-route, and `MatchNode` is memoized per edge) — so detection never consults it, and the pattern it derives does not carry it either. `--output-config` shows the result:

```yaml
    requestBodyPatterns:
        - callRegex: ^fetch$
          recvTypeRegex: ^example\.com/repro\.\*?client$
          typeFromArg: true
          deref: true          # no requireRequestSource, no source to check
```

That derived pattern then matches `c.fetch(&out)` in the handler, ungated.

Two properties that hide it from fixtures:

- **Methods only.** The same helper as a plain function (`func fetch(out any) error`) is not a wrapper candidate and documents nothing — correct. Give it a receiver and it flips.
- **It replaces, not adds.** When the handler *also* decodes its own body (`json.NewDecoder(r.Body).Decode(&in)` directly or via a `decodeJSON(r, &in)` helper), the outbound type still wins: both are concrete `$ref`s, and on a tie `preferRequestInfo` is last-write-wins — a handler decodes its body first and calls the service afterwards, so the outbound type is the later one. In the consumer's case a `PUT` that plainly decodes `Settings` was documented as taking the accounting provider's `contactsResponse`.

Same result with `data, _ := io.ReadAll(resp.Body); json.Unmarshal(data, out)`.

## Suggested fix

1. `deriveRequestWrapper` should only derive from an inner decode whose source **is shown to be** a request source — run the body-source resolver on the inner edge at detection time (the wrapper's own `c.r.Body` / a `*http.Request` parameter both trace), and derive nothing otherwise. A wrapper whose source is one of its own parameters should carry that forward as `RequireRequestSource` + `BodySourceArgIndex` rather than dropping the gate.
2. Independently, precedence: a body found by a source-verified pattern should not be displaced by one from a pattern that verified nothing. Today `preferRequestInfo` ranks by schema specificity only and breaks ties by arrival order.

## Acceptance

- `testdata/outbound_decode_wrapper`: the method shape above beside a genuine `func (c *Ctx) Bind(dst any)` wrapper over `c.r.Body`, in one fixture — the first documents no body, the second still does. Include a handler that does both (own decode + outbound call) asserting its own type survives.
- Probe the same shape under gin/echo/fiber wrappers, since derivation is shared across frameworks (golden rule #5).

ehabterra:
Fixes #513 — the only *false* type found in a downstream consumer's 723-operation spec, and silent.

## The bug

A decoder wrapper is recognised by shape — a method forwarding its own parameter into a decode — and an outbound HTTP client has that shape too:

```go
func (c *Ctx) Bind(dst any) error     { return json.NewDecoder(c.Req.Body).Decode(dst) }
func (c *client) fetch(out any) error { return json.NewDecoder(resp.Body).Decode(out) }
```

Only where the bytes come from tells them apart, which is exactly what a request pattern's `requireRequestSource` asks at extraction time. **Derivation never asked it** — it decided on `MatchNode` alone — so the second derived a pattern as readily as the first:

```yaml
- callRegex: ^fetch$
  recvTypeRegex: ^example\.com/repro\.\*?client$
  typeFromArg: true
  deref: true            # no requireRequestSource, nothing to check
```

Every handler calling it then documented the provider's type, and *replaced* the handler's own body rather than appearing beside it — both are concrete, and on a tie `preferRequestInfo` is last-write-wins, so the outbound call (which comes after the handler decodes) wins. Two properties kept it out of the fixtures: it needs a **method** (the same helper as a plain function is not a wrapper candidate), and it is invisible unless a project also has an HTTP client.

## The fix

Derivation resolves the inner decode's source and answers three ways:

| the method reads | derived |
|---|---|
| the request itself (`c.Req.Body`) | as before, with nothing left to check |
| whatever it is **handed** (`func (c *Ctx) readFrom(src io.Reader, dst any)`) | with `requireRequestSource` carried to the call site |
| anything else | nothing at all |

The middle row is why the check could not simply be moved wholesale: a wrapper's source is often knowable only from its caller. In the fixture the same helper documents a body when handed `r.Body` and none when handed a file.

## Three gaps this had to close first

Each was wrong on its own, and each stayed hidden because derivation did not consult the check:

1. **The chain walk stopped at the root.** A house context holds the request in a field, so `c.Req.Body` has a root typed `*Ctx` — and the answer was "reads nothing from the request" for every project that owns a context type. Every prefix is now offered its turn, driven by the configured `requestContext` alone, so a house context wrapping gin's, echo's or fiber's is covered by construction rather than by a branch per framework (golden rule #5).
2. **A method's own assignments were invisible.** The canonical call-site lookup (#182) reached a method scope only through `ParentFunction`, i.e. only for a closure declared inside one, because methods live in `Type.Methods` rather than the file's function table. So for a call written directly in a method body, every local assigned there was invisible and each resolver fell back to its "cannot tell" answer.
3. **A body read into bytes could not be traced past the read.** `data, _ := io.ReadAll(r.Body)` then `json.Unmarshal(data, &v)` answered "not the request" exactly as an outbound response does — conflating *provably not* with *cannot see* — so such a handler documented **no body at all**, with no wrapper involved. New `requestContext.bodyReaders` names the calls that turn a reader into bytes, the mirror of `responseContext.bodyTransforms` on the way out. Serializer-level, so every framework shares them.

`bodyReaders` is **named rather than walked** deliberately. Recursing into every call an argument came from ran a full origin trace per argument and cost **10% of a run** on gitea (`IsRequestSource` 0 → 9.7% of samples, `traceVariableOriginHelper` 5.98% → 13.80%); naming the readers makes it a regex check on a field the assignment already records.

## Measured

**gitea** (988 paths), before → after:

| | base | fix |
|---|---|---|
| paths / operations / responses / components | 988 / 1243 / 2495 / 246 | **identical** |
| request bodies | 148 | 149 |
| runtime (2 runs) | 196s, 165s | 169s, 165s |
| peak RSS | 4.47 GB | 4.37 GB |

No request body lost. The one gained is `POST /restore_repo`, which reads its body with `io.ReadAll(ctx.Req.Body)` and had never been documented — a real miss, not drift.

**All 130 `testdata/` fixtures are byte-identical**, so nothing else moved.

## Coverage

`testdata/outbound_decode_wrapper` puts all eight shapes side by side; 5 of the 8 subtests fail on `main`:

| route | shape | before | after |
|---|---|---|---|
| `POST /settings` | house `Bind` over `c.Req.Body` | Settings | Settings |
| `POST /contacts/sync` | outbound decoder chain | **providerReply** | none |
| `PUT /settings` | own decode **and** an outbound one | **providerReply** | Settings |
| `POST /items` | carried source, given `r.Body` | Item | Item |
| `POST /items/import` | carried source, given a file | **Item** | none |
| `PATCH /settings` | plain `io.ReadAll` + `Unmarshal` | **none** | Settings |
| `POST /settings/replace` | the same read, through the wrapper | Settings | Settings |
| `POST /items/remote` | the same read, from the provider | **providerReply** | none |

Plus unit tests on the chain walk across every shipped framework preset, and one that **records a limit rather than a guarantee**: the check asks whether a value is an `*http.Request`, never *whose*, so a request the handler built to send is accepted like the one it is serving. That was already true at the root before this change; the walk only makes it reachable one accessor deeper via the `Request` field on an `*http.Response`. Closing it means running provenance to the handler's own parameter, which #513 describes and this does not attempt.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved request-body detection when data is read through `io.ReadAll`, `ioutil.ReadAll`, or application context fields.
  - Prevented outbound HTTP response bodies from being incorrectly documented as incoming request bodies.
  - Improved detection of request bodies passed through local variables, wrappers, and nested request chains.
  - Added support for identifying request bodies across Gin, Echo, Fiber, and net/http applications.
  - Prevented internal provider-response components from appearing in generated API specifications.


ehabterra:
Addressed the review.

**`resp.Request.Body` (Major)** — confirmed and fixed, and you were right that it was a widening this branch introduced: it answered `false` before only because the walk stopped at the root.

The fix is in the shared check, not the wrapper gate. What separates the two chains is where the root came from — a parameter or receiver is something the handler was *given*, while `resp` is a local it built from `http.Get`:

```go
c.Req.Body         // c is the receiver — the context holds our request
resp.Request.Body  // resp came from http.Get — the request we SENT
```

So the walk past the root is offered only to the former. One assignment lookup, nothing new. Verified end to end on a throwaway module: a handler whose only decode is `json.NewDecoder(resp.Request.Body)` documented `providerReply` before and documents no body now, and it is in the fixture as `POST /items/echo`.

I did **not** claim the root case as fixed. An outbound request assigned to a local is still accepted there, exactly as on `main` — that is what makes `req := r; req.Body` work — and the test records it as a limit rather than a guarantee. Closing it needs provenance run to the handler's own parameter, which #513 describes in full.

**Negative `SourceArgIndex` (Minor)** — valid, fixed. While checking it I found `ResponseContext.BodyTransforms` has the identical shape and the identical hole (bounded from above only), predating this branch; guarded alongside rather than left as a landmine I'd just found.

**Framework coverage (Minor)** — added `testdata/outbound_decode_wrapper_gin` with the three end-to-end cases you asked for: genuine decode → body, identical-shaped outbound decode → none, both in one handler → the genuine schema. The house context holds a `*gin.Context` there, so the request sits two accessors from the root instead of one. I probed echo and fiber the same way first — they resolve identically, so I did not add two more modules; the preset table in `body_source_chain_test.go` covers all four request contexts.

**gofmt** — fixed in 2e56330.

Re-measured on gitea after all of it: 988 paths, 1243 operations, 246 components unchanged; 0 request bodies lost, `POST /restore_repo` still gained; runtime 147s against a 165–196s baseline. All 130 unrelated fixtures byte-identical.

PR #528, opened 2026-09-20
fix(engine): merge a --config over the detected defaults instead of replacing them

### LINKED ISSUE #524: A partial --config silently replaces the framework patterns: any config without a framework: block documents 0 paths and exits 0
## Summary

A `--config` file **replaces** the detected framework configuration instead of
being merged on top of it. A config that sets only `info`, `naming`,
`typeMapping` — anything without its own `framework:` block — therefore carries
no route patterns, and the run documents **zero paths** while exiting `0`.

This is the first thing a user does after reading the README's advice to set
`naming` for readable operation ids, so the cost is a silent empty spec at the
exact moment someone starts configuring the tool.

## Reproduction

`testdata/gin` (any project works). With no config:

```console
$ apispec --dir testdata/gin -o openapi.yaml
[engine] spec mapped (2 paths) in 5ms
Successfully generated: openapi.yaml
```

Now add a config that changes nothing about routing:

```yaml
# apispec.yaml
info:
  title: Users API
  version: 2.0.0
```

```console
$ apispec --dir testdata/gin -c apispec.yaml -o openapi.yaml
[engine] spec mapped (0 paths) in 1ms
[engine] no route registrations matched: 0 paths from 35 call edges across 1 package(s), with gin patterns in effect
[engine] if this project serves HTTP, then its router is unsupported, is wired in a style no pattern matched, or was excluded by --include-*/--exclude-* filters — docs/DEBUGGING.md walks through telling those apart
Generated (0 paths — nothing matched): openapi.yaml
$ echo $?
0
```

The generated document has `paths: {}` and `components: {}`.

## Cause

`internal/engine/engine.go` (~L797) selects the configuration with an
`else if` — a supplied file is used *instead of* the composed framework config,
never combined with it:

```go
if e.config.APISpecConfig != nil {
    apispecConfig = e.config.APISpecConfig
} else if e.config.ConfigFile != "" {
    apispecConfig, err = spec.LoadAPISpecConfig(e.config.ConfigFile)
    // ...
} else {
    apispecConfig, _ = ComposeFrameworkConfigFrom(frameworks, "")
}
```

The presets that run afterwards (`ApplySecurityPresets`,
`ApplyEntrypointPresets`, `DetectRouterWrappers`) do augment the user config, so
the replacement is specific to the framework pattern set.

## Two things make it hard to diagnose

1. **The diagnostic actively points the wrong way.** `reportNoRoutes`
   (`internal/engine/engine.go:1498`) prints *"with gin patterns in effect"*
   from `routeDiscovery.Frameworks` — the **detected** framework — which is
   independent of whether those patterns survived into the config that ran. It
   says gin's patterns are in effect at the exact moment they have been
   replaced, and the follow-up line then sends the reader looking for an
   unsupported router they do not have.
2. **Exit code is `0`.** Nothing distinguishes "this project has no routes" from
   "your config removed the ability to find them".

## Is the replacement itself intentional?

Partly, and that part is sound: merging *unscoped* patterns across frameworks is
what #211/#212 are about — gin's `Handle(method, path, h)` and mux's
`Handle(path, h)` misparse each other's calls, so "what you write is what
matches" is a defensible rule for `framework.*`. The README's mixed-framework
table already states it that way.

What is not defensible is that the rule silently extends to configs that say
nothing about `framework` at all. A file containing only `info:` has expressed
no opinion about route patterns and should not lose them.

## Suggested fix

In rough order of preference:

1. **Merge per top-level key.** Start from the composed framework config and
   overlay the user's file key by key. A user `framework:` block still replaces
   the framework block wholesale (keeping the #211 property) — but omitting it
   keeps the detected patterns, which is what every partial config means.
2. Failing that, **fail loudly**: when a `--config` is supplied whose
   `framework.routePatterns` is empty *and* a framework was detected, say so
   explicitly ("your config defines no route patterns and replaced the detected
   gin ones — start from `--output-config`") and exit non-zero.
3. At minimum, **stop the misleading diagnostic**: `reportNoRoutes` should
   report the patterns that were actually in effect, not the detected
   framework, and should name a user config as the likely cause when one was
   supplied.

Docs have been corrected in the meantime (`docs/CONFIGURATION.md` previously
documented the opposite — "your file is loaded *on top of* the detected
defaults … the framework patterns you omit still apply") and both the README
and the reference now prescribe building a config from `--output-config`. That
is a workaround, not the fix: `--output-config` dumps ~600 lines of framework
patterns that a user then has to keep in step with each release's defaults,
just to change a title.

## Version

`main` @ f773127 (v0.5.9+).

ehabterra:
Fixes #524.

## The bug

A supplied `--config` **replaced** the composed framework configuration rather than layering over it. So a file that changes nothing about routing:

```yaml
info:
  title: Users API
```

```console
$ apispec --dir testdata/gin -c apispec.yaml -o openapi.yaml
[engine] spec mapped (0 paths) in 1ms
$ echo $?
0
```

Zero paths, exit 0 — at exactly the moment someone starts configuring the tool, since readable operation ids (`naming`) are the first thing the README suggests reaching for.

## The fix: a real merge

A config file is now layered **over** the composed configuration key by key, at every level of nesting. What it states wins; what it does not mention keeps the composed value. Map keys merge too, so `securitySchemes` adds to the detected ones rather than displacing them.

Emptying a part is said **explicitly**, which is the distinction a struct cannot carry on its own — an omitted key and one written empty are the same zero value, so the difference has to come from the document:

```yaml
framework:
  routePatterns: []   # this project registers no routes my way
```

Measured on `testdata/gin`:

| config | before | after |
|---|---|---|
| `info.title` only | **0 paths** | 2 paths |
| `framework: {requestContext: …}` | **0 paths** | 2 paths |
| `framework: {}` | **0 paths** | 2 paths |
| `framework: {routePatterns: []}` | 0 paths | 0 paths |

Replacing was wrong in a subtler way than the report showed: a config that *did* name `routePatterns` silently lost the response, parameter and security patterns it never mentioned. That is fixed by the same change.

This needs no new merge code — unmarshalling onto an already-populated struct is exactly these semantics, so `LoadAPISpecConfigOnto` is the plain loader with a base.

`LoadAPISpecConfig` is unchanged for library callers: a file still parses on its own terms. Both are re-exported from the public `spec` package.

### Code-built configs

A config handed to the library as a struct has no document to layer, so it cannot express "absent" either. Those keep a narrower rule — take the detected framework only when carrying no routing opinion at all.

That distinction matters: an earlier version judged *every* config by "did the file declare `framework`?", which answers false for all code-built ones — including `NewGenerator(DefaultChiConfig())` with its patterns deliberately narrowed, whose scoping was then replaced by the detected defaults. The same silent loss, pointed the other way. `TestTestdata_PatternCallerScope` caught it; both directions now have a test.

## The diagnostic was wrong the same way

It reported `with gin patterns in effect` from the **detected** framework — independent of whether those patterns survived into the config that ran. It asserted they were in effect at the moment they had been replaced, and the follow-up line sent the reader after an unsupported router they do not have.

It now reports what was actually in effect, and when nothing was, names the cause:

```
[engine] 0 paths documented: the supplied config carries no route patterns, and a config
REPLACES the detected gin ones rather than adding to them — remove the `framework:` key to
keep them, or start from --output-config
```

The generic "your router may be unsupported" advice is suppressed there, because the cause is known.

## Measured

- **All 135 `testdata/` fixtures byte-identical.**
- Coverage 96.0% against the 96% floor.

Issue item 2 (a non-zero exit) is deliberately **not** included: that is `--strict`'s job in #297, and this makes the run succeed in the case it was actually failing.

## Summary by CodeRabbit

- **Bug Fixes**
  - Configuration files now layer over detected framework settings instead of replacing them, preserving route, response, parameter, and security patterns unless explicitly overridden.
  - Empty-spec diagnostics now describe the route patterns actually in effect and clarify when no patterns are configured.

- **New Features**
  - Added support for loading configuration onto an existing base configuration, with nested settings merged key by key.

- **Documentation**
  - Clarified standalone configuration loading behavior and documented how omitted and explicitly empty settings are handled.


ehabterra:
Both findings were valid — fixed in e13390f.

**`hasFrameworkOpinion` omitting fields (Major)** — confirmed. It enumerated the pattern slices and missed `HandlerInterfaceMethods`, the `RequestContext` accessors (including the `bodyReaders` added in #513) and most of `ResponseContext`. A code-built config setting only one of those was read as having no opinion, so `AdoptFrameworkPatterns` replaced its whole framework block and discarded the setting.

Rather than a longer list — which is the same bug waiting for the next field — it now asks the whole struct:

```go
func (c *APISpecConfig) hasFrameworkOpinion() bool {
	return !reflect.ValueOf(c.Framework).IsZero()
}
```

`TestHasFrameworkOpinionCoversEveryField` walks `FrameworkConfig` field by field and asserts each one, on its own, both counts as an opinion and survives `AdoptFrameworkPatterns`. All 10 fields pass, and a field added later is covered automatically.

**Zero-route-patterns remedy (Minor)** — confirmed, and you caught half of why. `UserConfig` did cover a config file and an in-code `APISpecConfig` alike, and a struct has no YAML key to remove and no `--output-config` to start from.

The other half is mine: that message was written for the replace-wholesale version. Now that a file is **merged** over the defaults (previous commit in this PR), *"remove the `framework:` key"* is wrong for a file too — a merged file reaches zero route patterns only by setting `routePatterns: []` on purpose. So the advice was stale, not just source-blind.

`RouteDiscovery` now records the source instead of a yes/no, and each gets the remedy that applies:

```
file: the config sets `framework.routePatterns` to an empty list, so the detected gin
      patterns were replaced by none — drop that key to inherit them again
code: the APISpecConfig passed in declares framework settings but no routePatterns, so the
      detected gin ones were not applied — add them, or leave the framework block empty to inherit
```

The test pins the old advice as **absent**, so the message cannot drift back to it.

Re-measured: all 135 fixtures byte-identical, coverage 96.0%.

Issue #359, opened 2026-08-16
Detected middleware implies responses, headers and params that never reach the spec (401/403, problem+json, 413/429/504, Idempotency-Key)


ehabterra:
## Context

apispec already does the hard part: it walks the middleware chain (router-level, group, per-route,
wrapper, var-assigned) and knows *which* middleware guards *which* route — that is how auth detection
emits `security`. But the middleware's other observable effects on the HTTP contract are dropped. A route
behind a bearer-auth middleware documents `security: [bearerAuth: []]` and then lists only the handler's
own responses — a client generated from that spec has no `401` to handle.

`grep -rn '"401"' --include='*.go' .` (excluding testdata) returns nothing: no status code is ever
inferred from middleware, only from handler calls.

Effects that are stated by the middleware and are pure spec facts:

| Middleware | Spec fact |
|---|---|
| auth (basic / bearer / API key / OAuth2) — already detected | `401`, and `403` where a scope/role check exists |
| CORS | `options` operation for preflight; `Access-Control-*` response headers |
| request size limit | `413` |
| timeout | `504` |
| rate limit | `429`, `Retry-After` / `RateLimit-*` response headers |
| panic recovery | `500` |
| problem details (RFC 9457) | error responses use `application/problem+json`, not `application/json` |
| idempotency | `Idempotency-Key` request header parameter |
| request ID | `X-Request-Id` response header |
| content-type validation | `415` |
| maintenance / drain | `503` |
| Accept-Patch (RFC 5789) | `application/merge-patch+json` / `application/json-patch+json` on PATCH bodies |

Common Go implementations are widely shared across frameworks (`gin-contrib/cors`,
`echo/middleware.*`, `gofiber/fiber/middleware/*`, `go-chi/chi/middleware.*`, `gorilla/handlers`,
`didip/tollbooth`, `justinas/alice` chains), so the detection surface is a recognisable set of packages
plus a config hook for custom ones.

## Proposed

Config-driven and framework-agnostic (golden rule #5), reusing the existing middleware walk rather than a
new one:

1. A `middlewareEffects` section in the framework/global config: match a middleware by package+symbol
   regex (the same matching style as the auth patterns) and attach *effects* — responses (status +
   description + optional content type), response headers, request header parameters.
2. Ship a default table for the well-known middleware above, per the recognised-libraries approach used
   for JWT/auth today; anything unrecognised produces the existing "unmapped middleware" warning and
   contributes nothing (golden rule #7) rather than a guess.
[...]

PR #529, opened 2026-09-20
fix(security): report a middleware as unmapped auth only if it reads a credential

### LINKED ISSUE #520: The unmapped-middleware warning reports every middleware as "auth": a request logger and a rate limiter are flagged, burying the true positive
From a downstream consumer's output-quality review (v0.5.9): 6 warnings with their config, 16 on defaults, and at least 4 of the 6 are not authentication. Reproduced on `main`.

## What

```go
r.Use(requestLogger)   // logs, never refuses
r.Use(rateLimit)       // 429, not auth
r.Use(authMiddleware)  // the only auth; mapped in securityMappings
```

```
[security] 2 auth middleware not mapped to a security scheme (add securityMappings to resolve):
  example.com/repro.requestLogger, example.com/repro.rateLimit
```

Neither is auth, and one of them cannot refuse a request at all.

## Cause

There is no classification step. In `extractor.go`, everything found in a "definite" middleware slot (`r.Use`, group, per-route, `With`) goes to `resolveSecurityWithShapes`, and whatever matches no `SecurityMapping` is recorded by `recordUnresolved` — then reported under the words "auth middleware". The list is really *"middleware with no security mapping"*, which for a normal service is most of it: logging, recovery, request-id, CORS, compression, rate limiting, timeouts.

## Why it matters

The true positive — an auth middleware nobody mapped, so its routes are documented as **public** — is the most important warning apispec prints. Buried among false ones it gets trained past, and it makes #297 (`--strict`) unusable as specified: a non-zero exit on "unmapped auth middleware" would fail every project with a logger.

## Proposed

Decide from facts, not names (golden rule #9 — no denylist of "logger"/"cors"):

1. **Known non-auth middleware** from recognised libraries (`chi/middleware.Logger|Recoverer|RequestID|Compress|Timeout`, `gin.Logger|Recovery`, echo/fiber equivalents, `rs/cors`, …) is classified by package+symbol in config, the same way the auth libraries are — and is exactly the table #359 needs for middleware *effects*, so the two should share one registry.
2. For a **house** middleware, report it as possible auth only when its body shows the auth shape: it reads a credential (`Authorization` header, a cookie, `r.BasicAuth()`, an API-key header/query — what the scheme detectors already look for) **and** has a refusing path (writes 401/403 or returns without calling `next`). `requestLogger` has neither; `rateLimit` refuses but reads no credential.
3. Everything else is not mentioned at warning level. A `--verbose`/insight listing of "middleware with no recognised effect" is fine.
4. Word the warning for what it then is, and name the consequence: *"N middleware look like authentication but map to no security scheme — their routes are documented as public: …"*.

## Acceptance

- `testdata/unmapped_middleware_classification`: logger, rate limiter, recognised-library recoverer, a mapped auth middleware and an **unmapped** house auth middleware — only the last is reported. Two frameworks / two wiring styles (golden rule #5).
- Unit test on the classifier for the credential-read + refusal rule.
- Unblocks the `security` category of #297.

ehabterra:
Fixes #520.

## The problem

Every middleware in a `Use`/group/per-route slot was announced as *"auth middleware not mapped to a security scheme"*:

```
[security] 5 auth middleware not mapped to a security scheme (add securityMappings to resolve):
  requestLogger, rateLimit, houseAuth, delegatingAuth, basicGuard
```

On a normal service most of that is logging, recovery, CORS, compression, rate limiting and timeouts. The reporting project saw **6 warnings where 2 were real**, and 16 on defaults.

That matters because of what the true one means: an auth middleware nobody mapped has its routes documented as **public**. It is the most important thing apispec prints, and a warning that is mostly false trains people past it.

## Scope, corrected

Half of the issue's proposal was already implemented — `securitySkipBundles()` marks the known non-auth middleware of chi, echo, gin, fiber, gorilla and rs/cors as `Skip`. A project's **own** middleware matches none of it, so all of it was reported. That house half is what this fixes.

## The rule

The split is made on what the body **does** — whether the middleware, or anything it calls, takes a credential out of the request — walking the same two edge sets `middlewareMatchesThrough` does, since a middleware is a function returning a handler and the work lives in the closure.

Never on the name: a denylist of `logger`/`cors` is the guess golden rule #9 forbids, and a house middleware called `guard` should be judged by what it reads.

**Refusing a request is deliberately not sufficient.** A rate limiter refuses too — which is why both non-auth middleware in the fixture refuse, so the test would fail a classifier keying on "can this reject a request".

## Nothing is dropped

Middleware showing no credential read is listed under `--verbose` and exposed on `SecurityDiagnostics`/`Engine.GetUnclassifiedMiddleware()`:

```
Middleware with no security mapping and no credential read (not reported as auth):
  requestLogger, rateLimit
```

So a project whose auth middleware fetches its credential somewhere this walk does not reach can still find it. The strict rule costs visibility, not the answer — which is what makes it safe to be strict.

The warning also states the consequence rather than only the condition:

```
[security] 3 middleware read a credential but map to no security scheme, so the operations
they guard are documented as PUBLIC (add securityMappings to resolve): …
```

## Where the table lives

`framework.credentialReads` — the credential names, and the calls that *are* a read whatever they are passed (`r.BasicAuth()`, a context's `Cookie`). Config, both so a house credential can be declared and because it is the surface #359 needs to attach a middleware's other effects; that issue can extend this table rather than introduce a second one.

One thing worth knowing: the receiver is matched in **both** spellings metadata uses. A method call on `*http.Request` records `RecvType` as a bare `*Request` with the path in `Pkg`, and pinning the qualified form alone made `r.BasicAuth()` invisible to the classifier — a false negative I only caught because the fixture has a BasicAuth case.

## Measured

| | before | after |
|---|---|---|
| probe (6 middleware, 1 mapped) | 5 reported | **3**, all real — header name, a helper one hop down, and `BasicAuth` |
| gitea | 1 reported | **0** — it was `wrapMiddlewareAndHandler`, a generic wrapper |

gitea's 988 operations have **identical `security` blocks**, so only diagnostics changed. All 135 fixtures byte-identical; coverage 96.0%.

The name table is guarded against the obvious failure: `X-Request-Id`, `X-Api-Version`, `X-Forwarded-For` and friends must **not** read as credentials, since matching one would report every logging middleware as auth — the bug inverted.

This unblocks the `security` category of #297, which could not have shipped while a non-zero exit on "unmapped auth middleware" failed every project with a logger.

## Summary by CodeRabbit

- **New Features**
  - Improved security diagnostics by identifying middleware that reads credentials or rejects unauthorized requests.
  - Clearly distinguishes authentication-related middleware from logging, recovery, and rate-limiting middleware.
  - Reports when authentication mappings are missing and routes may otherwise appear public.
  - Added configurable credential-reading patterns and refusal statuses for supported web frameworks.
  - Added a **Credential reads** section to the configuration interface.

- **Documentation**
  - Added unreleased changelog documentation covering the updated security diagnostics.


ehabterra:
Implemented the OR in 595c42b — you were right that a status is the stronger signal, and it closes a real miss in the first version.

I probed both rules against six middleware chosen to separate them:

| middleware | credential-read | 401/403 |
|---|---|---|
| reads `Authorization`, writes 401 | ✅ | ✅ |
| reads a cookie, **redirects** to /login | ✅ | ❌ miss |
| reads a header, **returns an error** for a shared renderer | ✅ | ❌ miss |
| reads `X-Acme-Request-Signature`, writes 403 | ❌ miss | ✅ |
| rate limiter, 429 | excluded | excluded |
| tenancy guard, 403 | excluded | ⚠️ included |

So: **as well, not instead.** Either signal is now sufficient.

**Why the status half earns its place.** A credential *name* table is a guess about spelling and will never contain `X-Acme-Request-Signature`; a status is semantic, since HTTP defines 401 as "not authenticated". That was a silent false negative in the first version — routes documented public with no warning.

**Why the credential half stays.** A status is often written somewhere the middleware's own call graph never reaches: session auth redirects to a login page and writes no 401 (gitea is exactly this shape), and a middleware returning an error for echo's `HTTPErrorHandler` or a house `renderErr` decides its status elsewhere entirely.

**The accepted cost** is the last row. A middleware that forbids on something other than a credential — a tenancy or IP guard — is now reported. It is the same family (403 *is* authorisation), and the trade is deliberate: a false positive costs one line, a miss costs a route documented as public.

429/413/415/504 stay excluded, which is what stops "refuses a request" from becoming the signal — a rate limiter refuses too. `TestRefusalStatusesAreAuthOnly` pins that, since counting them would report every guard as auth and undo the whole fix.

Statuses resolve in both spellings a handler writes them: a bare `401`, and `http.StatusUnauthorized` via `HTTPStatusByName` — the same table the schema mapper uses — matched on the trailing identifier so an aliased import works.

The fixture now carries all six shapes, and each signal is the only thing catching at least one of them, so a rule built on either alone fails the table.

Re-measured: gitea unchanged (0 reported, all 988 operations' security blocks identical), 135 fixtures byte-identical, coverage 96.0%.

ehabterra:
All four addressed in fc2cb7d — two were silent defects, and the Major one was right about the shape of the whole thing.

**Argument values without call semantics (Major)** — the strongest finding. A signal was any credential-looking *value* in any argument of any call. I reproduced all three of your examples and every one was reported before, none is after:

| middleware | before | after |
|---|---|---|
| `log.Println("Authorization", …)` | reported as auth | not reported |
| `r.Header.Set("Authorization", …)` (adds an **outbound** credential) | reported as auth | not reported |
| `observe(http.StatusForbidden)` | reported as auth | not reported |

Both signals now come from what the **call does**. A credential name counts only as the argument of a by-name read (`credentialReads.namedReads`, receiver-scoped so `Get` cannot match `cache.Get`). A refusal status counts only where the framework's own `responsePatterns` say a status is written, at the argument index they name — which keeps it framework-agnostic per golden rule #5 and reuses config a project can already edit, rather than inventing a second list.

That exposed a convention worth recording: a response pattern scopes a **package-level** writer by putting the package in `RecvTypeRegex` (net/http's `Error`, `NotFound`, `Redirect`), and a package-level call records no receiver. Reading the field literally matched none of them and lost every `http.Error` refusal — caught because the fixture has a 403 case.

**`DefaultHTTPConfig` not wired (Major)** — confirmed, and worse than stated: the secondary merge view (`config_merge.go`) was missing it too. A plain net/http project had an empty credential surface, so `callSignalsAuth` returned false before looking at anything and **all** of its auth middleware went to the unclassified list. That is this issue's bug, inverted and silent. Both sites fixed.

**Memoized negatives (Major)** — confirmed. Whether a function shows a signal depends on the walk state that reached it, so a helper first visited at the depth bound or through a cycle cached "no signal" and made the next middleware to reach it by a shorter route silent. Only positives are cached now; a positive is state-independent.

**Changelog wording (Minor)** — fixed. Either signal *is* sufficient; neither *covers* every shape. Also replaced "shows no credential read" with "shows neither signal" throughout.

The fixture now carries the three false positives as negative cases, so a rule that accepts a bare value fails it.

Re-measured: gitea unchanged (0 reported), 135 fixtures byte-identical, coverage 96.0%.

Issue #368, opened 2026-08-16
Pointer fields lose nullability: `*string` is documented as `type: string`, so a `null` the encoder really emits is undocumented


ehabterra:
## Context

The pointer branch of the schema mapper returns the pointee's schema unchanged
(`internal/spec/mapper.go:2868-2875`), with a comment that says as much:

```go
	// Handle pointer types
	if goTypeRef.Kind == typemodel.KindPointer {
		underlyingType := strings.TrimSpace(goTypeRef.Elem.Raw())
		// For pointer types, we generate the same schema as the underlying type
		// but we could add nullable: true if needed for OpenAPI 3.0+
```

`encoding/json` writes `null` for a nil pointer field that has no `omitempty`, so the spec is claiming a
shape the API does not actually guarantee, and a strict client validator rejects a response the server
legitimately sends.

### Reproduction

```go
type Item struct {
	Name string  `json:"name"`
	Note *string `json:"note"`  // nil -> `"note": null` on the wire
}
```

Generated:

```yaml
    probe_gin_Item:
      type: object
      properties:
        name: {type: string}
        note: {type: string}     # no null allowed
```

## Proposed

Derive nullability from what the encoder actually does, not from the pointer alone — that keeps the
rule honest (golden rule #7) and mechanical:

| Field | Wire behavior | Schema |
|---|---|---|
| `*T` without `omitempty` | present, may be `null` | nullable, and part of `required` only if it already would be |
| `*T` with `omitempty` | absent when nil, never `null` | unchanged (optional, as today) |

Encoding, OpenAPI 3.1 (`openapi: 3.1.1` is what apispec emits — JSON Schema 2020-12, so no `nullable:`
keyword):

- primitive/inline schema → `type: [string, "null"]`;
- `$ref` → `anyOf: [{$ref: '#/components/schemas/X'}, {type: "null"}]`, since a `$ref` sibling `type` is
  not allowed;
- pointer-to-slice / pointer-to-map → same union treatment on the array/object schema.

Scope note: Go slices and maps are also nil-able and encode as `null`; this issue is deliberately about
pointers only, which are the case where the *declared* type carries the intent. A follow-up can decide
whether a nil slice deserves the same union.

Existing fixtures whose structs use pointers will gain the union — a deliberate output change with its
own fixture, not a refactor (the "zero output drift" rule in CLAUDE.md applies to refactors).

## Acceptance

[...]

PR #530, opened 2026-09-20
feat(spec): derive `required` from what encoding/json does, opt-in

### LINKED ISSUE #516: Schemas never get a `required` list from JSON tags: a field without `omitempty` is always on the wire, and 77% of properties on a real service are
From a downstream consumer's output-quality review, where it is measured as the single largest cost to a generated TypeScript client. Reproduced on `main`.

## What

No schema ever gets a `required` list unless a field carries `validate:"required"` (`mapper.go`, `validationConstraints.Required` is the only producer). But `encoding/json` already states presence precisely: a field with no `omitempty`/`omitzero` is **always** on the wire.

```go
type Settings struct {
	Name  string  `json:"name"`            // always present
	Note  *string `json:"note"`            // always present, may be null  (#368)
	Extra string  `json:"extra,omitempty"` // may be absent
}
```

```yaml
    Settings:
      type: object
      properties: { extra: {type: string}, name: {type: string}, note: {type: string} }
      # no `required`
```

Measured by the consumer against `go/types`: of 2,900 properties across 452 schemas, **2,233 (77%) are always present** and 386 are nullable pointers; the spec marks 0 and 0. `openapi-typescript` therefore emits 2,875 of 2,878 members as optional, so the client null-checks 2,233 fields the server always sends and gets no signal on the ones that really can be absent.

## Proposed

Derive `required` from what the encoder does — the same principle #368 uses for nullability, and the two should land as one coherent rule:

| Field | Wire | Schema |
|---|---|---|
| `T`, no `omitempty` | always present | in `required` |
| `*T`, no `omitempty` | always present, may be `null` | in `required` **and** nullable (#368) |
| any, with `omitempty` / `omitzero` | may be absent | not required |
| `json:"-"`, unexported | never | not a property (already right) |

Design points to settle here rather than in review:

- **Responses vs requests.** The rule above is a fact about *encoding*. On *decode* every field is optional unless validation says otherwise, so applying it to a request body would reject payloads the server accepts. Either (a) apply it only to schemas reached from responses, (b) emit it everywhere behind an option and document the caveat, or (c) split a type used in both directions. (a) is the honest default (golden rule #7); a type used both ways keeps today's behavior unless `validate:"required"` says more.
- **Opt-in first.** It changes nearly every component, so ship it behind a config switch (`schema.requiredFromTags` or similar), default off for one release, with the intent to flip it.
- **Custom marshalers** (`MarshalJSON`) make the field set unknowable — skip them (ties to #361).
- Embedded structs: promoted fields follow the same rule; an embedded *pointer* struct's fields are present only when the pointer is non-nil, so they are not required.
- Merge with, never replace, `validate:"required"`.

## Acceptance

- `testdata/required_from_tags`: the table above, plus an embedded struct, an embedded pointer struct, a type used as both request and response, and a `MarshalJSON` type.
- Option off ⇒ zero drift across every existing fixture and golden.
- `required` lists are sorted (determinism, golden rule #1).

ehabterra:
Closes #516. The evaluation measured this as the largest single gain available to a TypeScript consumer.

## What was missing

`encoding/json` already states which fields are always on the wire — one with no `omitempty` and no `omitzero` is written on every encode — and none of it reached the document.

On the 452-schema service that reported it, **2,233 of 2,900 properties are always sent and not one said so**. A generated TypeScript client therefore null-checked every field, and got no signal on the 386 that really can be absent.

## Presence only

Whether a value may be `null` is a separate statement about the *value*, and belongs to #368. A `*T` with no `omitempty` is always **present** — written as `null` — so it is `required` here and nullable there. The two compose; neither implies the other.

## The rule

| field | wire | `required` |
|---|---|---|
| `T`, no `omitempty` | always written | ✅ |
| `*T`, no `omitempty` | always written, as `null` when nil | ✅ |
| any, `omitempty` / `omitzero` | may be absent | ❌ |
| promoted through an embedded **pointer** | absent entirely when the embed is nil | ❌ |
| `json:"-"`, unexported | never | not a property (unchanged) |

The embedded-pointer row is the one a tag cannot tell you: `encoding/json` writes *nothing* for a nil embed, however the promoted field's own tag reads. `effectiveField` now carries whether its path passed through a pointer embed, sticky down the rest of the path — so it holds at any depth, and a **tagged** pointer embed (which carries the object rather than promoting) is correctly unaffected.

`validate:"required"` is **merged**, not replaced — a field can be mandatory on the way in and always present on the way out — and the list is sorted, since it is now fed from two tag sources and its order must not depend on which arrived first (golden rule #1).

## Why opt-in

It states what the **server sends**. That is exactly right for a response and an over-claim for a request body: a client is not bound by the server's struct tags, and the server decodes a payload that omits the field perfectly happily. A type used as both declares more than the server enforces.

That is a trade worth taking on knowingly rather than inferring for someone, so `schema.requiredFromJSONTags` is **off by default** — and with it off, **all 135 fixtures are byte-identical**, which is every existing project.

## Coverage

`testdata/required_from_tags` carries every row of the table plus the controls: a value embed, a pointer embed, `omitzero`, a `json:"-"` field, a `validate:"required"` field (asserted to appear exactly once), and an external type. Tests cover both settings, and `TestEffectiveFieldsCarryPointerEmbeds` pins the pointer flag at depth directly.

One small cleanup came out of the coverage ratchet: a guard in the tag reader was unreachable (`Split` on a string that already contains the key always yields a second part), so it is deleted rather than tested.

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in option to derive required schema fields from JSON tags.
  * Fields without `omitempty` or `omitzero` are marked as required.
  * Validation-based requirements are preserved and merged without duplicates.
  * Fields promoted through nil embedded pointers remain optional.
  * Required fields are listed in a consistent order.
* **Documentation**
  * Documented the new configuration option and its default-disabled behavior.


ehabterra:
Both fixed — and the first was my own doc comment promising a check I never wrote.

**Self-marshaling structs (Major)** — confirmed. `requiredFromTags`' doc comment already claimed marshaler types were excluded; nothing implemented it. A type declaring `MarshalJSON` does not put its declared fields on the wire, so saying they are always present describes a shape that is never sent. The guard is computed once per struct (it's a property of the type, not of each field), and the fixture gains a `Money`/`Invoice` pair so an ordinary struct beside a self-marshaling one is shown unaffected:

```
Money    required=[]            # declares MarshalJSON
Invoice  required=[total, ref]  # ordinary
```

**Tag parsing (Minor)** — confirmed. Splitting on `"json:"` also matches a key *ending* in json, so `myjson:"id,omitempty"` was read as an omitempty the encoder never sees, leaving an always-present field out of `required`. Now `reflect.StructTag.Lookup("json")`, with three tag shapes pinned including `myjson:"id" json:"id,omitempty"` to prove the right one wins.

**One thing your review led me to that neither comment named.** Verifying the marshaler fix showed `Invoice.required = [total, ref]` — unsorted, although I had added a sort. The sort had gone into the alias/enum function rather than the struct one, where it did nothing; `Item` only *looked* sorted because its declaration order happens to be alphabetical.

Moving it to the right place then drifted **five existing fixtures**, reordering the `required` list of every project already using `validate:"required"`. So I dropped the sort instead: its justification was wrong, because both tag sources append inside one loop over a fixed field order, so the list is already reproducible without it. Declaration order, with a test that says so.

Back to **all 135 fixtures byte-identical**, coverage 96.0%.

PR #531, opened 2026-09-20
feat(spec): admit null for every kind the encoder nulls, opt-in

### LINKED ISSUE #368: Pointer fields lose nullability: `*string` is documented as `type: string`, so a `null` the encoder really emits is undocumented

ehabterra:
Closes #368. **Stacked on #530** (`feat/516-required-from-tags`) — review that first; this targets it rather than main, because the two are halves of one statement.

## The bug

A `*T` with no `omitempty` is always written, and written as `null` when the pointer is nil. Documenting it as `type: string` claims a shape the API does not guarantee, so a client validating against the document rejects a response the server legitimately sends.

With `omitempty` the field is absent instead of null, so it is left alone.

## Why it belongs with #516

#516 makes such a field `required`. On its own that produces **the one combination worse than neither** — the document insisting a field is always there and never null, while the server sends null. Composed, they say what is actually true:

```yaml
ptr:                                   # *string, no omitempty
  anyOf: [{type: string}, {type: "null"}]
required: [..., ptr, ...]              # always written
```

Always **present**, sometimes **null**. Different statements; neither implies the other.

## The encoding

`anyOf: [{…}, {type: "null"}]`, not `type: [T, "null"]`. `Schema.Type` is a single string, and more to the point **a `$ref` may carry no sibling keywords**, so the union is the only form available there — using one shape for both saves a generated client from handling two.

| field | result |
|---|---|
| `*string` | `anyOf: [{type: string}, {type: null}]` |
| `*Base` | `anyOf: [{$ref: …}, {type: null}]` |
| `*[]string` | union wraps the **array**; its items are not nullable |
| `*string` with `omitempty` | unchanged — absent, never null |
| `string` | unchanged |

The description stays on the property rather than moving into a branch, where a reader would not look for it.

## Opt-in

`schema.nullableFromPointers`, off by default, like its other half — **all 135 fixtures byte-identical** with it off.

The issue asked for default-on. I've kept it opt-in deliberately: the pair is better flipped together in a release that documents the change than one half at a time, and turning `required` on without this is the combination to avoid. Happy to flip both defaults in a follow-up if you'd rather.

A type that declares `MarshalJSON` is excluded here too — its declared fields are not the wire shape, so neither statement about their tags applies.


ehabterra:
**Note on CI:** this PR targets `feat/516-required-from-tags` (#530), and the workflows are configured as `pull_request: branches: [main, develop]` — so no CI runs on a stacked PR. Only the CodeRabbit check reports.

Verified locally in the worktree instead, on the full stack:

- `go test ./...` — pass
- `make lint` — pass (golangci-lint, vet, gofmt)
- `scripts/check-coverage.sh` — **96.0%** against the 96% floor
- **135 fixtures byte-identical** with the option off

Once #530 merges I'll retarget this to `main`, which will pick up the full CI run — or say the word and I'll retarget now, at the cost of the diff showing both changes together.

ehabterra:
## Hold this — a reviewer question found a real gap

> `Tags *[]string `json:"tags"`` — I don't know but slice itself is a pointer

Correct, and it goes further than the odd field. Verified with a throwaway program:

```go
type T struct {
	Slice    []string       `json:"slice"`
	Map      map[string]int `json:"map"`
	PtrSlice *[]string      `json:"ptrSlice"`
	Str      string         `json:"str"`
	Arr      [2]string      `json:"arr"`
	Iface    any            `json:"iface"`
}
json.Marshal(T{})
```
```json
```

A nil `[]string` writes `null`. So does a nil map and a nil `any`. Only `string` and `[N]T` are safe.

### Two problems with this PR as it stands

**1. The fixture field is unidiomatic.** `*[]string` is a double indirection almost nobody writes — precisely *because* a slice is already nilable. I used it to exercise the pointer-to-container path and in doing so picked a shape real code does not have, while the shape real code *does* have is untested.

**2. Plain `[]T` and `map[K]V` are not widened at all.** A response DTO with `Items []Item` that is nil when empty is extremely common: the document says `type: array` and the server sends `null`. That is the exact bug #368 is about, in the case that actually occurs — so this PR fixes the rare shape and misses the common one.

### Why it was scoped out, and why that was wrong

#368's own text says: *"Go slices and maps are also nil-able and encode as null; this issue is deliberately about pointers only, which are the case where the declared type carries the intent. A follow-up can decide whether a nil slice deserves the same union."*

The intent argument is real — `*T` signals "this can be absent" where `[]T` nil-ness is often an accident of construction. But the spec describes the **wire**, and on the wire the two are identical. On the evidence, the scoping was wrong.

### The genuine counter-argument

Many Go APIs deliberately return `[]T{}` rather than nil, so `null` never appears on them, and widening every array is noise. apispec cannot tell which convention a project follows — which makes this a policy call rather than a pure correctness fix, and is probably why I scoped it out.

### Proposed

Extend to everything the encoder writes as `null` — slice, map, interface; **not** `string`, **not** `[N]T` — and rename the option, since `nullableFromPointers` would no longer describe it (`nullableWhenNil` or similar). Projects that always return `[]T{}` leave it off, which is already the default.

### Status

Holding rather than merging: the fixture advertises a shape I would change, and the option name would be renamed, so merging now would ship an API surface I already expect to revise. #530 (the `required` half) is unaffected and still ready on its own.

Recorded here so the finding is not lost — credit to the review question, which is the only reason it surfaced before release.

ehabterra:
## Hold lifted — rule widened, option renamed

Acted on the finding above in 8c1d8eb, so this PR now fixes the shape that actually occurs.

**The rule** is every kind whose zero value `encoding/json` writes as null, not pointers alone:

| field | widened? | why |
|---|---|---|
| `[]string` | ✅ | nil slice → `null` — **the case the first version missed** |
| `map[string]string` | ✅ | nil map → `null` |
| `any` | ✅ | an interface holding nothing → `null` |
| `*string`, `*Base` | ✅ | as before |
| `[2]string` | ❌ | a fixed-size array cannot be nil |
| `string` | ❌ | writes `""` |
| any of the above with `omitempty` | ❌ | absent instead of null |

The type is **parsed, not prefix-matched**: `[]T` is nilable and `[2]T` is not, and both start with `[`. That one distinction is why a string check would have been wrong.

**Renamed** `schema.nullableFromPointers` → `schema.nullableWhenNil`, since the old name no longer describes the rule.

**Fixture corrected.** `Tags *[]string` — the double indirection you flagged — is now `Tags []string`, plus a map, an `any`, a `[2]string` and a slice with `omitempty`. The old field advertised a shape real code does not have, which is exactly how the gap went unnoticed.

**A named type whose underlying type is nilable is left alone.** Resolving that needs the declaration, and an alias for a slice is rare beside the cost of being wrong about a struct (golden rule #7).

Still opt-in, and now for a second reason beyond composing with #516: many Go services deliberately return `[]T{}` rather than nil, so `null` never appears on them and widening every array would be noise — a convention apispec cannot see.

All 135 fixtures byte-identical with it off; coverage 96.0%; full suite and lint green locally. CI still does not run here because the PR targets #530's branch — retargeting to `main` once that merges.

PR #532, opened 2026-09-20
feat(spec): admit null for every kind the encoder nulls, not just poi…

### LINKED ISSUE #368: Pointer fields lose nullability: `*string` is documented as `type: string`, so a `null` the encoder really emits is undocumented

ehabterra:
…nters

Review question on the first version: "slice itself is a pointer". Correct, and it makes the pointer-only framing the wrong shape. Verified:

A nil slice, map or interface is written as `null` exactly as a nil pointer is. Only a string and a fixed-size array cannot be nil.

#368 scoped itself to pointers because `*T` carries the INTENT to be absent where a nil `[]T` is usually an accident of construction. That is true of reading the Go and irrelevant to a document describing the wire, where the two are the same value — so the first version fixed the RARE shape (`*[]string` is a double indirection almost nobody writes, since a slice is already nilable) and missed `Items []Item`, which is on every list response there is. The fixture advertised the rare one, which is how it went unnoticed.

So the rule is now every kind whose zero value encodes as null, and the option is renamed to match: nullableWhenNil, since nullableFromPointers no longer describes it.

The type is PARSED rather than prefix-matched, because the one distinction that decides it is invisible to a prefix: `[]T` is nilable, `[2]T` is not, and both start with '['. A named type whose underlying type is nilable is left alone — resolving that needs the declaration, and an alias for a slice is rare beside the cost of being wrong about a struct (golden rule #7).

Still opt-in, and now for a second reason beyond composing with #516: many Go services deliberately return `[]T{}` rather than nil, so null never appears on them and widening every array would be noise. apispec cannot see which convention a project follows.

All 135 fixtures byte-identical with it off; coverage 96.0%.

Closes #368

## Summary by CodeRabbit

* **New Features**
  * Added opt-in `nullableWhenNil` schema behavior for nil pointers, slices, maps, and interfaces.
  * Generated schemas now represent eligible nil values as accepting `null`, while preserving item types.
  * Fields using `omitempty` and non-nilable types such as fixed-size arrays remain unchanged.

* **Documentation**
  * Updated the unreleased changelog with configuration details and the distinction between nullability and required-field inference.

* **Breaking Changes**
  * Replaced `nullableFromPointers` with `nullableWhenNil`; existing configuration using the former key no longer takes effect.


ehabterra:
Checked this — **ready to merge**, and it is the one to keep. I closed #533, which I had opened for the same reason before seeing yours; the two branches are byte-identical:

```
$ git diff origin/fix/368-pointer-nullability origin/fix/368-nullable-when-nil --stat
(no output — identical trees)
```

Verified independently on a cherry-pick onto `main`: full suite and lint green, coverage **96.0%** against the 96% floor, and **all 135 fixtures byte-identical** with the option off. CI here agrees — all seven checks pass.

### One correction to the auto-generated summary

CodeRabbit lists this under **Breaking Changes**:

> Replaced `nullableFromPointers` with `nullableWhenNil`; existing configuration using the former key no longer takes effect.

Accurate about the code, misleading as a release note. `nullableFromPointers` has **never been in a release** — it reached `main` only via #531 earlier today, and no tag contains it:

```
v0.5.9: 0   v0.5.8: 0   v0.5.7: 0   occurrences of nullableFromPointers
```

So no configuration in the wild can be using the old key, and nothing breaks. Worth dropping that bullet before it lands in the changelog, or it will send people looking for a migration they do not need.

It does mean this is best merged **before** the next release — after one, the rename really would be breaking.

### Why the rename earns its keep

The old name stopped describing the rule. A nil slice, map or interface is written as `null` exactly as a nil pointer is:

```json
```

The pointer-only version now on `main` fixes the rare shape (`*[]string` — a double indirection almost nobody writes, since a slice is already nilable) and misses `Items []Item`, which is on every list response there is.

PR #533, opened 2026-09-20
feat(spec): admit null for every kind the encoder nulls, not just pointers

ehabterra:
Follow-up to #531, which merged at its **first** commit — the widening I pushed after the review finding (`8c1d8eb`) never reached `main`. This lands it.

`main` today has `nullableFromPointers`, the `Tags *[]string` fixture field, and no coverage of nil slices or maps, which is the state I'd asked to hold. Nothing has been released from it, so renaming the option now costs nothing.

## The finding, restated

The review question on #531 was *"slice itself is a pointer"*. Verified:

```json
```

A nil slice, map or interface is written as `null` exactly as a nil pointer is. Only a string and a fixed-size array cannot be nil.

#368 scoped itself to pointers because `*T` carries the **intent** to be absent where a nil `[]T` is usually an accident of construction. That is true of reading the Go and irrelevant to a document describing the wire, where the two are the same value — so the merged version fixes the **rare** shape (`*[]string` is a double indirection almost nobody writes, since a slice is already nilable) and misses `Items []Item`, which is on every list response there is.

## What this changes

| field | widened? | why |
|---|---|---|
| `[]string` | ✅ | nil slice → `null` — **the case that was missed** |
| `map[string]string` | ✅ | nil map → `null` |
| `any` | ✅ | an interface holding nothing → `null` |
| `*string`, `*Base` | ✅ | unchanged |
| `[2]string` | ❌ | a fixed-size array cannot be nil |
| `string` | ❌ | writes `""` |
| any of the above with `omitempty` | ❌ | absent instead of null |

The type is **parsed, not prefix-matched**: `[]T` is nilable and `[2]T` is not, and both start with `[`. That one distinction is why a string check would have been wrong, and it has its own case.

`schema.nullableFromPointers` → **`schema.nullableWhenNil`**, since the old name no longer describes the rule.

**Fixture corrected.** `Tags *[]string` becomes `Tags []string`, plus a map, an `any`, a `[2]string` and a slice with `omitempty`. The old field advertised a shape real code does not have, which is how the gap survived review in the first place.

A **named type** whose underlying type is nilable is deliberately left alone: resolving it needs the declaration, and an alias for a slice is rare beside the cost of being wrong about a struct (golden rule #7).

## Measured

- **All 135 fixtures byte-identical** with the option off.
- Coverage 96.0% against the 96% floor; full suite and lint green.

Still opt-in, and now for a second reason beyond composing with `requiredFromJSONTags`: many Go services deliberately return `[]T{}` rather than nil, so `null` never appears on them and widening every array would be noise — a convention apispec cannot see.


ehabterra:
Closing as a duplicate of #532 — same change, and yours came first.

I opened this after finding that #531 had merged at its first commit, so the widening (`8c1d8eb`) never reached `main`. You had already opened #532 from the original branch for the same reason. Verified they are identical:

```
$ git diff origin/fix/368-pointer-nullability origin/fix/368-nullable-when-nil --stat
(no output — identical trees)
```

Both are fully green. #532 is the one to merge.

Issue #517, opened 2026-09-20
A body streamed through a writer built on `w` (`csv.NewWriter(w)`, `io.Copy(w, …)`) is not a response: the operation documents no success at all


ehabterra:
From a downstream consumer's output-quality review; reproduced on `main`, default chi config. Split from #354 (media type) and #471 (buffer sinks) because it is a third shape neither covers.

## What

A handler that streams its body through a **writer wrapped around `w`** has no success response at all — only whatever error branch happens to use a recognised helper.

```go
func csvExport(w http.ResponseWriter, r *http.Request) {
	w.Header().Set("Content-Type", "text/csv")
	cw := csv.NewWriter(w)
	if err := cw.Write([]string{"a", "b"}); err != nil {
		respondJSON(w, 500, map[string]any{"error": err.Error()})
		return
	}
	cw.Flush()
}
```

### Actual

```yaml
  /api/files/csv:
    get:
      responses:
        "500": { … }        # the only response
```

### Expected

A `200` with `text/csv` (from the constant `Content-Type`; see #354) and a string/binary schema. A client generated from today's spec believes this endpoint can only fail.

## Why it is its own issue

- #354's remaining half is "a raw `w.Write` after an explicit `Content-Type` is still `application/json`" — there, a write **is** detected and only the media type is wrong. Here **no write is detected**.
- #471 / #510 follow a *buffer forward* to the call that flushes it into `w` (`BufferSinks`). Here there is no buffer: the wrapper is constructed **over** `w` and writes through it. The relation is the constructor argument, i.e. the same backward resolution `destFromReceiver` already does for `json.NewEncoder(w).Encode(v)` — it just is not declared for anything but encoders.

Shapes in the same family, all "a writer built on `w`, written to via its own methods":

| constructor | body writes |
|---|---|
| `csv.NewWriter(w)` | `Write`, `WriteAll` |
| `gzip.NewWriter(w)` / `zlib` / `flate` | `Write` |
| `bufio.NewWriter(w)` | `Write`, `WriteString` |
| `zip.NewWriter(w)`, `tar.NewWriter(w)` | `Create`/`Write` |
| `io.Copy(w, src)`, `io.CopyN`, `src.WriteTo(w)` with a non-buffer `src` (a file, an upstream body) | the call itself |
| `fmt.Fprintf(w, …)`, `io.WriteString(w, …)` | the call itself |
| `http.ServeContent` / `http.ServeFile` | the call itself |

## Proposed

[...]

Issue #519, opened 2026-09-20
A user-declared serializer pattern (`^Marshal$`) bypasses the forward trace the defaults use, so outbound request bodies are documented as responses — now under `200`


ehabterra:
Follow-up to #294 (closed by the config lint in #303 / #500). From the same downstream consumer's re-review on v0.5.9: the lint fires, and the false responses are still in the spec. Their position, which I think is right: *a consumer should not have to keep its config aligned with each release's defaults to get a correct spec.*

## What

The shipped defaults handle `json.Marshal` correctly — `ResponseContext.BodyTransforms` traces the marshalled bytes forward and documents the value only when they reach the response writer (#195/#197). A **user-declared** pattern naming the same serializer gets none of that:

```yaml
responsePatterns:
  - callRegex: ^Marshal$
    typeFromArg: true
    deref: true
```

```go
func send(to string) error {
	b, _ := json.Marshal(Payload{To: to})   // OUTBOUND provider body
	req, _ := http.NewRequest(http.MethodPost, "https://provider.example/send", bytes.NewReader(b))
	_, err := http.DefaultClient.Do(req)
	return err
}

func createWidget(w http.ResponseWriter, r *http.Request) {
	if err := send("a@b.c"); err != nil { http.Error(w, "upstream", 502); return }
	w.WriteHeader(http.StatusCreated)
	_ = json.NewEncoder(w).Encode(Widget{ID: "1"})
}

func getWidget(w http.ResponseWriter, r *http.Request) {
	b, _ := json.Marshal(Widget{ID: "1"})   // legitimate: bytes go to w
	w.WriteHeader(http.StatusOK)
	_, _ = w.Write(b)
}
```

### Actual on `main` (default chi config + that one pattern)

```
[config] 1 response pattern(s) match a bare call name anywhere in the call graph …
POST /widgets   200: Payload     ← the provider's request struct
                201: Widget
                502: string
GET  /widgets/1 200: Widget
```

Note it is now worse than when #294 was filed: since #369 an undetermined status resolves to `200`, so the outbound body is no longer a harmless-looking `default:` — it is a **false success type** sitting beside the real `201`.

Without the pattern both routes are already correct, including `getWidget` — `BodyTransforms` covers the legitimate shape.

On the consumer's service: 56 operations carry the spurious block, 8 internal component schemas (a provider payload, `context.Context`, a credentials struct, a generated query struct) are published that should not be, and removing the one pattern loses **0** correct responses.

[...]

PR #534, opened 2026-09-20
feat(spec): document a streamed body from its declared Content-Type

ehabterra:
Refs #517. **Please read the trade-off section before merging** — this ships with a known, measured cost that I could not close.

## The problem

Response detection recognises a **value being encoded**. A handler that *streams* has nothing for it to see, so the operation documented no success at all — only whatever error branch happened to use a recognised helper:

```
GET /export.csv   500 only    ← claims an endpoint that can exclusively fail
GET /file.pdf     500 only
```

On the reporting service that was 21 downloads plus a CSV export.

## The signal: what the handler declares, not which writer it used

```go
w.Header().Set("Content-Type", "text/csv")
```

| | after |
|---|---|
| `GET /export.csv` | `200` **text/csv** |
| `GET /file.pdf` | `200` **application/pdf** |

I first tried enumerating writers. It reached **nine** patterns (`csv`, `gzip`, `zlib`, `flate`, `bufio`, `io.Copy`, `fmt.Fprint*`, `io.WriteString`, `ServeContent`) and still missed excelize, PDF libraries, `archive/zip` and any house streamer — a list that grows forever. Worse, it could only *guess* the media type: that approach documents the PDF as `application/octet-stream`, where the header states `application/pdf`.

One call every streaming handler makes beats a list that never ends, and it carries the exact type — which closes **#354's remaining half** by construction rather than as separate work.

### Per framework, and configurable

`responseContext.contentTypeWrites` carries call + receiver + which argument is the header NAME and which the VALUE:

- net/http (chi, mux, http): `w.Header().Set(k, v)`
- gin: `c.Header(k, v)`, plus the net/http entry for `c.Writer.Header().Set`
- fiber: `c.Set(k, v)`
- echo: reaches it through net/http, so no shorthand needed

A project with a house context declares its own entry.

### Deliberate limits

- The body's schema says **bytes**, rather than naming a Go type that was never encoded.
- A `Content-Type` whose value is decided at **runtime** documents nothing rather than guessing (golden rule #7).
- A writer built over a **file** is not the response — the destination gate keeps an internal CSV dump out, and the fixture asserts it.

## ⚠️ The trade-off, measured

A handler that declares a content type and **then states a different status** gains a spurious `200` beside the real response:

```go
w.Header().Set("Content-Type", "application/json")
w.WriteHeader(http.StatusCreated)   // 201
json.NewEncoder(w).Encode(user)
```

**Five operations across four fixtures:**

| fixture | operation | gains |
|---|---|---|
| `enum_validation` | `POST /products/` | `200` |
| `enum_validation` | `POST /users` | `200` |
| `mux` | `POST /users` | `200` |
| `servemux` | `POST /users` | `200` |
| `wrapper_router` | `PUT /items` | `200` |

**Why it is here.** `ImplicitStatus` is the semantically right field — a status taken only if no explicit write claims the body first — and with it these handlers keep their `201` alone, with **zero drift across all 135 fixtures**. I implemented that and it works. But an opaque fragment carrying only an implicit status does not survive fragment assembly, so the streamed body is never documented at all, which is the whole point of the issue.

So today it is either "the feature works, with five spurious 200s" or "no spurious responses, and the feature does nothing". This PR is the first; the commit history carries the second, and the follow-up is to make an opaque fragment survive assembly and then switch the field back.

Both states are recorded in the changelog under **Known limitation** so it is not a surprise at release time.

## Coverage

`testdata/streamed_body` covers the CSV writer, `io.Copy`, and the negative case (a `csv.Writer` over a file, which must stay undocumented). Unit tests pin the header matcher — the canonical-spelling case, a non-Content-Type header, a runtime value, the wrong receiver — and that a framework shorthand is *added* to the stdlib entry rather than replacing it.

Suite, lint and the 96% coverage floor all green.

## Summary by CodeRabbit

- **New Features**
  - API documentation now recognizes streamed response bodies and uses declared media types such as CSV and PDF.
  - Binary and file responses are documented with appropriate byte-based schemas.
  - Responses with runtime-determined content types are documented without an assumed schema.
  - Supported web frameworks now better recognize content types set through response headers.

- **Known Limitations**
  - Declaring a content type before writing a later response status may produce an additional implicit `200` response in the documentation.


ehabterra:
Both findings fixed — and the first is the more important, because it contradicted the whole reason these calls are configuration.

**Fixed setter-name filter (Major)** — confirmed. `MatchNode` applies the pattern's `CallRegex` *before* `ExtractResponse` can inspect the header-NAME argument, so the hardcoded `^(Set|Header|Type)$` silently discarded any project configuring its own setter. A house context with `SetContentType` would have had its `contentTypeWrites` entry never evaluated and its streamed response dropped — the calls being config would have meant nothing.

Now ORed from the configured entries, deduped and sorted so the derived expression is reproducible.

One subtlety worth recording: with no writes configured it must match **nothing**, and `$^` does not do that — it matches the empty string, so an unconfigured pattern would have claimed every call whose name renders empty. `[^\s\S]` is a class that cannot match any character. A test pins it.

**Missing writer provenance (Major)** — confirmed. gin, echo and fiber declare no `WriterTypeRegexes`, so the destination resolver is disabled and `RequireResponseDestination` could not run at all. A `Header().Set("Content-Type", …)` on an **outbound** request would then be documented as an opaque response body — the confusion of #513 and #519, a third time.

I have **not** added provenance to those three frameworks here. Enabling the destination resolver changes response gating for *every* pattern in them, not just this one, and needs its own drift measurement — it is a larger change wearing a small disguise.

Instead: unverifiable is not claimed. A framework whose `ResponseContext` declares no writer types documents no streamed body until it does. chi, mux and net/http carry provenance, so the reported CSV and PDF cases are unaffected. gin/echo/fiber get nothing rather than something possibly wrong — which is their status quo before this PR, so no regression.

Giving those three real writer provenance is worth doing and I would suggest it as a follow-up issue rather than folding it in here.

Re-measured: suite, lint and the 96% floor green; drift is the **same four fixtures already disclosed**, unchanged.

PR #535, opened 2026-09-20
fix(spec): read a declared media type through constant resolution

ehabterra:
Follow-up to #534, which **merged before this commit landed** — so `main` currently ships the streamed-body feature in exactly the state that failed when it was tried on a real project.

Reported symptom:

> still `GET /api/v1/audit/export.csv` get only 500 nothing more

## Cause

The matcher accepted the media type only as an **inline string literal**. Real handlers name it with a constant:

```go
const mimeCSV = "text/csv"

w.Header().Set("Content-Type", mimeCSV)      // documented nothing
w.Header().Set("Content-Type", "text/csv")   // worked
```

So any project that keeps its media types in one place — which is most of them — got nothing from the feature at all. The same applied to another package's constant (`echo.MIMETextPlain`, a house `mime` package).

The fixture only had the inline form, which is why nothing caught it: it tested the shape I had written, not the shape people write.

## Fix

Both arguments now go through `ContextProvider.ConstantValue`, which already resolves a literal, a package-local constant and a qualified one — and is what the rest of the extractor uses for exactly this. The private `literalText` helper it replaces should never have existed beside it.

A value that is **not** constant still yields nothing rather than a guess: the media type is decided at runtime, and an operation documented with the wrong content type is worse than one documented with none (golden rule #7).

## Verified

| handler | before | after |
|---|---|---|
| inline `"text/csv"` | ✅ | ✅ |
| `const mimeCSV` | ❌ **nothing** | ✅ `text/csv` |
| declared in a shared helper, beside `Content-Disposition` | ✅ | ✅ |
| `io.Copy` + `application/pdf` | ✅ | ✅ |
| `csv.Writer` over a **file** (must stay undocumented) | ✅ | ✅ |

The fixture gains the two shapes real code has — the constant, and the declaration made inside a shared header helper.

Suite, lint and the 96% coverage floor green; fixture drift unchanged from what #534 already disclosed.

## Note for the reporter

If the export is still bare after this, the next likely difference is the **router**. gin, echo and fiber declare no `WriterTypeRegexes`, so the pattern refuses rather than guesses without writer provenance — deliberate, and called out in #534. Giving those three real provenance is the outstanding follow-up.

## Summary by CodeRabbit

* **Bug Fixes**
  * API documentation now correctly identifies response media types declared through named constants or shared header helpers.
  * Generated response documentation omits media types when the header name or value cannot be resolved reliably.
  * CSV and plain-text response types are now documented consistently, including when declarations are not inline string literals.


ehabterra:
## Your doubt was right in principle — and fixing it removed the trade-off

> I have a doubt that whenever content type is not default the body is always string

I could **not** reproduce the reported output across two probes (a typed `json.NewEncoder` body, and an unresolvable `w.Write` body, both with `application/json; charset=utf-8`) — in both the real schema or the existing `w.Write` pattern won. So I cannot claim this is the cause of what you saw on `DELETE /api/v1/masters/src-code-memory/{id}`.

But the principle holds regardless, and the code was wrong: an opaque `{type: string, format: binary}` is the truth for a stream and a **lie** for JSON. It stops a generated client parsing a document it could read — worse than saying nothing.

**So a media type a serializer already describes is never given an opaque body.** Where the body resolved, that serializer documents it properly; where it did not, *unresolved* is the honest answer. Which types those are is read from the config — `Defaults.ResponseContentType` plus any response pattern that both names a media type and reads a body type — not a hardcoded list.

The charset is ignored in the comparison, so `application/json; charset=utf-8` is the same format. Comparing whole strings would have let exactly your case through.

### It also deletes the known limitation

The four fixtures I disclosed as a trade-off in #534 were **all** handlers setting `Content-Type: application/json` and then encoding a real value — precisely the shape now refused:

| | before | after |
|---|---|---|
| fixture drift | 4 fixtures, 5 spurious `200`s | **0** |

So the cost you accepted to get this feature is gone. The **Known limitation** changelog entry is deleted rather than reworded.

Streams are untouched — `text/csv`, `application/pdf` and the rest still document their body — and a boundary test pins both directions, including the `.xlsx` vendor type.

### Still worth checking

If that DELETE is still wrong after this, I would want to know the router and how the body is written, since neither probe reproduced it. gin/echo/fiber still produce nothing here by design (no `WriterTypeRegexes`), which remains the outstanding follow-up.

PR #538, opened 2026-09-20
feat(cli): --strict fails the build on a quality shortfall (#297)

### LINKED ISSUE #297: Add --strict: promote quality warnings (unmapped auth middleware, dropped response types) to a non-zero exit

ehabterra:
Closes #297.

## The problem

Several conditions today are a `log.Printf` and nothing else:

```
[security] 1 middleware look like authentication ... but map to no security scheme,
           so the operations they guard are documented as PUBLIC
```

The message is correct and helpful. The **consequence** is that the operation is emitted with no `security` block. In a CI log that line scrolls past, and its effect is invisible in the spec diff to anyone not already looking for it: an operation whose `security` block was dropped reads exactly like an operation that genuinely has none. "A route silently lost its documented auth" is the class of change that should fail a build, and today it cannot.

The same goes for a route that lost its path, a response that lost its schema, and a package that never loaded.

## What this adds

`--strict`, off by default, which promotes those already-reported conditions to exit code **3**.

```bash
apispec -o openapi.yaml --strict                 # gate on everything
apispec -o openapi.yaml --strict=security,paths  # only these (attach the value with =)
apispec -o openapi.yaml --strict=false           # turn off a --strict inherited from a wrapper
```

| Category | Fails when | What it costs the document |
|---|---|---|
| `security` | auth middleware matched no `securityMappings` entry | the endpoints behind it are documented as **public** |
| `paths` | a registration's path is built at runtime, or nothing matched at all | the endpoint is **missing entirely** |
| `schemas` | a response came out with an empty schema, or a `$ref` had no component | the operation is there, its **shape is not** |
| `truncation` | `--max-nodes` / `--max-nodes-per-route` cut the walk short | the operation reads as finished with **fewer params/responses** than the code has |
| `packages` | in-module packages failed to load or parse | whatever they registered is absent, by an **unknown amount** |

Granular because these are not equally tolerable — a project may accept a truncated route while refusing to let an endpoint lose its documented authentication, which is the design note the issue asked to settle.

## Decisions worth reviewing

**Exit 3, not 1.** A CI script has to be able to tell "apispec could not run" from "apispec ran and the result is below the bar": the second is a spec to look at, the first is a build to fix. 2 is taken — the flag package exits with it on a usage error.

**The document is unchanged and still written.** `--strict` reads only state the engine already recorded, so a strict run's spec is byte-identical to a non-strict one (verified by diff). A failing job can still publish the artifact and diff it, which is how anyone works out what was lost. It also means a project can turn the gate on without re-reviewing its output.

**The `security` gate fires even when no `securityMappings` are configured at all**, which is the one place it parts company with the warning it promotes. That warning stays quiet there because auth detection is effectively off and the line would be noise on every run; a gate that was explicitly asked for is not noise, and a project with auth middleware and no mappings is the *worst* case of this finding, not an exempt one. Noted in the code and the CHANGELOG.

**`strictFlag` implements `IsBoolFlag`.** Without it the flag package swallows the next argument, so `apispec --strict ./api` would parse as `--strict="./api"` and analyse the default directory. The cost is that the value must be attached with `=`, the same rule Go's own bool flags follow; the help text says so and `TestStrictFlagDoesNotEatTheDirectory` guards it.

**Only `security` counts classified auth middleware.** `UnclassifiedMiddleware` (logging, recovery, rate limiting) is deliberately excluded — #520 separated those two precisely because a gate that fires on every logger is a gate people switch off.

## Verification

- `internal/engine/strict_test.go`: per-category findings, each asserted to fire in its own category and **nowhere else**; a clean run and an empty category list produce nothing; report order is fixed across repeated runs (golden rule #1); middleware names sorted and capped. `strict.go` is at 100% statement coverage.
- `cmd/apispec/strict_test.go`: flag parsing (bare / `=true` / `=false` / list / unknown category), the positional-argument guard, and an end-to-end run over a module with a call graph and no routes — generate, then assert the gate fails with `[paths]`, and that `--strict=security` passes on the same run.
- `go test ./...`, `make lint`, `go vet` clean; coverage ratchet at 96.0% (floor 96).
- **gitea (988 paths)**: exits 3 on `truncation` only — `security`, `paths`, `schemas` and `packages` are all clean, so the gate is not noisy on a real project.
- Zero spec drift: same fixture with and without `--strict` diffs identical.

Docs: README (a "Fail the build when the spec comes out incomplete" section under Quick start), `docs/TOOLS.md` (flag table + a `--strict` section), `cmd/apispec/README.md`, CHANGELOG.

## Summary by CodeRabbit

- **New Features**
  - Added a `--strict` mode that fails the build with exit code `3` when the generated API specification is incomplete.
  - Supports enabling all checks or selecting categories such as security, paths, schemas, truncation, and packages.
  - The specification is still written unchanged, allowing failed builds to publish and compare the output.

- **Documentation**
  - Added usage details, category descriptions, exit-code guidance, and CI examples for strict mode.


ehabterra:
Both CodeRabbit findings applied in fa6e25e — both accurate.

1. **Count findings, not gates.** The log line printed `%d quality gate(s) failed` over `len(findings)`, and one category can produce more than one: `schemas` reports a response that lost its schema and a reference that lost its component separately, because they are different things to go and look at. Now `%d strict finding(s)`, with the reason in a comment, and the test pins the new wording.

2. **"Instead of only warning" was wrong**, in exactly the case this PR documents as deliberate: the `security` gate fires when no `securityMappings` are configured at all, where there is no warning above it to promote. Reworded to "instead of exiting 0" in the flag usage, README, `docs/TOOLS.md` and `cmd/apispec/README.md`, and the exception is now stated in README and TOOLS.md rather than only in the CHANGELOG.

Issue #546, opened 2026-09-21
A value assigned in a handler never binds to the helper it is handed to — producer relations exist only for main and directly-called functions


ehabterra:
Found while fixing #544. Root cause is in metadata; the symptom shows up wherever the tracker binds a helper's parameter to the call that produced its argument (golden rule #11's binding class).

## Measured

```go
func declarePDF(h http.Header) { h.Set("Content-Type", "application/pdf") }

func a(w http.ResponseWriter, r *http.Request) { declarePDF(w.Header()); … }   // ✅ documented
func b(w http.ResponseWriter, r *http.Request) { h := w.Header(); declarePDF(h); … } // ❌ lost
```

`b`'s declaration never reaches the extractor. The tracker drops `w.Header()` **and** `declarePDF(h)` from `b`'s children (the barren-subtree prune, #318, finds nothing reachable under them).

## Root cause — two facts that compose

1. **`BuildAssignmentRelationships` records a caller-scope `v := call()` link only for `main`.** Every other function contributes relations only through `edge.AssignmentMap`, which holds a *callee's* body — so a function nothing calls directly contributes none. That is every handler (registered as a value, never called), and every concrete method reached only through an interface. So `argProducerKey` finds no producer for `h`, and the binding fails.
2. **Claims are global, bindings are per call site.** When *another* call site binds the helper's parameter successfully (`declarePDF(w.Header())` in `a`), the helper's `h.Set` is marked `claimed` and removed from the helper's own body for **every** caller. Wherever a binding then fails — as in `b` — the write is on no path at all.

A third defect surfaced when (1) was widened: `buildCallGraph` hands a pending `AssignStmt` to the **next call the walk meets** even when the assignment has no call on its right-hand side. `c := &Ctx{…}` followed by `c.JSON(…)` records `c.JSON` as the producer of `c`. Harmless while relations were main-only; once every function's relations are read, the call becomes its own producer and the cycle guard drops it (`testdata/wrapper_router` `/api-items` loses its status). Two metadata goldens carry the leaked value today (`example.yaml` `callee_recv_var_name: u`, `generic.yaml` `…Container[T].Value`).

## A fix exists, and is parked — because it moves much more than this

Branch `wip/handler-producer-relations` (7f4497c):

- relations for every declared function and method, keyed by receiver type so same-named methods don't overwrite each other, via one index (`callerScopes`) rather than a per-edge scan;
- `argProducerKey` looks an ident up by the **exact** caller identity first;
[...]

Issue #470, opened 2026-09-09
One status can carry only one media type: a handler that writes both JSON and XML documents whichever fragment wins


ehabterra:
Split out of #354, with a reproduction, because it is a structural limit rather than a matching failure.

## What

`RouteInfo.Response` is `map[string]*ResponseInfo` (`internal/spec/extractor.go:57`) and `ResponseInfo` carries a single `ContentType` (`:196`). A status is therefore a key that holds **one** media type, so an operation that renders two cannot be documented — the mapper writes one entry under `content` (`mapper.go:1062`).

## Reproduction

`testdata/renderer_content_types` (on `main` after #354's response half):

```go
func getItemBoth(w http.ResponseWriter, r *http.Request) {
	if r.Header.Get("Accept") == "application/xml" {
		_ = xml.NewEncoder(w).Encode(Item{})
		return
	}
	_ = json.NewEncoder(w).Encode(Item{})
}
```

Generated:

```yaml
  /items/both:
    get:
      responses:
        "200":
          content:
            application/xml:      # the json branch is not documented
              schema: {$ref: ...}
```

Which one survives is decided by response-fragment ordering, not by anything about the handler — before the media types were correct it was the JSON one, now it is the XML one. Both are half the truth: a client reading the document has no way to know the other representation exists, and content negotiation is exactly what this endpoint shape is for.

`TestTestdata_RendererContentTypes` pins the current behaviour as a change-detector (`len(both) != 1` fails), with a comment saying the assertion to write when this is fixed is *both* media types.

## Why it is not a matching problem

Both fragments are found and both are extracted correctly — with the right media type each. They are then merged into one slot because the slot is keyed by status alone. Nothing in the extractor can fix that; the shape has to change.

## Sketch

`ResponseInfo` grows a per-media-type dimension, or `RouteInfo.Response` becomes keyed by (status, content type) with the mapper folding them into one `Response` with several `content` entries. The dedup in `pairAndFillResponses` (frame chain, call site, status, body) needs the media type in its key too, or the second fragment is still dropped before it reaches the map.

Touches `extractor.go` (ResponseInfo, pairing), `method_dispatch.go` (which builds its own `Response` map) and `mapper.go` (content construction). Worth doing with #363 (response headers), which needs a similar widening of the same struct.

PR #547, opened 2026-09-21
fix(spec): raw bytes take the media type the handler states (#544)

### LINKED ISSUE #544: A Content-Type declared by a helper that takes http.Header is never documented
Found while fixing #543; present on `main` before that fix too, so it is not caused by it.

## Measured

```go
func declarePDF(h http.Header) {
	h.Set("Content-Type", "application/pdf")
}

func exportPDF(w http.ResponseWriter, r *http.Request) {
	declarePDF(w.Header())
	_, _ = w.Write(pdfBytes)
}
```

documents `GET /export` with **`200 application/json`** — the default content type — rather than `200 application/pdf` as an opaque body. The declaration inside the helper is never claimed.

The same header write made directly in the handler IS claimed, and so are the other indirect shapes:

| shape | documented |
|---|---|
| `w.Header().Set(…)` in the handler | ✅ |
| `h := w.Header(); h.Set(…)` | ✅ |
| helper handed `w`: `declareCSV(w)` → `w.Header().Set(…)` | ✅ |
| set through a wrapper, `lw := &loggingWriter{w}; lw.Header().Set(…)` | ✅ |
| **helper handed `w.Header()`**: `declarePDF(w.Header())` → `h.Set(…)` | ❌ |

So the gap is specifically a header write whose receiver is a **parameter of type `http.Header`**. Not diagnosed further — this is a false negative (a missing streamed body), not a wrong one, and it is out of #543's scope.

Note it cannot be a media type the serializers cover: `application/pdf` is not one (`text/plain` would be skipped deliberately by `serializerCovers`, which is a different and correct outcome).

## Reproduction

`testdata/outbound_content_type` `/pdf-param` (added with #543's fix), asserted as a change-detector so it flips when this is fixed. `/forwarded` in the same fixture is the negative twin — the same helper handed an OUTBOUND request's header — and must stay undocumented once this is fixed; #543's `HeaderWriteDetached` already follows a parameter to its caller's argument for exactly that case.

Related: #517, #543.

ehabterra:
Closes #544. Follow-up: #546.

## #544's diagnosis was wrong — the real bug is bigger

The issue said a `Content-Type` declared by a helper taking `http.Header` is lost. Probed, the helper is fine on its own. What loses the declaration is a **raw write on the same status**, in the plainest shape there is:

```go
w.Header().Set("Content-Type", "application/pdf")
w.Write(pdf)
```

→ `200 application/json`, `{type: string, format: byte}` — a base64 string a generated client decodes as JSON, for a PDF. `^Write$` types its body from the `[]byte` argument under the JSON default, the declaration is an opaque fragment with no body type, and the slot merge keeps whichever fragment has a type — so the byte slice always won.

## The fix, as one principle across frameworks

**A byte slice names no media type, so raw bytes take the one the handler states.**

- **Stated in a header** (every framework's `w.Write`): `ResponsePattern.RawBody` marks a call that writes its argument verbatim. A body it writes that no `bodyTransforms` serializer produced is *raw*, and `declaredOverRawBytes` keeps the declaration over it on that status.
- **Stated in the call** (gin `c.Data(code, contentType, data)`, echo `c.Blob(code, contentType, b)`): `ContentTypeFromArg` / `ContentTypeArgIndex`. These were broken harder — gin's `c.Data` fell into the renderer catch-all, which reads the body from argument **1, the media type**, and echo's `c.Blob` wasn't recognised at all. `rawBodyRendererPatterns` declares both frameworks' raw renderers (plus the reader forms `DataFromReader` / `Stream`) ahead of the catch-all, from one constructor.

Raw bytes under a media type no serializer describes are `binary`; under JSON they stay typed (`serializerCovers`, same rule as #517). Both knobs are config, documented in `CONFIGURATION.md` with a house-renderer example, and editable in the UI.

## Measured against `main`

| project | changed | what |
|---|---|---|
| 2 private services (830 paths) | 25 ops | each a base64-JSON body → its declared type: 20 XLSX exports, 3 PDF, 1 ZIP, 1 HTML |
| **photoprism** | 29 ops | every one a `c.Data(…, "image/svg+xml", icon)` that was documented as a JSON **string** |
| **gitea** (988 paths) | 3 ops | two Atom feeds, one `text/json`; instance-cap drops and spec-mapping time unchanged |
| 8 other external projects | — | byte-identical |

Every changed operation was classified; none is anything other than this fix.

## Coverage

- `testdata/outbound_content_type`: `/pdf-direct`, `/pdf-param`, `/invoices` document `application/pdf` / `application/zip` as binary.
- `testdata/streamed_body_{gin,echo,fiber}`: a `/raw.pdf` route in each framework's own idiom, asserted binary under `application/pdf` in the per-framework test. gin and echo fail without this change (`application/json` / `default`).
- `TestDeclaredOverRawBytes` for the merge rule.

## Split out: #546

What's left of the reported shape — a helper handed a **variable** holding the header, when another call site hands the same helper `w.Header()` directly — is a tracker binding gap, not a response one: a handler's `h := w.Header()` records no producer (relations exist only for `main` and directly-called functions), and the other call site's binding has claimed the helper's write away from its own body for every caller. A fix exists on `wip/handler-producer-relations`, but it moves ~45 gitea operations, misattributes three on photoprism, and costs +7% on gitea — so it gets its own review. `/pdf-var` and `/reports` are asserted here as change-detectors.

One shape deliberately **not** changed: an *undeclared* raw write (`w.Write(b)` with no `Content-Type`) still documents the JSON default. That's a separate call — see the PR discussion before widening it.

## Summary by CodeRabbit

- **New Features**
  - Improved API documentation for handlers that return raw byte data.
  - Responses can now derive their media type from renderer arguments, such as PDF or ZIP output.
  - Raw byte responses are documented with the declared content type and a binary schema instead of a base64-encoded JSON string.
  - Added configuration guidance and examples for raw-body and media-type handling.

- **Tests**
  - Added coverage for raw responses across supported web frameworks and streamed content.


ehabterra:
Re CodeRabbit's outside-diff finding (**Preserve binary media-type alternates**, `extractor.go:1617-1624`): verified and fixed in e480568.

**Reproduced** on a negotiated gin endpoint — `c.Data(200, "application/pdf", pdf)` on one branch, `c.JSON(200, invoice)` on the other. It documented `application/json` alone: the stated PDF has an empty `BodyType`, the slot merge checked "keep the informative one" before checking for a different media type, and `otherMediaType` required a body type on both sides.

**Fix**, as suggested: `isRepresentation` treats bytes under a media type the handler states (`MediaTypeDeclared` + `{type: string, format: binary}`) as a representation, and the media-type case now runs ahead of the body-type preference, so the two compose into `content` via `addAlternateMediaType` (#470). Keyed on the declaration as well as the schema, so an opaque body with no stated type still cannot claim a media type.

**Coverage**: `testdata/streamed_body_gin` `/invoice`, asserted in `TestTestdata_StreamedBodyNegotiated` (fails before the fix). Eleven external projects are byte-identical, since none has the shape.

PR #548, opened 2026-09-21
wip(tracker): producer relations for every function; scope pending assignments to their own call (#546)

ehabterra:
**Draft — not ready to merge.** Opened to put the change and its measured drift up for review. See #546 for the full diagnosis.

## What it changes

1. **`BuildAssignmentRelationships` records caller-scope `v := call()` links for every declared function and method**, not only `main`. Handlers (registered as values, never called) and concrete methods reached only through an interface contributed none, so a value a handler builds could never bind to the helper it's handed. Indexed once (`callerScopes`) rather than scanned per edge, and keyed by receiver type so same-named methods don't overwrite each other.
2. **`argProducerKey` looks an ident up by the exact caller identity first**, then falls back to the bare-name key.
3. **A call takes a pending assignment only if it is one of that assignment's right-hand sides.** `c := &Ctx{…}` had been stamped onto the next call (`c.JSON(…)`), which then recorded itself as `c`'s producer. Two metadata goldens change for exactly this (`callee_recv_var_name: u`, `…Container[T].Value` removed).

The suite passes, and it fixes `/pdf-var` and `/reports` in `testdata/outbound_content_type` (the change-detectors #547 added will flip).

## Why it's a draft

| project | vs `main` |
|---|---|
| 2 private services, 8 other external projects | identical beyond #544's own fix |
| **photoprism** | 3 ops: two upload endpoints **lose** their `200 Files` schema and **gain** `307`/`401`/`429` HTML from a static-route redirect and the rate limiter; one loses a `500`. **Bisected to change 3 alone**: the leak was masking a latent misattribution elsewhere. |
| **gitea** | ~45 ops: 28 parameter sets, 10 statuses gained, 2 lost, 7 bodies — some improvements, some dubious. **+7% spec mapping**, instance-cap drops 26M → 38M. |

## Before this can land

- [ ] Trace the photoprism misattribution that the leak was hiding, and fix it at its root.
- [ ] Classify each gitea change as a fix or a misattribution.
- [ ] Try making claims per-binding instead of global (#546, point 2): it may fix the variable shape without widening relations at all, which would be the smaller change.
- [ ] Rebase on #547, then flip its `/pdf-var` / `/reports` change-detectors into assertions.


ehabterra:
Superseded by #549 for #546: per-call-site binding claims fix the reported shape with no metadata change. Gitea gains 17 real `redirect_to` cookie params with unchanged cost, and photoprism gains one real operation fix, against this branch's ~45 mixed gitea changes and +7%.

The one idea here still worth landing on its own, the pending-assignment leak fix (`assignsCall`), is now #550, blocked on the photoprism misattribution it exposes. Leaving this draft open for you to close; the branch keeps the code for #550.

ehabterra:
Closing unmerged: superseded by #549 for #546. The branch wip/handler-producer-relations is kept for #550 (the pending-assignment leak fix).

Issue #552, opened 2026-09-21
A query key read off ANY url.Values is documented as a query parameter — including a config URI the client never sends


ehabterra:
The request-side twin of #543. Found while measuring #550 on gitea, and it blocks #550 from landing cleanly.

## Measured

```go
var connString = "redis://localhost:6379/0?clientname=api"

func connect() string {
	u, _ := url.Parse(connString)
	return u.Query().Get("clientname") // a CONFIG URI, not the request
}

func list(w http.ResponseWriter, r *http.Request) {
	_ = connect()
	page := r.URL.Query().Get("page") // the real query parameter
	…
}
```

documents `GET /items` with **both** `page` and `clientname` as query parameters.

On gitea, with #550's fix, 8 operations gain `clientname`, `skipverify` and `insecureskipverify`, read by `modules/nosql/manager_redis.go` off the Redis connection URI (`POST /`, `POST /-/admin/`, `…/compare`, `…/pull/{index}/target_branch`, `…/{type}/{index}`, `…/merge_box`, …). On `main` they're hidden only because less of that code is reachable.

## Cause

The query pattern in the net/http preset (inherited by every framework) is

```go
{CallRegex: "^Get$", ParamIn: "query", ParamArgIndex: 0, RecvType: "net/url.Values"}
```

with no provenance check at all: any `url.Values.Get` anywhere in a handler's call graph is a query parameter. `url.Values` sits on both sides the way `http.Header` does (#543): it's the request's query, but also any URL a handler parses, builds or forwards, plus form values it constructs for an outbound request.

The header patterns already carry `ExcludeRecvOriginRegex` for the response side. This needs the positive version: the `Values` must **originate from the request**, i.e. `r.URL.Query()`, `r.Form`, `r.PostForm` on a value of a configured `RequestContext` type, and not `url.Parse(…)`, `url.Values{}`, or `u.Query()` on some other URL. Following golden rule #7: drop only what provably doesn't come from the request (a `url.Parse` / literal / constructed origin), and keep what can't be placed.

## Reproduction

The program above; a fixture in `testdata/` with the fix, asserting `page` present and `clientname` absent. Check the other frameworks' own query accessors too (golden rule #5). Most read through their context (`c.Query`) and are already scoped, but any that go through `url.Values` inherit this.

Blocks #550.

Issue #551, opened 2026-09-21
gin's AbortWithStatusJSON / AbortWithStatus document nothing — directly or through a helper


ehabterra:
Found while measuring #550 on photoprism, where **120 of 140 operations document only `200`**: its error helper answers through gin's `c.AbortWithStatusJSON(code, resp)`.

## Measured

```go
func direct(c *gin.Context) {
	if c.Param("id") == "" {
		c.AbortWithStatusJSON(http.StatusNotFound, ErrorBody{Error: "not found"})
		return
	}
	c.JSON(http.StatusOK, Item{})
}

func bare(c *gin.Context) {
	if … { c.AbortWithStatus(http.StatusForbidden); return }
	c.JSON(http.StatusOK, Item{})
}

func abort(c *gin.Context, code int, msg string) { c.AbortWithStatusJSON(code, ErrorBody{Error: msg}) }
func viaHelper(c *gin.Context) { if … { abort(c, http.StatusConflict, "conflict"); return }; c.JSON(200, Item{}) }
```

All three document `200 Item` and nothing else — no `404` + `ErrorBody`, no `403`, no `409`.

## Cause

The gin preset's renderer patterns match `^(?i)(JSON|XML|YAML|String|HTML|ProtoBuf|Data|File|Redirect)$` on `*gin.Context`. `AbortWithStatusJSON(code, obj)` and `AbortWithStatus(code)` are neither — they are gin's own status-carrying writes, and every gin codebase that aborts from middleware or an error helper uses them. `c.AbortWithError(code, err)` is the same shape (status, no documented body).

## Wanted

Patterns in the gin preset, ahead of the catch-all:

- `AbortWithStatusJSON` — status arg 0, body type arg 1 (JSON);
- `AbortWithStatus` / `AbortWithError` — status arg 0, no body;

and a fixture in the style of `testdata/streamed_body_gin` covering all three shapes above, including the helper one (the parameter-traced status is the shape photoprism uses for everything). Then re-measure photoprism — this should move most of its 120 status-less operations.

Worth checking the same for echo (`c.NoContent` is covered; `echo.NewHTTPError(code, …)` returned as the handler's error is not, as far as I know) — golden rule #5.

PR #554, opened 2026-09-21
fix(metadata): a variable no call produced has no producer (#550)

### LINKED ISSUE #550: A pending assignment is handed to the next call even when its right-hand side has no call — and fixing it exposes a photoprism misattribution
Split out of #546 / draft #548.

## The defect

`buildCallGraph` records an `AssignStmt` when it meets one, then hands it to the **next `CallExpr` the walk visits**, whatever that is. When the assignment's right-hand side has no call, the next call is on a *later* statement:

```go
c := &APICtx{Ctx: &Ctx{Resp: w, Req: r}}   // no call here
c.JSON(http.StatusOK, []Item{})            // ← recorded as the producer of c
```

`c.JSON`'s edge gets `CalleeRecvVarName: c`. Two metadata goldens carry leaked values today: `internal/spec/tests/example.yaml` (`callee_recv_var_name: u` on `u.SetAge`, after `u := &User{…}`) and `generic.yaml` (`…Container[T].Value`, after a field assignment).

It's harmless only because producer relations are built for `main` and directly-called functions (see #546). Anything that reads caller-scope relations more widely turns such a call into its own producer, and the cycle guard drops it (`testdata/wrapper_router` `/api-items` loses its status).

## The fix is one line of logic

A call takes the pending assignment only if it **is** one of that assignment's right-hand sides (`assignsCall`, on `wip/handler-producer-relations` / #548). Both golden changes are the leak disappearing.

## Why it isn't landed

On its own, with nothing else from #548, it changes photoprism: two upload endpoints (`POST /services/{id}/upload`, `PUT /users/{uid}/upload/{token}`) **lose** their `200` schema and **gain** `307` / `401` / `429` HTML responses from a static-route redirect and the rate limiter, and `POST /users/{uid}/passcode` loses a `500`. Bisected: restoring the leak makes photoprism byte-identical. So the leak is accidentally *suppressing* a real misattribution elsewhere, most likely by claiming an edge away (or cycling it out) along the path that would otherwise carry those responses across.

## To resolve

1. Find the misattribution the leak masks on photoprism and fix it at its root.
2. Land the leak fix with the regenerated goldens, measured on gitea, photoprism and the external set.

Related: #546, #548, golden rule #9 (a fix that trades one wrong answer for another is not progress, which is why this waits).

ehabterra:
Closes #550. **Merge after #553 (#552).** This change makes more real code reachable, and without #553 gitea picks up false query parameters read off Redis connection URIs on 8 operations.

## Two defects, one of them already on `main`

1. **The leak.** `buildCallGraph` handed a pending `AssignStmt` to the *next* call the walk met, even when the assignment had no call on its right-hand side. `c := &Ctx{}` followed by `c.Do()` recorded `c.Do` as the producer of `c`. Now a call takes the assignment only if it **is** one of its right-hand sides. The two metadata goldens change for exactly this: `callee_recv_var_name: u` and `…Container[T].Value` are removed.
2. **The fallback, which is what #550 was really about.** `BuildAssignmentRelationships` linked a variable assigned inside a callee's body with no producing call (`name := r.URL.Path`, `counts[k] += 1`) to the **invocation of the function it's written in**. For a handler closure that function is the *route registration*, so any argument reading the variable expanded every route the registration registers beneath itself.

The leak had been pointing many of those variables at the next call instead, which was wrong but local. That's why fixing the leak alone *moved* the misattribution rather than exposing it: it was already on `main` (photoprism's `passcode`). The fallback now applies only to an assignment in the edge's **caller** scope. That's how `main`'s `g := r.Group(…)` is recorded, and it holds for every variable of a tuple. Anything else has no producer.

## Measured against `main`

| project | effect |
|---|---|
| **photoprism** | `POST /users/{uid}/passcode` loses a static route's `307`, the rate limiter's `401`/`429` HTML and an empty `500`; its `200` becomes the real `Passcode` type, matching its three sibling routes. `PUT /labels/{uid}` gains its `Label` type. Three operations lose OAuth/download query params they never read. The share preview loses a `200` that #549 had gained only through this fallback. |
| **a private service** | 3 of 17 routes behind one capability guard lose a `401` and `X-Screen` borrowed from the auth middleware through a variable named `claims`, so all 17 now agree |
| **gitea** | tree work drops **~45×** (instance-cap drops 26.4M → 0.58M). Among 84 changed operations, the samples I traced were fixes: form bodies from other handlers go, and `WebThemeApply` writes no response yet was documented with `200 net_http_Request`, `400` and `404`. Proxy headers from the shared `GuessCurrentAppURL` move between operations. **Needs #553** for the Redis-URI query params. |
| 8 other projects | byte-identical |

## Coverage

- `testdata/closure_local_producer`: two handler closures in one registration function. On `main`, `POST /items` documents its sibling's `307`.
- `TestPendingAssignmentDoesNotLeakToNextCall`, `TestAssignmentWithoutProducingCallHasNoRelation`: both fail on `main`.
- `TestCallerScopeAssignmentKeepsItsProducer`: the case the fallback exists for (tuple included) keeps working.

Found while measuring, and filed separately: #551, gin's `AbortWithStatusJSON` / `AbortWithStatus` document nothing, which is why 120 of photoprism's 140 operations show only `200`.


ehabterra:
gitea with #553 applied first (the intended merge order): the 8 operations that gained Redis-URI query parameters under #554 alone no longer have them. Relative to #554 alone the pair changes exactly those 8 operations, each losing `clientname`, `skipverify` and `insecureskipverify`, and nothing else. Tree work stays ~45× lower than `main` (instance-cap drops 0.58M vs 26.4M).

Issue #556, opened 2026-09-21
echo and fiber errors RETURNED from a handler (echo.NewHTTPError, fiber.NewError) document no status


ehabterra:
Found while fixing #551 (gin's abort family); the same gap, reached differently on the frameworks that answer errors by **returning** them.

## Measured

```go
func get(c echo.Context) error {
	if c.Param("id") == "" {
		return echo.NewHTTPError(http.StatusNotFound, "not found")
	}
	if c.Param("id") == "x" {
		return echo.ErrForbidden
	}
	return c.JSON(http.StatusOK, Item{})
}
```

documents `200` only: no `404`, no `403`.

## Cause

gin writes an error through a call on the context (`c.AbortWithStatusJSON(code, …)`), which a response pattern can match (#551). echo and fiber don't. The handler **returns** an error value carrying the status, and the framework's error handler renders it:

- echo: `echo.NewHTTPError(code, msg…)` and the `echo.Err*` sentinels (`ErrNotFound`, `ErrForbidden`, `ErrUnauthorized`, …); the default `HTTPErrorHandler` writes `{"message": …}` with that code.
- fiber: `fiber.NewError(code, msg…)` and `fiber.Err*` sentinels; the default `ErrorHandler` writes the message as text with that code.

Nothing reads a handler's **return** value as a response, so these are invisible however they're written. The same is true of a house error handler (`e.HTTPErrorHandler = …`) that renders a custom body.

## Wanted

A config-driven notion of an **error constructor**: a call (or a sentinel variable) whose result carries a status at a known argument, recognised when the value is *returned* from the handler (directly, or through a helper's return), and documented with the framework's error-handler body. Per golden rule #5 that belongs in the pattern config, with echo and fiber presets. A fixture per framework covers constructor, sentinel and helper-returned forms.

There's existing machinery to check for reuse. `statusFromConstructorField` already resolves a status stored by a constructor call (#189), and the returned-value resolvers used for response wrappers may cover the "returned from the handler" half.

Related: #551.

PR #557, opened 2026-09-21
fix(spec): an error a handler returns documents the status it carries (#556)

ehabterra:
Part of #556: **constructors** (`echo.NewHTTPError`, `fiber.NewError`). Returned **sentinels** (`echo.ErrForbidden`, `fiber.ErrForbidden`) remain; the fixture asserts them as a change-detector, and #556 stays open.

## The gap
echo and fiber answer errors by *returning* them, and the framework's error handler writes the status and body. Nothing is called on the context, so no response pattern saw them.

## The fix
- **metadata:** `CallGraphEdge.ResultType`, the static type of a call's single result, from the type checker. It's needed because a dependency's declarations aren't loaded. It isn't serialized, so the goldens are unchanged; timing is within noise.
- **`ResponsePattern.TypeFromResult`:** the body type is the call's result.
- **registry:** `echo.HTTPError` as echo renders it, `{"message": …}` (v3/v4/v5).
- **presets:** echo `NewHTTPError(code, …)` → status arg 0, body from the result; fiber `NewError(code, msg)` → status arg 0, message as the `text/plain` body.

## Measured against `main`
| project | effect |
|---|---|
| fiber todo recipe | +11 statuses, each a `fiber.NewError` its services return (409/422/401), checked against source |
| 10 other external projects | byte-identical |

## Not in scope
- Returned sentinels (the rest of #556).
- Middleware applied with group-level `Use` isn't walked per route (go-clean-echo's JWT `401`s stay unattached). That's an existing limitation.

## Summary by CodeRabbit

* **New Features**
  * Added support for deriving documented response body types from handler-returned values.
  * Added automatic response documentation for Echo HTTP errors across supported Echo versions.
  * Added automatic response documentation for Fiber errors, including status codes and plain-text responses.
  * Added an option in response-pattern configuration to use a call’s returned value as the response type.
  * Improved schema generation for Echo HTTP error messages.


ehabterra:
@coderabbitai review

